PR #2, opened 2026-04-07
feat: unify Gemini export rendering and Drive sync in gmd

Sinity:
## Summary

I consolidated the early Gemini export scripts into a single `gmd` CLI and used that consolidation to make local renders, Drive syncs, attachment handling, auth bootstrap, and project-local defaults behave like one archive workflow instead of several adjacent tools. The result is a command surface that can render AI Studio JSON exports to annotated Markdown, mirror a Drive folder into a local Markdown tree, and preserve enough metadata for the archive to be useful in Obsidian-style note-taking flows. Pulled the rendering, Drive, config, and utility logic into `geminimd/` so the CLI coordinates shared behavior not embedding it. This is the point where the repo stops feeling like a collection of scripts and starts looking like a tool.

## Motivation

The underlying problem here was not just “convert one JSON file to Markdown.” AI Studio exports are structured as `chunkedPrompt.chunks`, model turns can include internal thought blocks, and attachments show up as Drive document/image references embedded in several different JSON shapes. That makes a shell-only or `jq`-heavy workflow brittle very quickly. I wanted a path that could handle both the local-export case and the “sync a whole Drive folder” case with the same renderer and the same attachment semantics.

The branch history shows that I started with script-shaped tooling (`sync_ai_studio.py` and a converter) and only later consolidated on `gmd`. That was deliberate. I used the first pass to pin down the Markdown shape, attachment naming, and Drive mtime behavior before deciding what the long-lived command surface should be. Once those semantics felt stable, I extracted `geminimd/` and folded the file-conversion path into `gmd render`, which let me delete the sync-specific script and stop duplicating Drive/render logic.

Moved toward a config-first workflow near the end of the branch. Passing credentials and folder names on every invocation is workable for a one-off demo, but it is not pleasant for repeated use inside a project archive. That is why `.gmdrc`, env overrides, `status`, `init`, and `auth` appear here even though some of that command plumbing still needs follow-up.

### 1. One CLI instead of multiple ad hoc entrypoints

`gmd.py` is the primary entrypoint. It exposes a small subcommand tree and routes each operation through the shared modules:

```bash
python3 gmd.py render path/to/export.json --output-dir out
python3 gmd.py sync --folder-name "AI Studio" --output-dir synced
python3 gmd.py status
```

The important design choice is that `render` is not “single file only.” It accepts a file or a directory, performs the same chunk parsing in both cases, and can optionally download attachments unless `--remote-links` is set. `sync` resolves a Drive folder by name, downloads each JSON file, renders it with the same Markdown code path, and then stamps the resulting Markdown with the Drive modified time so reruns stay Git-friendly.

Added optional `fzf` selection for both local files and remote chats. That is a pragmatic compromise: it gives interactive filtering without introducing another substantial dependency or a permanent TUI architecture at this stage.

### 2. Shared rendering and attachment discovery

The rendering logic moved into `geminimd/render.py`, and that is where the branch stopped being “a script with some helpers.” The renderer understands:

- user text vs model text
- model thought + response pairing
- collapse thresholds for long replies
- attachment callouts for Drive docs and Drive images
- YAML metadata for model params, token counts, citations, and attachment totals

The key helper here is `extract_drive_ids()`. AI Studio exports are not tidy enough to assume one path like `chunk.driveDocument.id`; IDs can show up in nested dicts, URL strings, and other provider-shaped structures. Centralizing that traversal means future render changes do not need to rediscover attachment discovery rules.

### 3. Drive auth and download behavior

`geminimd/drive.py` owns OAuth, token caching, folder discovery, pagination, metadata fetch, and chunked download. The caching behavior matters because Drive folder-name lookup is otherwise annoyingly expensive on repeated runs. I kept that state in `~/.gmd_state.json` and run history in `~/.gmd_runs.json` so the project directory does not accumulate ephemeral bookkeeping.

Kept the attachment behavior opinionated: attachments are linked, not embedded, and they live in per-chat `_attachments/` folders. That gives stable Markdown and avoids ballooning the document body with binary-ish content.

### 4. Project-local configuration and reproducible environment

I added `.gmdrc` loading from the current directory upward plus `GMD_*` environment overrides. That keeps repo-specific defaults in the repo while still allowing automation and shell overrides. The Nix shell work is intentionally simple: `flake.nix` and `shell.nix` include the Google auth/Drive libraries and the Python utilities the renderer relies on.

## Testing

There is no automated `tests/` suite in this branch yet. I documented and used manual smoke tests instead:

- local dry-run rendering with downloads disabled
- local rendering with attachment download enabled
- Drive sync with folder discovery by name
- config resolution through `.gmdrc` and env overrides

The exact manual commands are recorded in `AGENTS.md`, which is why that file ships in the same branch.

## Notes

- `init` and `auth` are exposed in the parser but their handlers still need follow-up work; the parser contract is ahead of the implementation there.
- Error coverage is still mostly operational rather than tested: malformed AI Studio JSON, token refresh failures, and odd attachment shapes need dedicated tests once a test suite exists.
- I intentionally kept the renderer opinionated. This branch is about getting one stable archive format working end to end, not exposing every formatting choice as a flag yet.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 6 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 6 finding(s) confirmed fixed/stale.

Issue #5, opened 2026-04-07
Conversation data is scattered across four providers with no unified archive path


Sinity:
## Context

The rendering pipeline currently handles Gemini exports well — Drive sync pulls JSON, the chunk pipeline normalizes it, and the output lands as Markdown with attachments, stats, and optional HTML previews. But conversations don't live exclusively in Gemini. ChatGPT, Claude.ai, Claude Code, and Codex all produce exportable transcripts in different formats, and there's no way to get them into the same archive.

Each provider has its own export shape:

- **ChatGPT**: ZIP exports containing `conversations.json` with a DAG-structured message mapping, plus a `files/` directory for attachments. Content blocks can be text, code, tables, images, tool calls, or tool results.
- **Claude.ai**: ZIP bundles with `conversations.json` (sometimes wrapped in `{"conversations": [...]}`). Content uses typed blocks — `text`, `tool_use`, `tool_result`, `image`, `file`. Attachments live under `attachments/`, `files/`, or `assets/`.
- **Claude Code**: Local JSONL sessions under `~/.config/claude/projects/` with `summary`, `user`, `assistant`, `tool_use`, and `tool_result` entry types. Tool results need to be stitched to their calls by ID.
- **Codex CLI**: Local JSONL sessions under `~/.codex/sessions/` with `response_item` wrappers around messages, function calls, and function outputs. Entries include environment preamble that should be filtered.

## Motivation

Without import support, users must either manually convert these exports or maintain separate tooling per provider. The existing Markdown pipeline (chunk normalization, attachment extraction, token counting, HTML preview, collapsible callouts) would work for any provider if the input were normalized to the same chunk format.

The CLI-based tools (Codex, Claude Code) also produce new sessions continuously, so one-shot import isn't sufficient — there needs to be a sync mode that mirrors local session directories with skip/prune semantics, similar to the Drive sync.

## Scope

- Importer adapters for ChatGPT, Claude.ai, Claude Code, and Codex that normalize to the existing chunk pipeline
- CLI subcommands for one-shot import and continuous local sync
- Shared utilities for token estimation and large-text extraction to attachments
- A doctor command to validate local session directories before import
- A stats command to summarize output directories across providers
- The `geminimd` package name no longer reflects multi-provider scope and should be updated

## Out of scope

- Changes to the Drive sync flow or Gemini rendering
- Push-to-vault or search index integration
- Provider auto-detection from arbitrary paths

## Acceptance criteria

[...]

PR #6, opened 2026-04-07
feat: add multi-provider importers and local session sync

Sinity:
## Summary

Extends the tool from a Gemini-only renderer into a multi-provider conversation archive. Four new importers (ChatGPT, Claude.ai, Claude Code, Codex) all normalize into the existing chunk-based Markdown pipeline, so attachment handling, token estimation, HTML previews, and stats collection work identically regardless of source. A local sync layer adds continuous mirroring for CLI-based tools (Codex, Claude Code) with mtime-based skip and prune.

## Motivation

The rendering pipeline and attachment machinery already work well for Gemini exports, but conversations are spread across multiple providers. Rather than building separate tools for each, the approach here is to write thin adapter modules that normalize each provider's export format into the shared `chunks + per_chunk_links` representation, then let `build_markdown_from_chunks()` do the heavy lifting. This keeps the importers focused on parsing quirks and the pipeline responsible for output quality.

The package rename from `geminimd` to `chatmd` happens in the same commit because both changes touch every import path — splitting them would create a throwaway rename commit that conflicts with every other line changed here.

## Importer architecture

Each importer lives in `chatmd/importers/` and returns `ImportResult` (markdown path, optional HTML path, attachments dir, the `MarkdownDocument` for stats). The shared utilities in `utils.py` handle two cross-cutting concerns:

1. **Token estimation** — uses tiktoken when installed, falls back to word count. The tokenizer is cached per model string.
2. **Large text extraction** — `store_large_text()` checks line/character thresholds and spills oversized content into `_attachments/`, replacing the inline text with a head+tail preview linking to the full file.

### ChatGPT (`chatgpt.py`)

ChatGPT exports use a DAG-shaped `mapping` structure rather than a flat message list. `_gather_messages()` walks the mapping, extracts messages, and sorts by `create_time`. The `_render_parts()` function handles the variety of content types (text, code, table, image_file, tool_calls, tool_result, link). Attachments from the export's `files/` directory are indexed by both filename and stem for flexible matching.

```python
# The mapping walk — messages aren't ordered, they're a graph
def _gather_messages(conv: Dict) -> List[Dict]:
    mapping = conv.get("mapping", {})
    messages = []
    for node in mapping.values():
        msg = node.get("message")
        if not msg:
            continue
        # ... extract and sort by create_time
```

### Claude.ai (`claude_ai.py`)

The Anthropic export wraps conversations in `{"conversations": [...]}` (or sometimes a bare list). Content blocks use a `type` discriminator — `text`, `tool_use`, `tool_result`, `image`, `file`. Tool interactions are rendered as paired call/result code blocks. The file index searches `attachments/`, `files/`, and `assets/` directories.

### Claude Code (`claude_code.py`)

Session JSONL has distinct entry types (`summary`, `user`, `assistant`, `tool_use`, `tool_result`). Tool results are stitched back to their originating call by ID and merged into a single chunk. Summaries (compaction checkpoints) are collected into front-matter rather than inline messages. The session locator supports absolute paths, filenames, and glob-based suffix matching.

### Codex (`codex.py`)

Uses Pydantic for entry validation since the JSONL format includes several entry types we want to skip. Messages starting with `<user_instructions>` or `<environment_context>` are filtered as boilerplate. `function_call` / `function_call_output` pairs are merged by `call_id`.

## Local sync layer

`chatmd/local_sync.py` provides `sync_codex_sessions()` and `sync_claude_code_sessions()`. Both delegate to `_sync_sessions()` which:

- Enumerates `*.jsonl` files from the provider's default directory
- Skips files whose mtime hasn't changed since the last render
- Optionally prunes output files that no longer have a source session
- Optionally generates delta diffs using `write_delta_diff()` (tries `delta` binary first, falls back to `difflib.unified_diff`)
- Preserves source mtime on output files via `os.utime()`

## CLI changes

The `gmd.py` CLI grows six new subcommands:

| Command | Purpose |
|---------|---------|
| `import chatgpt` | One-shot ChatGPT ZIP/dir conversion |
| `import claude` | One-shot Claude.ai ZIP/dir conversion |
| `import claude-code` | One-shot Claude Code session conversion |
| `import codex` | One-shot Codex session conversion |
| `sync-codex` | Continuous Codex session mirroring |
| `sync-claude-code` | Continuous Claude Code session mirroring |
| `doctor` | Validates local session directories |
| `stats` | Summarizes output directories with date filtering |

Import commands support interactive skim pickers (list conversations, select with fuzzy search) or batch mode via `--all`/`--conversation-id`. The `summarize_import()` helper renders Rich tables when the terminal supports it.

## Configuration

`gmd_config.py` adds a JSON config system that searches `$GMD_CONFIG`, then `$XDG_CONFIG_HOME/gmd/config.json`, then `~/.gmdconfig`. It provides defaults for collapse threshold, HTML preferences, and per-command output directories. The doctor command reports config status and candidate paths.

## HTML improvements

`_transform_callouts()` converts Obsidian-style callout blockquotes (`[!INFO]-`/`[!TIP]+`) into interactive `<details>` elements with CSS transitions on the disclosure triangle. Plain blockquotes pass through unchanged.

## Test coverage

Seven test modules covering the importers, sync layer, doctor, HTML callouts, and stats:

- **test_codex_importer.py**: Validates tool attachment extraction (120-line output → spilled to file with preview), empty attachment dir cleanup
- **test_importers_multi.py**: Basic import for all three web providers, tool call pairing in ChatGPT, ZIP-with-attachment round-trips for both ChatGPT and Claude
- **test_local_sync.py**: Codex sync creates markdown, second run skips, prune removes stale files; Claude Code sync end-to-end
- **test_doctor.py**: Invalid JSONL detection
- **test_html.py**: Callout wrapping, plain blockquote preservation, open/closed state
- **test_stats.py**: JSON output with per-provider breakdown and date filtering

Ref #5

## Release Notes

* **New Features**
  * Added local synchronization for Codex and Claude Code sessions
  * Added multi-provider import support (ChatGPT, Claude, Claude Code, Codex)
  * Added Doctor command for session validation
  * Added Stats command for conversation analytics
  * Added diff artifacts for render and sync operations
  * Improved HTML rendering with collapsible callouts

* **Documentation**
  * Updated README with direnv workflow, local sync menu, import guides, and Provider Cheat Sheet
  * Added configuration documentation with environment variable override support

* **Configuration**
  * Added configuration discovery system supporting standard config paths


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 3 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 17 finding(s) confirmed fixed/stale.

PR #8, opened 2026-04-07
refactor: rebrand as polylogue and move config/state to XDG

Sinity:
## Summary

I renamed the project from its Gemini-era naming to `polylogue`, moved the main entrypoint to `polylogue/cli.py` with a thin `polylogue.py` launcher, and shifted config/state handling onto XDG-compliant paths. I used the rename as a chance to fix the remaining layout mismatches: environment variables use the `POLYLOGUE_*` prefix, provider outputs no longer fall back to one generic render directory, and docs/tests read like a multi-provider archive tool instead of a renamed Gemini utility. Most of the diff is mechanical package movement, but the behavioral effect is important for unattended use because credentials, tokens, run logs, and state no longer depend on the project root. This is the branch where the name, on-disk layout, and default paths finally match what the tool has become.

## Motivation

By this point the tool was no longer honestly described by `gmd` or `chatmd`. It can import multiple providers, sync local session stores, generate stats, and validate archives. Keeping Gemini-shaped naming around would make every later feature harder to explain and every public interface slightly misleading. The rename is not cosmetic; it is the point where the code, docs, and runtime paths stop telling users they are working with a single-provider Gemini tool.

The XDG move belongs in the same branch. Once we are talking about a multi-provider archival tool, storing credentials, tokens, and runtime state in ad hoc project-root locations becomes increasingly awkward, especially for automation. XDG config/state dirs are a better fit for long-lived auth material, JSON state, and run logs, and they line up with the config story from the previous branch much better than the older `.gmdrc` / flat-file leftovers did.

Wanted to fix the output-root semantics while the rename was already touching CLI help and docs. Provider imports should not all default to the generic render directory; a multi-provider archive reads much more cleanly when ChatGPT, Claude, Claude Code, Codex, and Gemini each have their own archive root.

### 1. Rename the package and entrypoint surfaces

The mechanical part is simple but wide-reaching:

- `chatmd/` becomes `polylogue/`
- `gmd.py` becomes `polylogue/cli.py`
- a new top-level `polylogue.py` forwards into `polylogue.cli.main()`

Kept the thin launcher because it preserves the simple “run the repo directly” workflow while still putting the real CLI implementation inside the package.

On the config/settings side, I moved `gmd_config.py` and `gmd_settings.py` into the package as `polylogue/config.py` and `polylogue/settings.py`. That removes the odd split where package code depended on top-level modules that looked more like app leftovers than library code.

### 2. XDG-compliant credentials, tokens, and state

The more important behavior change is path handling. Credentials and tokens resolve under `$XDG_CONFIG_HOME/polylogue/` and runtime state (`state.json`, `runs.json`) moves under `$XDG_STATE_HOME/polylogue/`. The code creates parent directories before first write so:

```text
~/.config/polylogue/
~/.local/state/polylogue/
```

can appear lazily on first use without special install steps.

Renamed the env var surface from `GMD_*` to `POLYLOGUE_*`, which affects retry tuning, auth mode, token path, no-color behavior, and config resolution. That is a deliberate compatibility break, but it keeps the public surface consistent with the new identity.

### 3. Provider-specific defaults and path correction

`OutputDirs` gains provider-specific roots so imports no longer all default to the render directory. Instead, the CLI help now shows defaults like `/realm/data/chatlog/markdown/chatgpt` or `/realm/data/chatlog/markdown/claude` directly on the relevant subcommands. That makes the archive layout visible and predictable at the point of use.

Corrected Claude Code’s `DEFAULT_PROJECT_ROOT` from `~/.config/claude/projects` to `~/.claude/projects`. This is a small diff but a material usability fix: the importer/sync commands should target the path that the tool actually uses.

### 4. Documentation and test rewrite

Because the rename is user-facing, I updated `README.md`, `AGENTS.md`, and `multi-provider.md` in the same branch. The docs now describe `polylogue.py`, XDG config/state locations, provider-specific output roots, and the updated command names. Rewrote test imports from `chatmd.*` to `polylogue.*` so the suite keeps tracking the real package name rather than a transition alias.

## Testing

The existing pytest suite still runs after the rename, and the main test-side change here is namespace churn: test imports now target `polylogue.*` throughout. This branch does not add new behavior-specific tests for XDG migration or provider default output directories, so those pieces are primarily validated through manual CLI/doc checks.

## Notes

- This is an intentionally breaking rename for imports, env vars, and filesystem locations.
- I removed the older TOML `.gmdrc` loader rather than trying to keep two config stories alive under the new name.
- The path rename and provider-default changes are meant to make later automation work feel natural; they are not just cosmetic cleanups.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 2 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 2 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #10, opened 2026-04-07
feat: add automation, persistence layer, and NixOS module

Sinity:
## Summary

Adds the infrastructure needed for Polylogue to run unattended: a document persistence layer that makes reruns idempotent, file watchers for continuous sync, systemd/cron snippet generation from a shared target registry, and a NixOS module that wires everything into declarative timers. Every successful write now also updates a local SQLite FTS index so downstream tooling can query metadata without reparsing Markdown.

## Motivation

Up to this point, every `polylogue import` or `polylogue sync` run rewrites all output files unconditionally. That means:

- Cron jobs waste IO rewriting identical files, and git repos accumulate no-op diffs
- Users who hand-edit rendered Markdown lose their changes silently on the next run
- There's no single place to check whether a conversation has already been processed
- Setting up unattended sync requires manually writing systemd units or cron entries, duplicating CLI flag knowledge

The persistence layer solves the first three problems; the automation framework and NixOS module solve the fourth.

I considered adding skip logic directly in each importer, but the duplication was already bad — four importers plus render and sync all had near-identical write-then-optionally-HTML-then-optionally-diff sequences. Extracting `persist_document()` as a shared function was the natural consolidation point, and it gave a clean place to hang content hashing, dirty detection, and index updates.

## Document persistence (`document_store.py`)

The core abstraction is `persist_document()`, which every importer and both render/sync commands call instead of writing files directly. It:

1. Computes a SHA-256 content hash of the document body
2. Checks `state.json` for a matching `(provider, conversation_id)` entry
3. Skips the write if the stored hash, collapse threshold, attachment policy, and HTML flag all match — returns a `DocumentPersistenceResult` with `skipped=True`
4. Detects dirty-local state: if the on-disk body hash doesn't match the stored hash, the file was manually edited; the importer preserves the local content, marks `dirty=True` in both state and front matter, and skips
5. On write: stamps a `polylogue:` front matter block, writes Markdown + optional HTML, updates conversation state, and calls `update_index()`

The dirty-local path deserves a code snippet because the logic is non-obvious:

```python
# If the user edited the file locally (body_hash != stored contentHash)
# AND the remote content hasn't changed, preserve local edits
if remote_same and existing:
    dirty_meta = _ensure_polylogue_metadata(existing.metadata)
    polylogue = dirty_meta.get("polylogue", {})
    polylogue["dirty"] = True
    polylogue["localHash"] = existing.body_hash
    _write_markdown(markdown_path, dirty_meta, existing.body)  # keeps their edits
    return DocumentPersistenceResult(skipped=True, skip_reason="dirty-local", ...)
```

## Automation framework

`automation_targets.json` defines five targets with their CLI commands and default arguments:

```json
{
  "codex": {
    "name": "polylogue-sync-codex",
    "command": ["sync-codex", "--plain", "--json"],
    "defaults": { "collapseThreshold": 24, "html": false }
  },
  ...
}
```

`automation.py` reads this registry and exposes `systemd_snippet()` and `cron_snippet()` that merge target defaults with user overrides to produce ready-to-paste scheduler entries. The CLI surfaces these as `polylogue automation systemd --target codex --interval 10m`.

The NixOS module (`nix/modules/polylogue.nix`) reads the same JSON at Nix eval time, validates target names, and generates systemd service/timer pairs. Users get:

```nix
services.polylogue = {
  enable = true;
  targets.codex = { enable = true; timer.interval = "10m"; };
  targets."claude-code".enable = true;
};
```

## File watchers

`polylogue watch codex` and `polylogue watch claude-code` use the `watchfiles` library to monitor session directories. Changes are debounced (default 2s), and only `.jsonl` file changes trigger a sync. `--once` runs a single sync pass and exits, useful for testing.

Watch runs log to the same `runs.json` telemetry as scheduled runs, so `polylogue status --json --watch` can monitor both watchers and cron jobs from a single dashboard.

## Indexing

Every `persist_document()` call fires `update_index()`, which dispatches to the configured backend:

- **sqlite** (default): maintains a `conversations` table with content hash, token count, and dirty flag, plus an FTS5 virtual table for full-text search
- **qdrant**: upserts into a Qdrant collection (requires `POLYLOGUE_QDRANT_URL`)
- **none**: no-op

The SQLite path is lightweight enough to run on every write without noticeable overhead.

## Packaging

- Added `pyproject.toml` with `polylogue` console_scripts entry point
- Rewrote `flake.nix` to use `buildPythonApplication` and export `packages`, `apps`, `devShells.ci`, and `nixosModules`
- Added `pyperclip` and `watchfiles` to the dev shell

## Changes grouped by concern

**Importers**: All four importers (chatgpt, claude_ai, claude_code, codex) call `persist_document()` instead of writing files directly. `ImportResult` gains `skipped`, `skip_reason`, `dirty`, and `content_hash` fields. The CLI summary display accounts for skipped conversations.

**State management**: `util.py` gains `assign_conversation_slug()` for stable provider-scoped slug allocation, `conversation_is_current()` for skip detection, and `update_conversation_state()` for per-conversation state tracking.

**Status command**: Gains `--json`, `--watch`, and `--interval` flags. A new provider-level summary aggregates stats across command types (e.g., both `sync` and `sync-codex` roll up under their respective providers). Duration tracking added to all run records.

**Clipboard**: `--to-clipboard` flag on render and all import subcommands copies single-file output. Drive credential setup checks the clipboard for an OAuth JSON payload before prompting for a file path.

**Documentation**: `multi-provider.md` deleted; content distributed into five focused docs under `docs/providers/`. README updated with watcher usage, automation targets, NixOS module example, and indexing configuration.

## Test coverage

- `test_automation_snippets.py`: systemd/cron generation, describe roundtrip, CLI dispatch
- `test_commands_logic.py`: render/sync through persistence layer with SQLite index verification, skip-on-rerun behavior
- `test_importers_multi.py`: repeat-import skip detection with dirty-local scenario
- `test_status.py`: provider summary aggregation from run records

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ Polylogue is an interactive-first toolkit for archiving AI/LLM conversations—r
 - **Render local logs:** Choose a file or directory; skim previews candidates, rich shows progress, and outputs land in the configured render directory (default `/realm/data/chatlog/markdown/gemini-render`). Add `--html` for themed previews or `--diff` to see deltas when re-rendering.
 - **Sync Drive folders:** Connect to the default Drive folder (`AI Studio`) and pull chats to Markdown in `/realm/data/chatlog/markdown/gemini-sync`, downloading attachments unless you opt to link only.
 - **Sync Codex / Claude Code sessions:** Mirror local CLI transcripts from `~/.codex/sessions/` and `~/.claude/projects/` via `polylogue sync-codex` / `polylogue sync-claude-code`, with optional JSON summaries, pruning, diffs, and HTML previews. Outputs land in `/realm/data/chatlog/markdown/codex` and `/realm/data/chatlog/markdown/claude-code` by default.
+- **Watch local sessions in real time:** `polylogue watch codex` and `polylogue watch claude-code` keep those directories synced automatically; adjust debounce, HTML, and collapse settings per watcher.
 - **Import exported providers:** Convert ChatGPT zips, Claude exports, Claude Code sessions, or Codex JSONLs via `polylogue import …` subcommands. Skim lets you cherry-pick conversations; `--all` batches them.
 - **Doctor & Stats:** `polylogue doctor` sanity-checks source directories; `polylogue stats` aggregates attachment sizes, token counts, and provider summaries (with `--since/--until` filters).
 - **View recent runs:** The status dashboard shows the last operations, including attachment MiB and diff counts per command.
@@ -54,6 +55,7 @@ Although the CLI is interactive by default, the same functionality is available 
 - `python3 polylogue.py stats [--dir DIR] [--since DATE] [--until DATE] [--json]`
 
 `--plain` disables gum/skim/Rich styling for CI or scripts; `--json` prints machine-readable summaries.
+Use `--to-clipboard` on `render`/`import` commands to copy a single Markdown result directly to your system clipboard.
 
 ## Tooling & UX Stack
 The dev shell equips Polylogue with:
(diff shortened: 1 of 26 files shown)


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 4 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 9 finding(s) confirmed fixed/stale.

PR #12, opened 2026-04-07
refactor: move Drive sync to requests-based REST transport

Sinity:
## Summary

I replaced the Google Drive client-library integration with direct Drive v3 REST calls built on `requests` and `AuthorizedSession`. The main effect is a much smaller and more explicit transport layer: `polylogue/drive.py` no longer depends on `googleapiclient.discovery.Resource`, `HttpError`, or `MediaIoBaseDownload`, and the packaging story no longer needs the full Drive client stack. Fixed the fallout that surfaced while making that swap, including package discovery for `polylogue.*`, typed stats filtering during builds, and a console-auth fallback for environments where `run_console()` is absent. The end result is still the same user-facing Drive feature set, but the implementation is leaner and easier to reason about.

## Motivation

The old Drive path hid a lot of complexity behind the generated Google client objects. That is convenient initially, but it couples the project to a large dependency surface, awkward packaging requirements, and library-specific exceptions/helpers. It also makes relatively simple operations — list files, get metadata, download media — look more magical than they need to be.

The branch history shows that I explored a packaging-only mitigation first. I dropped `google-generativeai`, then briefly split Drive dependencies into optional extras / separate Nix outputs. That reduced the default footprint a bit, but it did not address the deeper problem: we were still carrying the big client-library stack just to make a handful of Drive v3 calls. Once I switched the transport to direct REST requests, the separate “drive extras” experiment stopped being worth the cognitive overhead, so the branch collapses back to a simpler core dependency set.

The auth compatibility issue was another motivator. Relying on `run_console()` as though it were always present is brittle across auth library versions. Since I was already rewriting the transport, I took the chance to make console OAuth more explicit and resilient.

### 1. Replace generated client objects with explicit REST helpers

`polylogue/drive.py` builds an `AuthorizedSession` and uses it to call the Drive REST API directly. The important pieces are:

- `_drive_get_json()` for GET + JSON parsing
- `_raise_for_status()` for normalized HTTP failure handling
- `list_children()` / `find_folder_id()` / `get_file_meta()` built on those helpers
- `download_file()` implemented via streaming `iter_content()`

Before, the code looked conceptually like this:

```python
service.files().list(...).execute()
service.files().get(...).execute()
MediaIoBaseDownload(...)
```

Now it is closer to:

```python
session.get(f"{DRIVE_BASE}/files", params=...)
session.get(f"{DRIVE_BASE}/files/{file_id}", params=...)
session.get(..., params={"alt": "media"}, stream=True)
```

That makes pagination, metadata fields, retry boundaries, and download streaming visible in our own code instead of being implicit in the client library.

### 2. Normalize errors and auth compatibility

I introduced `DriveApiError` so the rest of the code does not have to know about `HttpError`. The new exception carries a message, status code, and payload, which is enough for the transport layer and substantially easier to test/reason about later.

Added a manual console-flow fallback for auth stacks where `run_console()` is missing. That keeps the existing “authenticate in a terminal” workflow alive without pinning Polylogue to one exact auth-library version.

`DriveClient.service` catches transport/runtime failures from the new layer and translates them into user-facing banners / `SystemExit`, which keeps the interactive UX in line with the rest of the tool.

### 3. Simplify packaging around the new transport

`pyproject.toml`, `flake.nix`, and `nix/devshell.nix` all change here. The net effect is:

- remove `google-api-python-client`
- remove `google-auth-httplib2`
- remove `google-generativeai`
- add `requests`
- keep `google-auth-oauthlib` for OAuth token handling

Collapsed the short-lived optional-drive packaging split that appeared mid-branch. Once Drive transport only depends on auth + requests, the extra complexity of separate `polylogue-with-drive` outputs and a dedicated drive shell no longer pays for itself.

Finally, I fixed the wheel packaging issue by broadening `setuptools` discovery from `polylogue` to `polylogue.*`; otherwise the importer subpackages disappear in built wheels.

### 4. Opportunistic fixes exposed by the refactor

Two smaller fixes landed because the transport/package rewrite made them visible:

- `parse_input_time_to_epoch()` accepts numeric/date/datetime inputs, which fixes stats filtering failures in Nix build contexts that pass typed values.
- the stats front-matter parser in `cli.py` falls through on failure instead of returning an empty dict for the whole aggregation.

These are not Drive features, but they were genuine regressions/paper cuts surfaced by running the tighter build path.

## Testing

This branch does not add dedicated new tests for the REST transport itself. The existing pytest suite still runs via the flake build, and I fixed package discovery and stats parsing so those existing tests/package builds keep passing. The obvious gaps remain:

- no transport-level tests for REST pagination or error payload mapping
- no auth compatibility tests for missing `run_console()`
- no large-file streaming download tests

Those are good follow-ups once the REST surface settles.

## Notes

- `get_drive_service()` is intentionally a session factory, not a Google Resource factory. Any internal code that expected `service.files()` semantics has to stay on the helper functions in `polylogue/drive.py`.
- The optional-drive packaging experiment was useful as an interim checkpoint, but the final branch lands in a simpler place than it started.
- The README no longer needs to describe Drive extras because Drive support is back in the core package with a much smaller dependency footprint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ Polylogue is an interactive-first toolkit for archiving AI/LLM conversations—r
 
 ## Quick Start
 - Enable the direnv-managed dev shell: `direnv allow` (uses `.envrc` to call `nix develop`).
+- Drive integrations require optional Google client libraries; install them with `pip install "polylogue[drive]"`, run `nix develop .#drive`, or use the `polylogue-with-drive` package output.
 - Prefer manual entry? `nix develop` installs Python plus gum, skim, rich, bat, glow, etc.
 - Run `python3 polylogue.py` and pick an action from the gum menu (Render, Sync, Local Syncs, Doctor, Stats, etc.).
 - When a directory has multiple JSON logs, the skim picker previews files with `bat`; press `Ctrl+G` for a live `glow` render before confirming.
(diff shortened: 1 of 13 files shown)


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 5 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #16, opened 2026-04-07
refactor: split monolithic CLI into command modules with XDG-compliant paths

Sinity:
## Summary

The single-file CLI (`polylogue/cli.py`, 1835 lines) is decomposed into a `polylogue/cli/` package with eleven focused modules. Every hardcoded path is replaced with XDG base directory lookups, the per-conversation output layout migrates from flat files to directories, and two new inspection commands (`inspect branches`, `inspect search`) land as first-class CLI surfaces.

## Motivation

The old CLI module had become the project's junk drawer: sync logic for three providers, four import flows, the interactive menu, automation snippet generation, stats aggregation, and the doctor command all lived in one file. Adding the branch explorer and search features would have pushed it past 2500 lines with no clean separation of concerns.

Separately, every output path was rooted at `/realm/data/chatlog/`, which is a personal NixOS mount point that doesn't exist on any other machine. The `_attachments/` naming convention also conflicted with tools that treat underscore-prefixed directories specially (Jekyll, some static site generators).

I considered splitting by "concern" (commands vs. rendering vs. state) but settled on splitting by "CLI surface" instead — each subcommand family gets its own module, with shared defaults in `context.py`. This keeps the import graph shallow and makes it obvious where to add flags for a given command.

## CLI architecture

The new package layout:

```
polylogue/cli/
├── __init__.py          # public API shim for backwards compat
├── app.py               # parser, main(), interactive menu, inspect commands
├── context.py           # defaults, HTML resolution, settings helpers
├── sync.py              # sync drive|codex|claude-code
├── imports.py           # import chatgpt|claude|codex|claude-code
├── render.py            # render command + clipboard helper
├── watch.py             # provider-agnostic watcher
├── status.py            # status dashboard + stats
├── doctor.py            # health checks
├── automation_cli.py    # systemd/cron snippets
└── summaries.py         # OperationSummary bridge to results.py
```

Each module imports from the domain layer (`commands.py`, `pipeline.py`, etc.) rather than reaching across to other CLI modules. `context.py` owns all the computed defaults so no two modules independently read `CONFIG.defaults`.

## Command surface changes

The top-level sync commands collapse into a single `sync <provider>` with a positional argument:

```
# Before                              # After
polylogue sync-codex --html           polylogue sync codex --html
polylogue sync-claude-code --prune    polylogue sync claude-code --prune
polylogue sync --folder-name X        polylogue sync drive --folder-name X
```

The `--html` flag changes from a boolean to a tri-state (`on|off|auto`) so automation scripts can explicitly disable HTML without knowing the config default, while interactive users get auto-detection from their settings.

New commands:
- `inspect branches` — tree view, HTML explorer, unified diff against canonical
- `inspect search` — FTS5 query with provider/model/date/attachment filters
- `inspect stats` — relocated from top-level `stats`
- `prune` — removes legacy `slug.md` + `slug_attachments/` left by older releases

## XDG and output layout

All default paths derive from XDG base directories:

| Variable | Default |
|---|---|
| `CONFIG_HOME` | `~/.config/polylogue` |
| `DATA_HOME` | `~/.local/share/polylogue` |
| `STATE_HOME` | `~/.local/state/polylogue` |
| `CACHE_HOME` | `~/.cache/polylogue` |

The archive root moves from `/realm/data/chatlog/` to `$XDG_DATA_HOME/polylogue/archive`, overridable via `POLYLOGUE_ARCHIVE_ROOT`.

Per-conversation output migrates from flat files to directories:

```
# Before                      # After
out/my-chat.md                out/my-chat/conversation.md
out/my-chat_attachments/      out/my-chat/attachments/
out/my-chat.html              out/my-chat/conversation.html
                              out/my-chat/branches/branch-001/branch-001.md
```

This layout naturally groups conversation artifacts and makes branch directories siblings of the canonical transcript.

## Extracted domain modules

**`pipeline.py`** — `ChatContext` and `build_document_from_chunks()` extracted from `commands.py`. The attachment collector checks remote `modifiedTime` against local mtime before re-downloading, which avoids redundant Drive API calls on incremental syncs.

**`repository.py`** — thin facade over `document_store.persist_document()` so callers (commands.py, conversation.py) share a single entry point with consistent signatures.

**`results.py`** — `OperationSummary` replaces the inline summary-building logic that was duplicated across sync, import, and render. `summarize_import_results()` aggregates attachments, tokens, words, diffs, and branches into a single object with a `.lines()` method for display.

**`search.py`** — executes FTS5 queries against the `messages_fts` virtual table, applying provider/slug/branch/model/date/attachment filters. Post-filter on timestamp and model happens in Python because SQLite FTS5 doesn't support those natively in the MATCH clause.

**`branch_explorer.py`** — loads branch metadata from SQLite, renders ASCII trees, generates themed HTML explorers, and computes unified diffs between branches.

## Infrastructure changes

- **`StateStore`** in util.py replaces the load/save pattern with a `mutate(fn)` method that reads, applies a function, and writes atomically
- **`DiffTracker`** encapsulates the snapshot-before/diff-after pattern that was repeated in every command
- **`RunAccumulator`** replaces `defaultdict(int)` totals scattered across render, sync, and import
- **`document_store.py`**: content hash now includes metadata (excluding the hash itself), so metadata-only changes are detected; stale attachments are cleaned up on re-persist; HTML files are removed when HTML is disabled
- **`index.py`**: `IndexBackend` protocol with three implementations behind `POLYLOGUE_INDEX_BACKEND`
- **`safe_extractall`** in importers/utils.py validates every zip member path against the target directory to block path traversal
- **`html.py`**: Jinja template and markdown-it renderer are cached at module level; all user-supplied content passes through `markupsafe.escape`

## Nix packaging

The flake extracts `re2Overlay` to a top-level binding, adds `cliDeps` (gum, skim, delta, git) as both build and check inputs, and wraps the installed binary with `makeWrapper` so CLI tools are always on PATH. The check phase sets `XDG_STATE_HOME` and `XDG_CACHE_HOME` to `$TMPDIR` so tests don't pollute the builder's home directory. `tiktoken` and `qdrant-client` are added to `baseDeps`.

## Testing

Eight new test files cover the branch explorer, search filtering, CLI flag parsing, CLI wrapper dispatch, document store cleanup, Drive credential env detection, zip path traversal blocking, and the pipeline attachment collector. A `tiktoken` stub in conftest.py provides deterministic word-split tokenization so tests don't need the real BPE model. `pyproject.toml` gains pytest, coverage, and mypy configuration.


Sinity:
@coderabbitai review

PR #18, opened 2026-04-07
feat: add SQLite persistence layer, pipeline framework, and ingestion registrar

Sinity:
## Summary

The core persistence story changes from scattered JSON files to a single SQLite database, and the render/sync/import paths gain a composable pipeline abstraction that makes it simple to add stages (validation, telemetry, caching) without touching the per-provider business logic. ChatGPT and Claude.ai exports also become first-class local sync providers alongside Codex and Claude Code.

## Motivation

Three things were pulling in the same direction:

1. **JSON state was brittle.** `state.json` was a single dict mutated by every command — no transactions, no concurrent-safe writes, no way to query "which conversations have attachments?" without loading the whole file and scanning. `runs.json` grew unbounded and provided no aggregation support. Moving to SQLite gives us ACID writes, FTS5 for full-text search, and structured queries that doctor/status/search actually need.

2. **Render and sync duplicated pipeline logic.** Both paths independently loaded JSON, validated chunks, collected attachments, rendered Markdown, persisted state, and wrote diffs. Factoring this into Pipeline stages means a change to persistence (like adding content hashing) happens once in `RenderPersistStage` rather than separately in `render_command`, `sync_command`, and every importer.

3. **ChatGPT/Claude exports were second-class.** Import was a one-shot operation with no watch mode and no sync integration. Treating them as local providers — with a base directory under `$XDG_DATA_HOME/polylogue/exports/` — means the same `sync`/`watch`/`--branch-export` flags work everywhere.

I considered keeping the JSON files alongside SQLite as a migration bridge, but the state.json format was already provider-keyed in a way that doesn't map cleanly to the relational schema, and maintaining both would double the surface area for bugs. Clean cut felt safer.

## Schema v2

The schema bump drops denormalized stats from `conversations` and `branches` into `metadata_json`, which avoids schema migrations every time we want to track a new metric. The key structural changes:

```sql
-- conversations: stats move to metadata_json
ALTER TABLE conversations ADD COLUMN root_message_id TEXT;
-- removed: token_count, word_count, attachment_bytes, dirty

-- branches: is_canonical → is_current, stats → metadata_json
ALTER TABLE branches ADD COLUMN label TEXT;
ALTER TABLE branches ADD COLUMN is_current INTEGER;
-- removed: message_count, token_count, word_count, timestamps

-- messages: body → rendered_text, new columns
ALTER TABLE messages ADD COLUMN content_hash TEXT;
ALTER TABLE messages ADD COLUMN raw_json TEXT;
ALTER TABLE messages RENAME COLUMN body TO rendered_text;

-- new tables
CREATE TABLE runs (...);      -- replaces runs.json
CREATE TABLE state_meta (...); -- key-value store
```

The FTS5 virtual table now indexes `rendered_text` as `content`:

```sql
CREATE VIRTUAL TABLE messages_fts USING fts5(
    provider, conversation_id, branch_id, message_id,
    content, tokenize='unicode61'
);
```

## Pipeline Architecture

The `Pipeline` runner is intentionally minimal — a list of stages, a shared context, and abort propagation:

```python
class Pipeline:
    def run(self, context: PipelineContext) -> PipelineContext:
        for stage in self._stages:
            if context.aborted:
                break
            stage.run(context)
        return context
```

Each flow composes the stages it needs. Render:

```python
Pipeline([
    RenderReadStage(),       # JSON → raw payload
    RenderNormalizeStage(),   # validate + extract chunks
    RenderDocumentStage(),    # collect attachments, build links
    RenderPersistStage(),     # write markdown, update registrar
])
```

Drive sync reuses the last two stages but swaps in download/normalize stages:

```python
Pipeline([
    DriveDownloadStage(),    # fetch bytes from Drive
    DriveNormalizeStage(),   # validate + extract chunks
    RenderDocumentStage(),   # same as render
    RenderPersistStage(),    # same as render
])
```

Import flows use a different pattern — `ImportExecuteStage` wraps the provider-specific import function and `ImportSummarizeStage` handles the UI summary output.

## ConversationRegistrar

The registrar is the single write coordinator. Instead of each command independently calling `upsert_conversation`, `replace_branches`, `replace_messages`, and `update_conversation_state`, they all go through:

- `registrar.record_document(...)` — persists state entry + conversation row after markdown write
- `registrar.record_branch_plan(...)` — writes branches, messages, and FTS entries
- `registrar.mark_dirty(...)` — flags a conversation as locally modified
- `registrar.get_state(...)` — reads the current state entry

The document_store, conversation processor, and search all share a consistent view.

## Local Sync Providers

The `LocalSyncProvider` dataclass unifies the four providers:

```python
@dataclass(frozen=True)
class LocalSyncProvider:
    name: str
    sync_fn: LocalSyncFn
    default_base: Path
    default_output: Path
    list_sessions: Callable[[Path], List[Path]]
    watch_suffixes: Tuple[str, ...] = (".jsonl",)
    supports_watch: bool = True
    supports_diff: bool = True
    create_base_dir: bool = False
```

ChatGPT and Claude exports use `watch_suffixes=(".zip", ".json")` and `create_base_dir=True` since the export directory might not exist yet. The `_sync_export_bundles` helper discovers ZIPs and `conversations.json` directories automatically.

## Branch Export Modes

`--branch-export canonical` is useful when you want a flat archive without branch directories. `overlay` is the middle ground — you get the canonical transcript plus per-branch overlay.md files showing only the divergent turns, but without the full per-branch document copies that `full` mode writes. When switching modes, stale artifacts (common.md, branch directories) are cleaned up.

## Drive Observability

`DriveMetrics` tracks requests, retries, and failures per operation type (metadata, download, chat, attachment). `snapshot_drive_metrics(reset=True)` captures the counters and resets them at the end of each sync, writing the stats into the run record. Doctor now surfaces retry/failure rates from recent runs.

## Testing

- `test_pipeline_runner.py` exercises stage ordering, abort, error recording, and context.require
- `test_conversation_registrar.py` verifies branch metadata, message content hashes, and tool call persistence
- `test_snapshot_flows.py` provides end-to-end render→rerender and sync→resync scenarios with HTML and diff output
- `test_validation.py` covers schema validation for Gemini payloads
- `test_watch_cli.py` tests suffix filtering and debounce behavior
- `test_settings_cli.py` verifies persistence and reset of settings.json
- `test_drive_client.py` adds coverage for redirect URI assignment, cached tokens, and retry/failure metrics
- Existing test suites updated to use SQLite-backed state via `conftest._configure_state`

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added support for ChatGPT and Claude export workflows via `polylogue sync` and `polylogue watch` commands
  * Introduced `--branch-export` flag controlling conversation output layout (full/overlay/canonical)
  * New `polylogue settings` command for managing HTML preview mode and theme preferences
  * Enhanced `polylogue status` with data export and filtering capabilities
  * Added status reporting options to automation workflows

* **Improvements**
  * Migrated conversation state storage to SQLite database
  * Added Drive operation retry/failure metrics to status reports


Sinity:
@coderabbitai full review

PR #20, opened 2026-04-07
feat: shell completions, dashboards, and doctor improvements

Sinity:
## Summary

Rework the CLI from an interactive-menu-first tool into a command-first surface with proper shell completions, richer diagnostics, and automatic plain-mode detection. Along the way, retire the automation snippet generator and NixOS module, simplify branch export to always-full, and add several new subcommands for operational visibility.

## Motivation

The interactive menu was the original entry point, but in practice every workflow was already reachable via `polylogue <command>`. The menu added ~400 lines of prompt/handler code that duplicated the argument parsing, and it made the tool harder to script—you had to remember `--plain` for every non-interactive invocation. Meanwhile, shell completions were missing entirely, so tab-completing provider names or slugs required memorization or `--help`.

The `--branch-export full|overlay|canonical` flag added cognitive load without much practical value: almost every user wanted `full`, and the `overlay`/`canonical` modes were undertested edge cases. Dropping them simplifies every importer, the render path, and the sync flow.

The automation module (`polylogue automation systemd|cron`) and NixOS module were speculative infrastructure—useful in theory but not battle-tested and carrying a maintenance cost (the NixOS module alone was 226 lines of Nix). Removing them now lets us focus on the CLI ergonomics that actually get used daily.

### Shell completions

The `completions --shell bash|zsh|fish` command emits a sourcing script. Bash and fish get static `compgen`/`complete` scripts. Zsh gets a dynamic completion function that calls back into `polylogue _complete --shell zsh --cword N -- words…`, which routes through `CompletionEngine`:

```python
class CompletionEngine:
    def complete(self, shell, cword, words) -> List[Completion]:
        # Routes to per-command completers that return live data:
        # - slug/conversation-id lookups from registrar state
        # - Drive chat IDs from extra_state metadata
        # - session file listings from provider.list_sessions()
        # - argparse option strings with help text
```

The devshell hook (`nix/devshell.nix`) auto-generates and sources the script for bash, zsh, or fish on `nix develop` entry.

### Automatic plain-mode detection

`--plain` is replaced by TTY detection in `_should_use_plain()`:

```python
def _should_use_plain(force_interactive: bool) -> bool:
    if force_interactive:
        return False
    if os.environ.get("POLYLOGUE_FORCE_PLAIN", "").lower() in _FORCE_PLAIN_VALUES:
        return True
    return not (sys.stdout.isatty() and sys.stderr.isatty())
```

`--interactive` overrides the detection for headless shells where you still want gum/skim prompts. The `InteractiveConsoleFacade` now raises `RuntimeError` (caught as `SystemExit`) when required tools are missing, without silently falling back—since the devshell guarantees them, a missing tool means you're outside the shell and should know about it.

### New commands: dashboards, runs, env, help, index check

- **`dashboards`** renders Rich tables of provider health (runs/items/attachments/failures) and the last N runs. `--json` emits the same data as JSON.
- **`runs`** is the filtered run-history viewer—`--providers`, `--commands`, `--since`/`--until`, `--json`.
- **`env`** dumps resolved configuration: config path, collapse threshold, HTML settings, all output directories, state DB path, runs DB path, watchable providers.
- **`help [COMMAND]`** prints the command's argparse help, or a command-listing table when no topic is given.
- **`index check`** runs `verify_sqlite_indexes()` (integrity check + FTS table presence) and optionally `verify_qdrant_collection()`. `--repair` rebuilds missing FTS tables.

### Doctor enhancements

Doctor runs five additional check categories:

1. **Dependency checks** — verifies gum, sk, bat, glow, delta are in PATH
2. **Config file check** — reports missing config with candidate paths
3. **Credential checks** — reports missing credentials.json and token.json
4. **SQLite index validation** — runs integrity check and FTS presence
5. **Qdrant reachability** — probes collection when backend is `qdrant`

Every `DoctorIssue` carries an optional `hint` string, and the Rich table output uses severity-colored rows.

### Config validation at startup

`polylogue/core/config_validation.py` defines Pydantic models (`ConfigModel`, `DefaultsModel`, `OutputDirsModel`) with `extra = "forbid"`. The config is validated on load, so typos like `html_theme: "purple"` or `collapse_threshold: "fast"` produce an actionable `SystemExit` instead of silent misbehavior.

### Progress bars and retry UX

Drive sync and local-sync loops wrap their iteration in `ui.progress()`, which renders a Rich progress bar (spinner + bar + count + elapsed + remaining) in interactive mode and is a no-op in plain mode. The Drive retry path accepts a `notifier` callback that prints yellow retry messages to the console.

### Removed: automation, NixOS module, branch-export, interactive menu, structured run logs

| Removed | Lines | Reason |
|---------|-------|--------|
| `automation.py` + `automation_targets.json` | ~340 | Speculative; not battle-tested |
| `nix/modules/polylogue.nix` | 226 | Speculative NixOS module |
| `cli/automation_cli.py` | 79 | CLI for above |
| Interactive menu + prompt_* handlers | ~400 | Redundant with CLI commands |
| `--branch-export` flag | across 15 files | Always-full simplifies all paths |
| `POLYLOGUE_RUN_LOG` stderr emission | ~30 | Run history lives in SQLite now |

### Provider registry

New `polylogue/providers/` package introduces `ProviderSession` protocol, `ProviderRegistry`, and `DriveProviderSession` (wraps `DriveClient` with `__getattr__` delegation). `CommandEnv` holds a `ProviderRegistry` and a `drive_constructor` factory, and `_ensure_drive()` registers the session so downstream code can look up providers by name.

### Vendor fallbacks

`polylogue/_vendor/` provides minimal fallback implementations for `frontmatter`, `pyperclip`, and `watchfiles` so the import paths are always valid. The `frontmatter` vendor is a ~65-line YAML front-matter parser; `pyperclip` is an in-memory stub; `watchfiles` raises `RuntimeError` on call (explicit failure instead of silent no-op).

## Testing

- New test files: `test_cli_help_env.py`, `test_dashboards.py`, `test_runs_cli.py`, `test_index_cli.py`, `test_index_health.py`, `test_config_validation.py`
- Updated: `test_cli_flags.py` (Drive sync selection, local sync sessions), `test_status.py` (json-lines), `test_util_runs.py` (latest_run, format_run_brief), `test_watch_cli.py` (error surfacing)
- Removed: `test_automation_snippets.py`, branch-export-mode tests from `test_branch_and_search.py` and `test_render_branch.py`


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 5 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 9 finding(s) confirmed fixed/stale.

PR #24, opened 2026-04-07
refactor: consolidate CLI surface, add dynamic completions, and NixOS module

Sinity:
## Summary

Reshapes the CLI from a flat list of seventeen commands into a grouped hierarchy (`sync`, `import`, `browse`, `maintain`, `config`), replaces static shell completion scripts with a dynamic engine that queries live state, and introduces a declarative NixOS module that generates systemd timers from a JSON target manifest. Along the way, the ChatGPT/Claude importers switch to streaming JSON parsing and the SQLite connection manager gains per-thread reuse.

## Motivation

The command surface had grown organically—`watch`, `status`, `dashboards`, `settings`, `env`, `inspect`, `prune`, `doctor`, `index` all lived at the top level, making `--help` output noisy and completions unreliable. Static completion scripts could only complete the first positional argument; subcommand flags, provider names, slugs, and branch IDs all required manual typing.

On the packaging side, the `_vendor/` shims for frontmatter, pyperclip, and watchfiles existed from before those dependencies were declared in `pyproject.toml`. They added confusion—tests could pass against the vendor stubs while production used the real packages—and made dependency auditing harder.

The ChatGPT and Claude importers loaded entire `conversations.json` files into memory via `json.loads()`, which worked fine for small exports but became a problem with archives containing thousands of conversations.

### CLI command reorganisation

The watch command folds into `sync --watch`—the dispatcher in `_dispatch_sync` checks the flag and delegates to `run_watch_cli` for local providers that support it. This eliminates a separate parser and keeps watch-mode flags (`--debounce`, `--once`) contextual:

```python
def _dispatch_sync(args, env):
    if getattr(args, "watch", False):
        provider = get_local_provider(args.provider)
        if not provider.supports_watch:
            raise SystemExit(f"{provider.title} does not support watch mode")
        run_watch_cli(args, env)
    else:
        run_sync_cli(args, env)
```

The `dashboards` command is removed entirely—its provider-summary table was redundant with `browse status`. Settings split into `config set` / `config show` / `config init`, and `env` becomes `config show --json`. A legacy `inspect` dispatcher is retained for backwards compatibility but prints a deprecation-style nudge toward `browse`.

### Dynamic shell completions

`CompletionEngine` is rewritten with per-subcommand completers (`_complete_browse`, `_complete_maintain`, `_complete_config`, `_complete_search`) that return contextual suggestions. Three new helpers query the conversation state repository:

- `_provider_completions()` — distinct providers from persisted state
- `_branch_completions()` — branch IDs across all conversations
- `_model_completions()` — model names from conversation metadata

Bash and Fish scripts call `polylogue _complete --shell bash --cword $COMP_CWORD` at runtime instead of hardcoding command lists. Zsh was already dynamic; the new bash/fish scripts follow the same pattern.

### NixOS module

`nix/modules/polylogue.nix` is a 307-line NixOS module that reads `polylogue/automation_targets.json` to generate systemd service+timer pairs. Each target (codex, claude-code, drive-sync, gemini-render, chatgpt-import) gets:

- A `services.polylogue.targets.<name>` option set with `enable`, `outputDir`, `timer.interval`, `html`, `collapseThreshold`, `extraArgs`, and per-target `environment`
- A generated systemd oneshot service with `POLYLOGUE_FORCE_PLAIN=1` and XDG paths
- A persistent timer with configurable `OnBootSec` and `OnUnitActiveSec`
- tmpfiles rules for output directories
- An activation script writing `config.json` from Nix-declared settings

`nix/python-deps.nix` centralises `commonDeps` and `devDeps` so `flake.nix`, `nix/devshell.nix`, and the NixOS module all reference the same list. The flake itself restructures around a `perSystem` pattern to cleanly separate per-system outputs from the system-independent `nixosModules`.

### Streaming JSON ingestion

ChatGPT's importer replaces `json.loads(convo_path.read_text())` with `ijson.items(fh, "item")`, yielding one conversation dict at a time. Claude's importer introduces `_iter_conversations()` that tries `ijson.items(fh, "conversations.item")` first (for the `{conversations: [...]}` wrapper) then falls back to `ijson.items(fh, "item")` for bare lists. Both importers preserve the same error messages but chain exceptions properly.

### Database connection reuse

`open_connection()` stores the active connection in `threading.local()` and tracks nesting depth. Nested calls to `open_connection()` with the same path return the same `sqlite3.Connection`; mismatched paths raise `RuntimeError`. The outermost context manager commits and closes. WAL journal mode is enabled on first open.

### Branch explorer batching

`_load_branch_overview` replaces N per-branch `_fetch_divergence_message` calls with a single query that fetches all messages for all branches, then indexes divergence messages by position in Python. The HTML generator switches from f-string interpolation with manual escaping to a Jinja2 `Environment(loader=BaseLoader())` template with `select_autoescape`.

### Prune safety

Both `_sync_sessions` and `_sync_export_bundles` now rename targets into a `.trash/` subdirectory before calling `shutil.rmtree`, so an interrupted prune leaves the data recoverable rather than half-deleted.

### Vendor removal

`_vendor/frontmatter.py`, `_vendor/pyperclip.py`, and `_vendor/watchfiles.py` are deleted. All imports switch to the real packages, which were already declared in `pyproject.toml`. Pydantic models in `config_validation.py` switch from inner `class Config` to `model_config = ConfigDict(extra="forbid")`.

## Testing

Seven new test modules:
- `test_attachment_failures.py` — verifies `collect_attachments` raises on download failure
- `test_branch_explorer_batching.py` — seeds branch/message data and asserts batched divergence lookup
- `test_cli_error_messages.py` — exercises `_dispatch_inspect` and `_dispatch_config` with missing subcommands
- `test_db_connection_reuse.py` — confirms connection identity across nested opens and mismatch rejection
- `test_importers_streaming.py` — creates 500-entry exports and streams them through both importers
- `test_prune_trash.py` — verifies rename-to-trash-then-delete flow
- `test_settings_persistence.py` — round-trips settings through `persist_settings_to` and `_load_settings_file`


Sinity:
@coderabbitai review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 6 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 5 finding(s) confirmed fixed/stale.

PR #26, opened 2026-04-07
refactor: extract drive, index, and exports into typed config objects

Sinity:
## Summary

Provider-specific settings — Drive OAuth credentials, Qdrant vector index connection, ChatGPT/Claude export directory roots — were resolved through ad-hoc environment variable lookups and hardcoded `Path` computations scattered across `drive.py`, `util.py`, and the NixOS module. This made them invisible to `config show`, impossible to validate early, and awkward to expose as NixOS options. This PR promotes all three into typed config objects with Pydantic validation, wires them through the config loader, and expands both the NixOS module and CLI to surface them.

## Motivation

The immediate trigger was the NixOS module: adding `qdrant` and `drive` options required structured config sections, not just raw environment variables stitched together at service-unit level. But the underlying problem was broader — `_retry()` in `drive.py` was parsing `POLYLOGUE_RETRIES` and `POLYLOGUE_RETRY_BASE` from the environment on every call, `util.py` was computing `CODEX_SESSIONS_ROOT` and `CLAUDE_CODE_PROJECT_ROOT` at import time with no fallback chain, and export directories were hardcoded in the README without any config surface.

The alternative was to keep environment variables as the primary config channel and just map NixOS options to env vars. I went with proper typed objects because:

1. Pydantic validation catches bad config at load time, not when a sync job runs at 3 AM
2. `config show` can emit the full resolved state, so operators don't have to reconstruct what the loader actually computed
3. The converter pattern (`_convert_drive`, etc.) keeps the core config module free of facade dependencies

## Config layer design

Three-tier structure for each new section:

```python
# Core dataclass (configuration.py) — the source of truth
@dataclass
class DriveConfig:
    credentials_path: Optional[Path] = None
    token_path: Optional[Path] = None
    retries: Optional[int] = None
    retry_base: Optional[float] = None

# Pydantic model (config_validation.py) — rejects bad input early
class DriveModel(BaseModel):
    credentials_path: Optional[str] = None
    token_path: Optional[str] = None
    retries: Optional[int] = Field(default=None, ge=1)
    retry_base: Optional[float] = Field(default=None, ge=0)
    model_config = ConfigDict(extra="forbid")

# Facade converter (config.py) — bridges core → public Config
def _convert_drive(core: Optional[CoreDriveConfig]) -> Optional[DriveConfig]:
    ...
```

`IndexConfig` follows the same pattern but adds a `model_validator` that normalizes and restricts the backend enum to `sqlite|qdrant|none`. `ExportsConfig` is the simplest — just two `Path` fields with `DEFAULT_EXPORTS_CHATGPT`/`DEFAULT_EXPORTS_CLAUDE` fallbacks anchored to `$XDG_DATA_HOME/polylogue/exports/`.

## NixOS module changes

The big structural move is decomposing the inline `envVars` dict in `mkService` into composable pieces:

```nix
baseEnv = cfg.environment // driveEnv // qdrantEnv // {
    XDG_CONFIG_HOME = toString cfg.configHome;
    XDG_DATA_HOME = toString cfg.dataHome;
    XDG_STATE_HOME = toString cfg.stateDir;
    POLYLOGUE_FORCE_PLAIN = "1";
};
```

`driveEnv` and `qdrantEnv` are computed from their respective option groups, and each filters out `null` values so unset options don't produce empty environment variables. The config JSON emits `drive`, `index`, and `exports` sections alongside `defaults`.

New option groups: `qdrant.{enable, url, apiKey, collection, vectorSize}`, `drive.{credentialsPath, tokenPath}`, `exports.{chatgpt, claude}`, and top-level `indexBackend` enum. Four validation warnings catch common misconfigurations (e.g., enabling Qdrant without setting a URL, or enabling drive-sync without a credentials path).

`configHome` default changes from `/etc/polylogue` to `/var/lib/polylogue/config` — the previous default was technically wrong for FHS, since `/etc` should hold static configuration while `configHome` hosts mutable state like OAuth tokens.

## CLI pre-flight checks

`_dispatch_sync()` validates that Drive credentials exist (checking config path, then the default) and that export directories exist for ChatGPT/Claude providers before dispatching to the sync runner. The error messages include the config key to set, which saves operators a trip to the docs.

## Path resolution cleanup

`util.py` replaces two hardcoded module-level `Path(os.environ.get(..., str(DATA_HOME /...)))` expressions with a `_resolve_path(env_var, home_default, xdg_default)` helper that follows a three-tier fallback: environment override → canonical tool home directory (e.g., `~/.codex/sessions`) → XDG data directory. This matters because Codex and Claude Code use their own home directories by default, and the previous code would always fall through to the XDG path even when the tool's native directory existed.

## Cleanup

- `docs/architecture.md` — stale Nov 2024 document that described a structure the codebase has since departed from. The inline docstrings and README are the living documentation.
- `polylogue.py` — dead entry point shim; the installed package uses `polylogue.cli:main` directly.
- `drive._retry()` env-var parsing — retry tuning now flows through `DriveConfig` instead of being parsed from the environment on every retry loop iteration.
- `chatgpt-import` automation target renamed to `chatgpt-sync` (matching the CLI verb); new `claude-sync` target added.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 7 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 4 finding(s) confirmed fixed/stale.

PR #32, opened 2026-04-07
feat: database-first ingestion, raw blob archive, and pure-Python CLI

Sinity:
## Summary

Rewrites the import pipeline around database-first storage, adds a raw blob archive for import resilience, and replaces every external binary dependency with pure-Python equivalents. The CLI surface gains modular command files, a `reprocess` command for failed imports, and several watch/sync hardening flags.

## Motivation

Three problems drove this work:

**Import fragility**: When a provider changes their export format, parsing fails and the user loses the entire import. There's no way to recover without re-exporting from the provider. Raw blob storage fixes this — we archive the original data *before* parsing, so a future parser update can reprocess it.

**External binary coupling**: The interactive UI required gum, sk, bat, glow, and delta — five external binaries that only exist reliably inside `nix develop`. Outside the devshell, every interactive command fails. Replacing these with questionary + pygments + rich eliminates the deployment fragility entirely.

**Configuration brittleness**: Hand-rolled dataclass parsing with manual `isinstance` checks for every field. No env var support, no validation errors that point at the problem. Pydantic-settings provides all of this for free.

## Why database-first?

The previous architecture wrote markdown files as the primary output and treated the database as a secondary index. This created several problems:

- Dirty file detection was fragile (hash comparison of rendered output)
- Re-rendering required re-parsing source files
- Branch tree generation was tightly coupled to import
- Skip/force/allow-dirty semantics were confusing

Now `process_conversation()` writes *only* to SQLite. The `DatabaseRenderer` generates markdown on demand from DB content. This separates concerns cleanly:

```python
# Before: import → parse → render markdown → write files → update DB
# After:  import → parse → write DB (done)
#         render → read DB → write files (separate step)
```

The `render --force` flag triggers regeneration, useful when templates change or collapse thresholds are adjusted.

### Raw import storage

New `raw_storage.py` provides `store_raw_import()` which compresses and writes per-conversation blobs with versioning:

```python
data_hash = store_raw_import(
    data=raw_bytes,
    provider="chatgpt",
    conversation_id=conv_id,
    source_path=export_path,
)
# Later, on parse success/failure:
mark_parse_success(data_hash)
mark_parse_failed(data_hash, traceback.format_exc())
```

Version safety: uses `MAX(version) + 1` so old versions can never overwrite newer ones. Automatic cleanup keeps last 100 versions per conversation.

The `reprocess` command retries failed imports, optionally using the fallback heuristic parser that extracts text from unknown JSON structures.

### Schema v5

Key additions to the messages table:
- `model` — source model name when available
- `content_text` — raw text content for FTS indexing
- `content_json` — structured content (tool calls, function calls)
- `is_leaf` — marks terminal messages in each branch

New tables:
- `assets` — per-message structured attachment metadata
- `raw_imports` — conversation-aware versioned blob storage

FTS is trigger-maintained — inserts/updates/deletes on the messages table automatically propagate to messages_fts.

### Dependency stack

| Before | After | Why |
|--------|-------|-----|
| requests | httpx[http2] | Async support, HTTP/2, connection pooling |
| aiohttp | (removed) | httpx covers both sync and async |
| pyperclip | platform-native | pbcopy/xclip/wl-paste via subprocess |
| gum/sk/bat | questionary/pygments | Pure Python, no binary deps |
| dataclasses config | pydantic-settings | Env vars,.env files, validation |

### CLI modularization

Nine command modules extracted to `polylogue/cli/commands/`:

```
commands/
├── sync.py         # Drive + local provider sync
├── render.py       # JSON → Markdown/HTML
├── config.py       # init/set/show with declarative detection
├── attachments.py  # stats/extract
├── browse.py       # branches/stats/status/runs/inbox
├── import_cmd.py   # chatgpt/claude/claude-code/codex
├── search.py       # FTS5 search
├── maintain.py     # prune/doctor/index/restore
└── status.py       # Drive info + run history
```

Each exports `setup_parser()` and `dispatch()`. The `config init` and `config set` commands check `is_config_declarative()` and refuse to mutate NixOS-managed configs.

### Watch and sync hardening

- `--watch-plan` prints the assembled command without running (useful for cron/systemd)
- `--fail-on-stall` exits with code 2 when no progress for `--stall-seconds`
- `--tail` logs changed paths as they're detected
- `--prune-snapshot` creates a rollback zip before pruning
- Base directory validation gives actionable hints ("pass --base-dir to override")

## Testing

- `test_importers_multi.py` updated for database-first: imports write to DB, then `_render_result()` generates markdown via DatabaseRenderer for assertion
- New `test_db_renderer.py` covers ConversationData/MessageData dataclasses, empty DB queries, provider filtering
- New `test_fallback_parser.py` covers text extraction, timestamp detection, message recovery from both array and mapping formats, degraded markdown generation
- New `test_raw_storage.py` covers hash computation, store/retrieve round-trip, duplicate detection, parse status marking, import statistics
- New `test_schemas.py` covers Pydantic ChatGPT/Claude schema validation, missing field detection, ID normalization, message format variants

## Release Notes

* **New Features**
  * Added modular CLI commands: `search`, `browse`, `config`, `import`, `sync`, `render`, `status`, `maintain`, `attachments` with enhanced options and filtering.
  * Introduced `reprocess` command to retry failed imports with fallback heuristic parsing.
  * Added database-driven markdown rendering and snapshot backup functionality.
  * Improved user interactions with native Python prompts (removed external dependencies).

* **Enhancements**
  * Configuration now detects declarative/immutable settings and prevents unsafe mutations.
  * Sync and watch modes support configurable retries, stall detection, and snapshot backups.
  * Search supports multiple output formats and filters.
  * Raw import tracking for debugging and recovery.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 3 finding(s) confirmed fixed/stale.

PR #34, opened 2026-04-07
feat: add Click CLI surface and restore file-write persistence

Sinity:
## Summary

Restores on-disk file rendering that the DB-first migration dropped, and introduces a Click-based CLI entrypoint that wraps all existing command dispatchers. The database remains the canonical data store, but sync and import workflows need immediate file outputs to function — this commit makes both paths live simultaneously.

## Motivation

The previous commit moved conversation processing to a DB-only write path: `process_conversation()` returned stub `ImportResult` values with projected paths but never actually wrote markdown or HTML. That broke every downstream consumer that expected files on disk — the sync summary reporters, the branch explorer, the `--open` flag on search, and the import clipboard feature all read from the filesystem.

Rather than rewrite all consumers to query the database, the pragmatic fix is dual-write: persist to DB first (source of truth), then call `persist_document()` to materialize files. This preserves the DB-first guarantee while unblocking the existing CLI surface.

The Click migration is motivated by argparse's poor composability for nested subcommand groups. Commands like `maintain doctor`, `browse branches`, and `config set` are natural Click subgroups but awkward argparse sub-sub-parsers. Click also gives us typed path parameters, choice validation, and `--help` for free.

### Conversation dual-write (`conversation.py`)

The core change: after `registrar.record_branch_plan()` and `registrar.record_attachments()`, we build a `MarkdownDocument`, call `persist_document()`, and populate `ImportResult` from the persistence result instead of returning stubs:

```python
persisted = persist_document(
    document=document,
    output_dir=output_dir,
    registrar=registrar,
    provider=provider,
    # ... full metadata
)

return ImportResult(
    markdown_path=persisted.markdown_path,
    html_path=persisted.html_path,
    skipped=persisted.skipped,
    # ...
)
```

Branch transcripts are written under `conversation_dir/branches/{branch_id}/` with overlay stubs. These are wrapped in try/except — branch files are auxiliary and must never block ingestion.

### Click CLI (`click_app.py`)

Every command follows the same translation pattern:

```python
@cli.command()
@click.argument("provider", type=click.Choice([...]))
@click.option("--dry-run", is_flag=True, ...)
@click.pass_obj
def sync(env: CommandEnv, **kwargs) -> None:
    args = Namespace(**kwargs)
    sync_cmd.dispatch(args, env)
```

Click collects typed kwargs, we wrap them in an `argparse.Namespace`, and delegate to the existing dispatch function unchanged. No business logic moves in this commit — the Click layer is purely a CLI-framework swap.

Subcommand groups: `browse` (branches/stats/status/runs/inbox), `maintain` (prune/doctor/index/restore), `config` (show/set/init), `attachments` (stats/extract).

### UI facade rewrite (`ui/facade.py`)

The `questionary` dependency is removed entirely. Interactive mode now shells out to `gum` for confirmation/input, `skim` for fuzzy selection, and expects `bat`/`glow`/`delta` for rendering. These are provided by the Nix devshell.

There's no graceful degradation — `_ensure_interactive_deps()` checks for all five binaries and raises `RuntimeError` with an actionable message if any are missing. This is deliberate: partial interactive support creates confusing UX.

### Registrar branch preservation

`record_document()` was clobbering `current_branch` with `None` on every re-render. Now it reads the existing value from the conversations table before upserting:

```python
row = conn.execute(
    "SELECT current_branch FROM conversations "
    "WHERE provider = ? AND conversation_id = ?",
    (provider, conversation_id),
).fetchone()
if row is not None:
    existing_branch = row["current_branch"]
```

This prevents branch state loss during re-sync operations.

### Drive download compatibility

`download_file()` previously used `with session.get(...) as resp:` which assumes httpx-style async context managers. The new code uses explicit `try/finally` with `resp.close()` and probes for both `iter_bytes` (httpx) and `iter_content` (requests) to support both session types.

### Config directory detection

`is_config_declarative()` was reporting false read-only when the config directory didn't exist yet (common on first run). The fix walks up to the nearest existing ancestor before checking `os.access()`.

### DIFF
--- a/polylogue/cli/__init__.py
+++ b/polylogue/cli/__init__.py
@@ -2,9 +2,11 @@
 
 from ..commands import CommandEnv
 from .context import default_html_mode, resolve_html_enabled, resolve_html_settings
+from .click_app import main, cli  # Click entrypoint
+
+# Legacy exports kept for compatibility with code importing from cli.*
 from .app import (  # pylint: disable=unused-import
     build_parser,
-    main,
     _should_use_plain,
     run_completions_cli,
     run_complete_cli,
@@ -21,29 +23,14 @@ from .app import (  # pylint: disable=unused-import
     _run_sync_drive,
 )
 
-# Legacy compatibility alias; prefer resolve_html_settings.
 _resolve_html_settings = resolve_html_settings
 
 __all__ = [
     "CommandEnv",
-    "build_parser",
+    "cli",
     "default_html_mode",
     "main",
     "resolve_html_enabled",
     "resolve_html_settings",
-    "run_completions_cli",
-    "run_complete_cli",
-    "run_help_cli",
-    "run_import_cli",
-    "run_inspect_branches",
-    "run_inspect_search",
-    "run_prune_cli",
-    "run_stats_cli",
-    "run_attachments_cli",
-    "run_sync_cli",
-    "run_watch_cli",
-    "summarize_import",
-    "_run_sync_drive",
     "_resolve_html_settings",
-    "_should_use_plain",
 ]
--- a/polylogue/cli/app.py
+++ b/polylogue/cli/app.py
@@ -68,6 +68,7 @@ from .context import (
     default_sync_namespace,
     resolve_html_settings,
 )
+from .examples import COMMAND_EXAMPLES
 from .render import run_render_cli
 from ..settings import ensure_settings_defaults, persist_settings, clear_persisted_settings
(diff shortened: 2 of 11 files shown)


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 4 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 6 finding(s) confirmed fixed/stale.

PR #36, opened 2026-04-07
refactor: stamp JSON outputs and harden DB schema resolution

Sinity:
## Summary

Every CLI command that emits JSON wraps its payload through `stamp_payload()`, making all machine-readable output self-describing with schema and application version metadata. Alongside that, the database layer gets three hardening changes that fix real breakage in tests and against older databases.

## Motivation

Two problems drove this work:

1. **JSON outputs were anonymous blobs.** Nothing in the output told a consumer which schema version produced the data, making it impossible to write forward-compatible parsers. Rather than documenting the version contract externally, embedding it in every payload is cheap and self-enforcing.

2. **DB_PATH was a module-level constant.** It captured `XDG_STATE_HOME` at import time, which meant tests that set the environment variable *after* importing `polylogue.db` would silently write to the real user database. This was a recurring source of flaky test failures and a data-safety concern.

A third issue surfaced during the same sprint: databases created before the conversation-aware `raw_imports` schema would crash on ingestion because the expected columns didn't exist. Since Alembic isn't available in the Nix devshell, an inline migration was the pragmatic fix.

### Schema-stamped JSON output

Every CLI module that prints JSON (`imports`, `open_helper`, `render`, `settings_cli`, `sync`) now imports `stamp_payload` from `polylogue.schema` and wraps its payload before serialization:

```python
# Before
print(json.dumps(payload, indent=2))

# After
print(json.dumps(stamp_payload(payload), indent=2))
```

`stamp_payload` adds `schemaVersion` and `polylogueVersion` via `setdefault()` (so callers can override), and dualizes snake_case keys to camelCase for JavaScript consumers.

The `sync/list` path was slightly restructured — the inline `json.dumps({"cmd": "list",...})` is a two-step build-then-print, which reads better and makes the stamping site obvious.

### Dynamic DB path resolution

`default_db_path()` replaces the module-level `DB_PATH` constant. It reads `XDG_STATE_HOME` at call time:

```python
def default_db_path() -> Path:
    raw = os.environ.get("XDG_STATE_HOME")
    state_root = Path(raw).expanduser() if raw else Path.home() / ".local/state"
    return state_root / "polylogue" / "polylogue.db"
```

All consumers (`open_connection`, `doctor.py`, `index_health.py`, `persistence/database.py`) call this function without referencing the constant. The legacy `DB_PATH` is kept for backward compatibility but computed once at import for non-test code paths.

### Legacy raw_imports migration

`_ensure_raw_imports_schema()` runs on every `open_connection()` call. It checks `PRAGMA table_info(raw_imports)` for the presence of `conversation_id` and `version` columns. When they're absent, it:

1. Renames the old table
2. Creates the new schema with conversation_id, version, imported_at_ns, and additional metadata columns
3. Copies existing rows, using `hash` as `conversation_id` and `1` as the default version
4. Drops the old table and creates indexes

This is idempotent — databases that already have the new schema skip the check in two column lookups.

### Messages table backward compatibility

`replace_messages()` now introspects the messages table before inserting. If `model` and `content_text` columns exist, it uses the full 20-column INSERT. Otherwise, it falls back to a 16-column INSERT that matches the legacy schema. This prevents crashes when running new code against an old database.

### Watch: auto-create base directory

In one-shot mode (`--once`) or when a provider sets `create_base_dir = True`, the watch command now creates the base directory instead of erroring. This removes a manual setup step for first-run workflows where the directory doesn't exist yet.

### Smaller fixes

- **config.py**: `persist_config()` accepts `OutputDirs` in the `roots` dict, calling `.render.parent` instead of blindly casting to `Path`
- **render.py**: Chunks receive `_polylogue_index` before any filtering, so `<a id="msg-N">` anchors stay stable when chunks are removed during preparation
- **document_store.py**: `_compute_content_hash()` strips trailing whitespace before hashing, preventing hash churn from YAML frontmatter round-trips
- **fallback_parser.py**: Timestamp extraction now recurses all dict values not only keys matching `time/timestamp/created/updated/date` — the range filter already gates valid candidates
- **local_sync.py**: Uses nanosecond-precision `os.utime(ns=...)` and creates an explicit registrar database inside the output directory's `.polylogue-state/` subdirectory

## Testing

The `default_db_path()` change was specifically motivated by test isolation failures — tests that previously needed to mock `polylogue.db.DB_PATH` can simply set `XDG_STATE_HOME` before calling the function. The migration and column-introspection changes were verified against both fresh databases and databases created with the pre-conversation schema.

### DIFF
--- a/polylogue/cli/imports.py
+++ b/polylogue/cli/imports.py
@@ -19,6 +19,7 @@ from ..importers.claude_code import DEFAULT_PROJECT_ROOT, list_claude_code_sessi
 from ..importers.base import ImportResult
 from ..pipeline_runner import Pipeline, PipelineContext
 from ..util import CODEX_SESSIONS_ROOT, format_run_brief, latest_run, path_order_key
+from ..schema import stamp_payload
 from .context import (
     DEFAULT_CLAUDE_CODE_SYNC_OUT,
     DEFAULT_CHATGPT_OUT,
@@ -109,7 +110,7 @@ def _emit_import_json(results: List[ImportResult]) -> None:
             for res in results
         ],
     }
-    print(json.dumps(payload, indent=2))
+    print(json.dumps(stamp_payload(payload), indent=2))
 
 
 def _emit_print_paths(results: List[ImportResult], ui) -> None:
--- a/polylogue/cli/open_helper.py
+++ b/polylogue/cli/open_helper.py
@@ -6,6 +6,7 @@ from pathlib import Path
 
 from ..commands import CommandEnv
 from ..util import latest_run
+from ..schema import stamp_payload
 from .editor import open_in_editor, get_editor
 
 
@@ -18,7 +19,7 @@ def run_open_cli(args: argparse.Namespace, env: CommandEnv) -> None:
     if not entry and fallback:
         path = Path(fallback)
         if getattr(args, "json", False):
-            print(json.dumps({"path": str(path), "source": "fallback"}))
+            print(json.dumps(stamp_payload({"path": str(path), "source": "fallback"})))
             return
         ui.console.print(str(path))
         return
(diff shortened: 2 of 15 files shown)


Sinity:
@coderabbitai review

Sinity:
@coderabbitai full review

PR #38, opened 2026-04-07
refactor: remove argparse wrappers and make Click primary

Sinity:
## Summary

I removed the last argparse compatibility layer and made the Click tree the only dispatch path, then used that cleanup to land two option-plumbing changes that had been awkward under the dual stack: redacted output writes and raw-hash-based local sync skips. The important architectural change is not just that `polylogue/cli/app.py` disappears; it is that command callbacks in `polylogue/cli/click_app.py` now talk directly to the real `run_*_cli` helpers without converting Click arguments back into `argparse.Namespace` objects and bouncing through `dispatch()` shims. That trims an entire stratum of parser builders, registry code, and helper functions that had to be kept in lockstep with the Click decorators. Once that indirection was gone, it became much simpler to thread `--sanitize-html` through sync/import/render and to put hash-based fast paths in the local sync pipeline without wiring every flag twice. The result is a smaller CLI core, a clearer option flow, and materially less unnecessary work during repeated local syncs.

## Motivation

The branch started from a codebase that was effectively running two CLIs at the same time. `polylogue/cli/click_app.py` owned the public interface, but it still wrapped the older argparse world by constructing `Namespace` objects and forwarding them into `polylogue/cli/app.py`, `polylogue/cli/registry.py`, and the `setup_parser()` / `dispatch()` pairs under `polylogue/cli/commands/`. That duplication had three direct costs.

First, every new flag or command tweak had to be declared twice. A flag like `--sanitize-html` is a good example: if I had left the argparse layer in place, I would have needed matching Click decorators, argparse option registration, wrapper namespaces, and dispatch handling just to carry one boolean to the persistence layer.

Second, the parser split made the code harder to test. The test suite had accumulated assertions against both parser objects and command behavior, but the thing users actually exercised was the Click group. That meant a lot of churn on parser wiring that added little confidence in end-to-end behavior.

Third, local sync was still paying for coarse freshness checks. The raw import state already had hashes available, but session and export-bundle sync still leaned heavily on mtimes, so harmless touches could trigger full parse/render/persist cycles. That was especially painful for Codex and export-bundle workflows where the inputs can be large but stable.

I grouped the redaction and freshness work into the same change because they both needed the new direct option path. They are not random extras bolted onto the CLI cleanup; they are the first concrete features that benefit from removing the compatibility layer.

### 1. Remove the argparse compatibility layer

The largest simplification is the deletion set:

- `polylogue/cli/app.py`
- `polylogue/cli/registry.py`
- `polylogue/cli/arg_helpers.py`
- the argparse `setup_parser()` / `dispatch()` wrappers in `polylogue/cli/commands/*`

In their place, `polylogue/cli/click_app.py` now dispatches directly into the actual command helpers:

```python
@click.pass_obj
def sync(env: CommandEnv, **kwargs) -> None:
    from .sync import run_sync_cli
    run_sync_cli(SimpleNamespace(**kwargs), env)
```

That is a much more honest boundary than:

```python
args = Namespace(**kwargs)
sync_cmd.dispatch(args, env)
```

because there is no hidden second parser anymore. Moved help/completion-facing introspection into Click-native helpers like `click_introspect.py` and `completions_cli.py`, so completions walk the actual command tree without reconstructing parser state from argparse subparsers.

A second-order benefit of this split is that the old monolith gets decomposed into modules that have a single reason to change: `branches_cli.py`, `search_cli.py`, `compare_cli.py`, `config_editor.py`, `failure_logging.py`, `prune_cli.py`, and `path_policy.py`. The CLI is still one product surface, but it no longer lives in one 1700+ line file.

### 2. Thread redaction through the real write path

The new `polylogue/redaction.py` adds `sanitize_text()` and centralizes the masking rules instead of sprinkling ad hoc substitutions across CLI commands. The actual behavior matters more than the regex list:

- redaction happens before the final content hash is computed
- the persisted metadata records that the document was redacted
- run payloads surface the same state for automation

That makes `--sanitize-html` deterministic. A redacted archive is not just a raw archive with a UI flag attached; it is a distinct persisted representation with its own content hash. That choice keeps cache behavior, drift detection, and downstream comparisons coherent.

The flag is carried all the way through the real pipeline:

`click_app.py` → sync/import/render helpers → `process_conversation()` → `persist_document()`

Intentionally kept the redaction at the persistence boundary. Doing it earlier would have made importer behavior provider-specific; doing it later would have let metadata and hashes drift away from the actual emitted document body.

### 3. Skip unchanged local inputs by hash, not just mtime

The local sync changes are small in code volume but meaningful in effect. `_sync_sessions()` now compares the current raw import hash against the stored raw hash before reparsing the session, and `_sync_export_bundles()` stores a per-bundle hash so unchanged ZIPs can be skipped before the import call even runs. I kept the mtime check as a fallback rather than deleting it outright, because hashes are not always available for every legacy path.

This gives the fast path I wanted:

- unchanged session file → no parse/render/persist cycle
- unchanged export bundle → no import pass for the bundle
- changed content → existing pipeline still runs as before

The key point is that freshness is tied to actual source bytes, not just filesystem metadata.

### 4. Preserve the operator-facing parts of the CLI

A pure deletion pass would have regressed usability, so I pulled out the useful CLI-facing bits that had been trapped in `app.py`. `config_editor.py` adds an interactive `config edit` path, `summaries.py` gets a richer render summary table, and `context.py` accepts plain objects instead of requiring argparse namespace semantics. These are the parts that let the Click-only surface feel complete rather than merely “less legacy”.

## Testing

I shifted the CLI tests toward the public surface by using `CliRunner.invoke()` instead of parser-construction assertions. The most important new coverage is:

- `tests/test_redaction.py` for email/token masking and the `polylogue.redacted` metadata stamp
- `tests/test_local_sync.py` for unchanged-session and unchanged-export-bundle hash skips
- the migrated CLI invocation tests, which now exercise the Click group directly rather than verifying argparse wrapper behavior

That testing posture matches the architectural direction of the patch: validate the surviving path, not the code that used to bridge to the removed one.

## Notes

There are still internal call sites that use namespace-like option objects, but they are plain Python objects instead of argparse-specific types. I left that shape in place deliberately because it keeps command helpers lightweight and avoids introducing a premature dataclass layer for every command. The redaction support also only targets persisted Markdown/HTML bodies in this pass; it does not attempt to mutate raw imports or database source payloads. That separation keeps the archival model intact while still allowing sanitized exports when operators need them.

## Release Notes

* **New Features**
  * Added `--sanitize-html` flag to render, sync, and import commands for redacting sensitive content.
  * Added interactive `config edit` command for managing settings.
  * Added branches command for exploring and comparing conversation branches.

* **Improvements**
  * Enhanced configuration display with additional system details.
  * Improved search and statistics command functionality.


Sinity:
@coderabbitai full review

PR #40, opened 2026-04-07
feat: verify archives, export metrics, and resume failed syncs

Sinity:
## Summary

I added three missing operational loops around the archive: a verifier that can prove rendered output still matches the database, a metrics surface that makes run history and provider counts observable without ad hoc SQL, and a resume path that can turn the metadata from a failed run into the input set for the next one. While wiring those pieces, I also added `--jobs` parallelism for local Codex / Claude Code sync, broadened attachment and inbox inspection, and made JSON / snapshot output much more deterministic. The common thread is run metadata: once runs record stable IDs, failure manifests, retry counters, and custom metadata, they become useful for verification, observability, and recovery instead of just being historical log rows. I intentionally let these features land together because each one depended on the same deterministic output work and the same richer run records. The branch also uses that determinism work to improve branch explorers, transcript rendering, and golden tests, so the new commands are backed by outputs that do not drift for incidental ordering reasons.

## Motivation

There were four practical gaps in the archive workflow.

The first was **drift detection**. A rendered `conversation.md` file could diverge from the database, front matter, or attachment index and there was no first-class way to notice short of opening the file and comparing it by hand. That becomes a real problem once the archive is used for search, export, or CI snapshots.

The second was **failed-run recovery**. When a Drive or local sync partially failed, the information about what went wrong existed only indirectly in console output or in inconsistent metadata blobs. Operators had to re-run broad syncs or manually reconstruct the failed subset.

The third was **observability**. We had rich state in SQLite and in the runs table, but nothing that could be scraped, dumped, or graphed without custom SQL. That made it hard to answer basic questions like “how many conversations are indexed per provider?” or “when did the last successful sync happen?”

The fourth was **throughput without determinism loss**. Local sync had obvious concurrency opportunities, but naively parallelizing it would have destabilized slug ordering and snapshot output, which would then undermine the verifier and golden tests.

The solution that kept recurring in the branch was: make run and render output deterministic first, then build verify/metrics/resume on top of that stable substrate.

### 1. Add a verifier that reasons about the whole archive surface

`polylogue/cli/verify.py` is the new user-facing entrypoint, but the real foundation is `polylogue/frontmatter_canonical.py`. I wanted `verify --fix` to do more than just “rewrite some YAML”; it needed to normalize key ordering and optionally scrub absolute paths so the same canonicalizer could also support portable snapshot tests.

The verifier checks several different drift vectors:

- front matter keys vs the conversation row
- content-hash agreement across rendered file, DB state, and recomputed body
- branch document existence
- attachment path existence
- unknown `polylogue.*` keys, with policy control

That last point matters because schema drift is often subtle. If someone adds or renames metadata keys, I want the archive to be able to treat that as ignored noise, a warning, or a hard error depending on the context. `--strict` and `--unknown` make the same command usable both for operator spot checks and for CI gating.

### 2. Turn run history into a real recovery mechanism

The key enabling change is richer run metadata. `polylogue/util.py` gains `get_run_by_id()`, but more importantly sync/render/import runs now record failure manifests and custom metadata in a consistent shape. Once the run row contains `failedChats`, `failedPaths`, failure counters, and retry settings, `--resume-from <id>` can build the next invocation mechanically instead of heuristically.

The recovery rules are conservative on purpose:

- limit the resumed run to the previously failed subset
- force `--all` so there is no picker ambiguity
- disable `--prune` so a recovery run cannot delete unrelated good output

That gives a focused rerun without making the operator remember the exact failed IDs or file paths.

Threaded `--meta key=value` through sync/import/render because the same mechanism is useful both for dashboards and for recovery bookkeeping. It lets external orchestration attach deployment IDs, batch IDs, or environment tags without inventing a separate side channel.

### 3. Add metrics without inventing a separate service

`polylogue/cli/metrics.py` exports a Prometheus-style text format and a structured JSON form from the existing DB plus run history. I kept it in-process and optionally served via a tiny HTTP server because the underlying data is already local and cheap to aggregate.

The metrics break down counts by provider and command, and they surface:

- items processed
- attachments / bytes
- duration
- retries / failures
- last-run timestamps

This is intentionally close to the run payload model. I do not want two parallel telemetry schemas. The same run records that power `--resume-from` should also power `/metrics`.

### 4. Parallelize local sync without destabilizing output

The `--jobs` work in `polylogue/local_sync.py` uses `ThreadPoolExecutor`, but the detail that matters is result collection order. Workers can finish whenever they want; output ordering still follows the original input index. That preserves deterministic slug ordering and keeps branch/render snapshots stable even when the work is concurrent.

I paired that with plain-mode ETA/progress reporting so the same code path remains usable in environments without Rich. Parallelism is only worth it if operators can still understand what the command is doing.

### 5. Pay down the incidental nondeterminism

A lot of smaller changes exist solely because the new verify / metrics / golden tooling would have been noisy without them:

- sorted JSON output via `sort_keys=True`
- sorted rendered attachments
- consistent Drive chat ordering by `modifiedTime`
- run IDs in the runs table output
- inline branch diffs and branch map snippets in rendered output
- attachment routing counts exposed in metadata
- inbox reason codes for malformed exports and ignored items

These are individually modest, but together they make the archive much easier to reason about mechanically.

## Testing

I added fixed-time Markdown and HTML golden snapshots for all four providers so ordering and metadata regressions show up immediately. The command-specific coverage includes:

- `tests/test_metrics_cli.py`
- verify CLI coverage for canonical rewrites and unknown-key handling
- resume-from tests for Drive and local provider paths
- attachment stats tests for provider filters, time windows, and orphan cleanup
- local sync ordering tests for `--jobs`
- inbox tests for malformed export detection and `.polylogueignore` handling

Those tests matter because the branch intentionally touches rendering, CLI metadata, retry logic, and concurrency at the same time. The snapshots and targeted unit tests are what keep that scope safe.

## Notes

I did not try to make `verify` a full schema-migration tool. It canonicalizes and flags drift, but it is still operating on top of the existing archive model. Likewise, the metrics server is intentionally minimal: it is a thin export surface over current state, not a long-running daemon with its own persistence story. The branch’s main value is that verification, observability, and recovery now all speak the same language: stable rendered output plus richer run metadata.

## Release Notes

* **New Features**
  * Added `verify` command to validate conversation integrity and frontmatter consistency.
  * Added `metrics` command to monitor system health and view run statistics.
  * Added resume-from functionality to retry failed syncs from the last failure point.
  * Added attachment filtering (by provider, date range) and orphan cleanup capabilities.
  * Added metadata tagging support for imports and renders.

* **Improvements**
  * Enhanced inbox detection with `.polylogueignore` support.
  * Added branch map visualization in rendered conversations.
  * Improved diff display for non-canonical branches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ Thumbs.db
 # Local dev environment
 .direnv/
 
+# Local scratch
+.tmp/
+
 # Python
 __pycache__
 .pytest_cache/
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -2,11 +2,10 @@
 
 Polylogue now follows a modular structure that separates orchestration concerns from provider-specific logic. The key building blocks are:
 
-## Command Registry
+## CLI (Click)
 
-- `polylogue.cli.registry.CommandRegistry` maps CLI verbs to handlers.
-- `polylogue.cli.app` registers commands and uses the registry for dispatch.
-- Makes it easy to add new commands or reuse handlers across CLI/parsers/UI.
+- `polylogue.cli.click_app` defines the Click command tree and dispatches into `polylogue.cli.*` helpers.
+- This keeps parsing/help/examples in one place while leaving command logic in dedicated modules.
 
 ## Console Facade & UI
 
@@ -59,7 +58,7 @@ Polylogue now follows a modular structure that separates orchestration concerns 
 - Build end-to-end snapshots for `sync`/`render` flows that assert both Markdown output and metadata side effects (state + SQLite) to guard against regressions.
 - Explore richer CLI ergonomics (completions, faceted status dashboards) on top of the existing registry/facade stack. Planned completion work includes:
   - bringing the dynamic engine to bash/fish so every shell benefits from live suggestions;
-  - expanding suggestions beyond slugs/providers to include filtered Drive chats, recent session files, and constrained flag values discovered from argparse metadata;
+  - expanding suggestions beyond slugs/providers to include filtered Drive chats, recent session files, and constrained flag values discovered from Click metadata;
   - caching completions to stay snappy even on large archives;
   - rethinking annotation formats (e.g., structured JSON) so shells can render descriptions/tooltips cleanly.
 
--- a/polylogue/branch_explorer.py
+++ b/polylogue/branch_explorer.py
@@ -4,6 +4,7 @@ import html
 import json
 import textwrap
 from dataclasses import dataclass, field
+from itertools import islice
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 from urllib.parse import quote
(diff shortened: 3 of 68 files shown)


Sinity:
@coderabbitai full review

PR #44, opened 2026-04-07
refactor: regroup CLI commands and switch prompts to questionary

Sinity:
## Summary

I reorganized the CLI around task-oriented groups, moved the core prompt/render facade onto questionary + Rich, and used that transition to clean up several pieces of command plumbing that had been getting harder to explain: completion generation, declarative config handling, OCR defaults, and Claude Code multi-file session parsing. The user-visible headline is the new command tree (`doctor`, grouped `import`, grouped `verify`, richer `browse`, `config prefs`), but the deeper change is that the console layer is testable without having to fake a full interactive shell session for every prompt. That in turn let me replace a lot of PTY-heavy prompt automation with deterministic prompt-file stubs while still keeping targeted end-to-end coverage where it matters. The branch also improves packaging and Nix integration by provisioning OCR data, generating shell completions at build time, and surfacing immutable/declarative config state without pretending every config file is writable. Finally, Claude Code imports get a more realistic session model by merging sidecar agent logs and folding summary-only files into the real session timeline not emitting empty transcripts.

## Motivation

There were two related problems at the start of this branch.

The first was **command sprawl**. The CLI had grown into a wide, shallow top-level tree where primary workflows and secondary utilities all sat side by side. `status`, `env`, `attachments`, `compare`, `prefs`, `open`, `reprocess`, and the maintenance commands were all visible at the same level as `sync`, `search`, and `render`. That made `polylogue --help` noisy, forced the completion engine to special-case one-off command paths, and made it harder to communicate which commands belonged together conceptually.

The second was **interactive/tooling coupling**. The old console path depended on shelling out to helper binaries like gum and delta for formatting and prompts. That made packaging heavier, made tests more fragile, and made it awkward to support the same interactive semantics in plain, CI, and Nix-managed environments.

OCR and declarative config handling were the two nearby pressure points that made this worth fixing now. OCR existed as a flag but not as a coherent defaulted capability that packaging could provision, and config mutation commands had no good way to explain “this file is managed declaratively, you should not edit it here.”

The branch therefore tackles the CLI as both an interaction surface and a packaging surface: group commands better, reduce shell-out behavior in the prompt layer, make completions more self-describing, and make environment/config state visible instead of implicit.

### 1. Rebuild the prompt/render facade around questionary + Rich

The most important code change is in `polylogue/ui/facade.py`. I collapsed the older facade split and moved the core interactions onto questionary for prompt collection and Rich for presentation. That covers:

- choose / confirm / input flows
- markdown rendering
- diff rendering
- banners / panels / status lines
- plain-mode progress text

Added `POLYLOGUE_TEST_PROMPT_FILE`, which lets a test feed a deterministic sequence of prompt responses into the UI without needing a live PTY. That gives us a much cheaper and more direct way to test prompt-driven behavior.

I did **not** try to turn every CLI interaction into a fake in-process abstraction. The branch still keeps targeted PTY coverage for real selection/browse flows, because some things are only trustworthy when the whole CLI entrypoint runs under a pseudo-terminal. But the common case no longer has to go through PTY orchestration just to answer “yes” to a config prompt.

### 2. Group the CLI by concern instead of by implementation history

`polylogue/cli/click_app.py` is much more intentional about command families:

- `maintain` becomes `doctor`
- `import` becomes a group (`run`, `reprocess`)
- `verify` becomes a group (`check`, `compare`)
- `config prefs` absorbs the old top-level `prefs`
- `browse open` absorbs the old top-level `open`
- `browse` gains `metrics`, `timeline`, and `analytics`

That has two benefits. The obvious one is better help output: there is a cleaner distinction between core archive workflows and diagnostic / inspection / utility commands. The less obvious one is that completions can follow the actual Click tree instead of preserving lots of historical command exceptions.

Explicitly let old paths fail as normal Click usage errors rather than silently aliasing everything forever. The regrouping is meant to simplify the public surface, not create a second compatibility maze.

### 3. Make completions data-driven enough to survive the regrouping

The regrouped command tree would have been a regression if completions had stayed naive. The work in `polylogue/cli/completion_engine.py` does three important things:

1. caches state-derived completions in `XDG_CACHE_HOME`
2. introspects Click choice/path types for option-value suggestions
3. handles grouped subcommands as first-class paths rather than concatenated special cases

The state-DB signature plus short TTL for session listings is especially useful because it keeps interactive completion snappy without freezing stale results in place. Build-time completion generation in `flake.nix` then turns that same surface into packaged Bash/Zsh/Fish completions.

### 4. Treat OCR and declarative config as product features, not afterthoughts

OCR has a proper default model:

- `--attachment-ocr/--no-attachment-ocr`
- default-on behavior
- Pillow + pytesseract runtime deps
- TESSDATA wiring in the Nix stack
- language validation in the module layer

That is a better story than “there is a flag somewhere if you happen to know which extra packages to install.”

Declarative config handling gets a similar cleanup. `config show` and `doctor` now surface whether the config is declarative / immutable and why. Mutation commands can then refuse to edit the file with a concrete explanation instead of failing later on write.

### 5. Improve Claude Code ingestion while the CLI plumbing is moving

Claude Code logs turned out to be a good fit for this branch because the interactive/CLI work already touched sync defaults and idempotency tests. `_collect_session_files()` and `_load_session_entries()` merge `agent-*.jsonl` siblings by `sessionId` and sort the merged entries chronologically. Summary-only files no longer create empty transcripts; instead they contribute their summary data to the primary session record or are skipped with an explicit reason.

This keeps Claude Code archives closer to what operators expect: one logical conversation timeline per session, even if the raw logs are split across sidecar files.

## Testing

I split the test strategy by what each surface needs:

- `tests/test_cli_interactive.py` uses prompt-file stubs for deterministic questionary flows
- `tests/test_cli_plain_smoke.py` covers plain-mode subprocess behavior
- retained PTY coverage still checks real interactive selection paths
- idempotency tests cover repeated Drive sync plus interrupted resume flows
- provider-specific sync tests expand for ChatGPT / Claude / Claude Code / Codex
- completion/help tests cover the regrouped command tree and new value suggestion logic

That mix is deliberate. Prompt-file stubs are faster and easier to debug for direct UI flows; PTY coverage remains where full entrypoint behavior is the thing being tested.

## Notes

The branch does not completely eliminate every external helper from every environment. Some preview-oriented tools are still present in packaging/tests where they remain useful, and I did not try to rewrite every historical helper path in one pass. The important shift is that the core prompt/render facade and config flows no longer *require* gum/delta-style shell-outs to function. The command regrouping also intentionally prefers clearer structure over backward-compatible aliases, so follow-up documentation and automation updates are expected wherever the old top-level paths were baked into scripts.

## Summary by CodeRabbit

* **New Features**
  * New doctor command group (status/check/prune/env/attachments) and verify compare subcommand.
  * Built-in interactive prompts (consistent across OS) and shell auto-detection for completions.
  * Browser-first opening for HTML outputs; OCR (Tesseract) support for attachments.

* **Changes**
  * Import command reorganized to use an explicit run subcommand.
  * Default Drive folder renamed to "Google AI Studio".
  * Attachment OCR enabled by default; sync/reporting now surfaces per-reason skip counts.
  * Browse always writes HTML explorers and can auto-open them.

* **Documentation**
  * README and docs updated to reflect new commands, flags, defaults, and UX notes.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 6 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 11 finding(s) confirmed fixed/stale.

PR #46, opened 2026-04-07
feat: resolve Drive settings from config and stream export detection

Sinity:
## Summary

I centralized path and credential resolution for Drive sync, replaced several import-time root globals with config-aware helpers, and rebuilt export detection so ChatGPT / Claude bundles can be found and classified reliably even when they are nested or too large for eager parsing. The practical outcome is that `config.json` can drive more of the real sync behavior without acting like documentation for a partially env-driven runtime. At the same time, the archive utilities that scan outputs (`attachments`, `prune`, `stats`, `status`) respect the active config object not whatever roots happened to be baked into module state at import time. The export-detection work is equally important because repeated sync of local exports only works well if the system can find the right `conversations.json` and classify provider data without pulling a giant archive into memory. Used the same pass to tighten doctor output, Claude Code path discovery, and the Nix module’s Drive-related settings so the effective runtime surface is more coherent end to end.

## Motivation

This branch came out of a cluster of bugs that all pointed to the same architectural weakness: path and config resolution was happening in too many places and at too many times.

Drive credentials are the clearest example. Users could declare credential paths in config or through Nix-managed settings, but parts of the runtime still looked directly at env vars or built-in defaults. That meant `config.json` was not actually the source of truth for the full sync path.

Output roots had a similar issue. `DEFAULT_OUTPUT_ROOTS` was computed at import time from the module-level config, which meant later changes to `env.config` or labeled roots were invisible to commands that should have been scanning the active archive layout. Tests had to patch globals, which is always a warning sign.

Export discovery was the third pain point. The archive scanner worked fine for shallow, conventional exports, but it was much less trustworthy for nested bundles or very large `conversations.json` files. Loading a massive export just to decide whether it looked like ChatGPT or Claude is unnecessary work, and it becomes fragile when the schema is slightly irregular.

The fix was not “patch one bug at a time.” The fix was to make config resolution explicit and pass the active config object all the way through the places that need it.

### 1. Make Drive client construction deterministic

The most important product change is the introduction of a single construction path for Drive access. `build_drive_client()` in `polylogue/commands.py` resolves:

- credentials path
- token path
- retry count
- retry base delay

using the same precedence chain everywhere:

1. explicit env overrides
2. config values
3. built-in defaults

Updated `DriveClient.__init__()` so callers can pass explicit credential/token paths instead of hoping the right environment variables are present. That makes the client much easier to reason about in CLI code, tests, and the Nix service layer.

The config side mirrors that change. `DriveConfig` is a first-class part of config parsing/persistence, and it accepts both snake_case and camelCase keys so the runtime is tolerant of how the value got there while still writing back a stable config shape.

### 2. Stop baking output roots at import time

`resolve_output_roots(config)` is the small helper with outsized impact in this branch. Instead of treating output roots as a static module-level constant, commands now ask for the roots from the active config object when they actually need them.

That affects:

- attachments stats / extraction
- prune
- archive stats
- status-like scans

The benefit is simple: these utilities now follow `env.config` and labeled roots correctly. It also makes tests much cleaner because they can modify the config object under test instead of monkeypatching imported globals.

I paired that with a prefs cleanup (`load_prefs()` as a public helper and `CommandEnv.prefs` populated at startup) so both config and per-command preference state now enter the CLI through explicit initialization rather than accidental module state.

### 3. Rebuild export detection around layered strategies

The detection pipeline now tries the cheapest dependable thing first and only escalates when necessary. The important pieces are:

- `find_conversations_json()` for recursive discovery
- streaming classification for large files
- full parse when the file is small enough or the stream path is insufficient
- snippet heuristics as a fallback for oversized or awkward cases
- provider filtering at discovery time so the sync path does less speculative work

That layered approach is what makes the change valuable. I did not want to replace one fragile classifier with another fragile classifier. The scanner has a progressive strategy that can handle both conventional exports and nested/unusual ones without assuming everything should be slurped into memory.

### 4. Tighten sync and doctor behavior around the new config model

Once paths were explicit, a few sync/doctor mismatches became obvious.

In local sync:

- ChatGPT / Claude export base dirs now come from config unless overridden
- output directories are checked for writability up front
- mtime checks run before content hashing for the fast path
- hash-based skips require the markdown output to exist, which avoids over-trusting partially written state

In doctor:

- expensive checks can be skipped explicitly
- progress callbacks surface long-running phases
- output directories are checked for existence/writability
- credential resolution uses the same config-aware path as sync

Claude Code path discovery also gets a needed update by checking both `~/.claude/projects/` and `~/.config/claude/projects/`, with `POLYLOGUE_CLAUDE_CODE_PROJECTS` still available as an escape hatch.

### 5. Carry the same rules into Nix

The Nix module changes are not incidental. Once Drive settings became explicit config, it made sense to expose:

- `credentialsPath`
- `tokenPath`
- `retries`
- `retryBase`

at the module level and to wire the right environment into the service units. That gives declarative installs a credible story for the same behavior the CLI uses locally.

## Testing

The branch broadens tests around the exact resolution rules it changes:

- Drive client tests for credential/token precedence and auth prompting
- config-root and stats tests for active-root resolution
- local sync tests for export detection and skip behavior
- doctor tests for output-path checks and progress/reporting
- interactive/plain config coverage for the updated initialization path

This is important because precedence bugs are usually subtle. The tests are there to pin down which layer wins rather than leaving that behavior implicit.

## Notes

Kept the env → config → default ordering instead of trying to outlaw env overrides. That is still the most practical model for CLI tools, tests, and Nix-managed services. The goal of the branch is not to eliminate env overrides; it is to make them deliberate and consistent. Intentionally left export classification layered rather than “one true parser,” because the source data is diverse enough that a progressive strategy is easier to keep reliable.

### DIFF
--- a/docs/providers/claude-code.md
+++ b/docs/providers/claude-code.md
@@ -1,6 +1,6 @@
 # Claude Code Session Imports
 
-Claude Code stores IDE transcripts beneath `~/.claude/projects/`. Polylogue mirrors those JSONL streams into Markdown while preserving workspace context, shell snapshots, and tool outputs.
+Claude Code stores IDE transcripts beneath `~/.claude/projects/` (or `~/.config/claude/projects/` on some systems). Set `POLYLOGUE_CLAUDE_CODE_PROJECTS` if you need to override the default. Polylogue mirrors those JSONL streams into Markdown while preserving workspace context, shell snapshots, and tool outputs.
 
 ## Project Layout
 
--- a/nix/modules/polylogue.nix
+++ b/nix/modules/polylogue.nix
@@ -13,6 +13,13 @@ let
   configPath = cfg.configHome + "/config.json";
 
   tessdataDir = "${cfg.ocr.tessdataPackage}/share/tessdata";
+  driveCredentialsPath = if cfg.drive.credentialsPath != null
+    then cfg.drive.credentialsPath
+    else cfg.configHome + "/credentials.json";
+  driveTokenPath = if cfg.drive.tokenPath != null
+    then cfg.drive.tokenPath
+    else cfg.configHome + "/token.json";
+
   envVars = cfg.environment // {
     XDG_CONFIG_HOME = toString cfg.configHome;
     XDG_DATA_HOME = toString cfg.dataHome;
@@ -20,7 +27,9 @@ let
     POLYLOGUE_FORCE_PLAIN = "1";
     POLYLOGUE_DECLARATIVE = "1";
     TESSDATA_PREFIX = tessdataDir;
-  };
+  }
+  // optionalAttrs (cfg.drive.credentialsPath != null) { POLYLOGUE_CREDENTIAL_PATH = toString cfg.drive.credentialsPath; }
+  // optionalAttrs (cfg.drive.tokenPath != null) { POLYLOGUE_TOKEN_PATH = toString cfg.drive.tokenPath; };
 
   missingOcrLangs = lib.filter (lang: !(builtins.pathExists "${tessdataDir}/${lang}.traineddata")) cfg.ocr.languages;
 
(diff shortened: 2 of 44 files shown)


Sinity:
@coderabbitai full review

PR #48, opened 2026-04-07
feat: replace provider workflows with a unified run pipeline

Sinity:
## Summary

I replaced the old “one workflow per provider or transport” model with a single source-driven pipeline that ingests, renders, and indexes conversations through one command surface. without maintaining separate sync/import/render stacks for Drive, local session folders, and provider-specific export bundles, the archive now reasons about a list of configured sources and runs them through one shared pipeline. The core of that shift is `plan_sources()` / `run_sources()`, a new config schema built around `Source` objects, and a canonical storage model that all ingestion paths target. The CLI shrinks accordingly: `run`, `index`, `search`, `health`, `export`, `config`, and `state` become the main surface, and the older provider-specific command families disappear with the code they depended on. This branch was developed in an isolated `v666` track before being promoted, which let me build the new stack in parallel, migrate docs/tests, and then delete the legacy modules once the unified path was viable.

## Motivation

The old architecture had become expensive in exactly the wrong way. The codebase supported several conceptually similar workflows:

- sync from Drive
- sync from local session directories
- import provider export bundles
- render stored data
- update indexes

but those workflows were implemented through different module families, different command surfaces, and different provider-specific helpers. That made several things harder than they needed to be:

1. adding a new source type or provider behavior meant teaching multiple pipelines about it
2. config was organized around provider sections rather than around the actual thing the system consumes: a source
3. search/index/render had to bridge across transport-specific storage and provider-specific archive structure
4. the Nix / automation story was still provider-watch-service-oriented instead of pipeline-oriented

The rewrite goal was not “make the code shorter.” The goal was to align the architecture with the actual product model: Polylogue ingests conversations from sources, persists them canonically, renders them, and indexes them. Providers and transports are attributes of the source, not reasons to duplicate the whole workflow surface.

## Approach: build the new stack beside the old one, then promote it

The branch history matters here. I did not mutate the legacy stack in place from the first commit. Instead, I incubated the new path under `v666`-named files and tests, which gave me room to do the following safely:

- shape the new config model
- build the canonical DB schema and storage API
- add a new CLI surface
- port render/search/health behavior
- migrate tests and docs
- prove the new path against Drive and local sources

Only once that surface was coherent did I remove the suffixes, delete the old modules, and make the new pipeline the default. That progression is why the branch can remove so much code at the end without the final commit feeling like a blind leap.

### 1. Redefine config around sources, not providers

`polylogue/config.py` now models a v2 config with a source list. Each source is either a local path or a Drive folder, and each one can be planned and run by the same pipeline. This is a better fit than provider-specific config sections because it matches the operator’s actual question: “what should Polylogue ingest?” not “which provider-specific command should I use?”

The config layer also handles forward migration from the older shape so existing users do not have to rewrite everything by hand before the new stack can start.

### 2. Make planning and execution pipeline-native

`plan_sources()` and `run_sources()` in `polylogue/run.py` become the center of gravity. They own:

- source resolution and filtering
- ingest planning
- render/index gating
- per-run state and output

That lets the CLI ask one coherent question (“run these sources through these stages”) without forcing the operator to choose among multiple transport/provider verbs first.

The CLI simplification is a direct consequence of that model. I did not remove commands just to make `--help` shorter; I removed them because the new pipeline gives us a cleaner abstraction that makes those commands redundant.

### 3. Canonicalize storage and ingestion

The new ingestion path standardizes on shared parsed record shapes and a canonical SQLite schema. Whether the source was a Drive folder or a local export bundle, it now yields the same `ParsedConversation` / `ParsedMessage` / `ParsedAttachment` model before persistence.

That standardization is what makes several downstream simplifications possible:

- `store_records()` can be idempotent without caring where the data came from
- `render.py` can read from the DB instead of re-deriving provider-specific archive layout
- `search.py` can query one canonical index not translating provider-specific storage assumptions
- assets can live under content-addressed paths rather than scattered per-provider attachment layouts

Intentionally narrowed indexing to SQLite FTS5 in this branch. Keeping one canonical search/index story was more important here than preserving multiple backends while the surrounding architecture was changing.

### 4. Replace provider-specific workflow code with shared ingestion

`source_ingest.py` and `drive_ingest.py` are the key enabling modules on the ingestion side. They let the pipeline reason about sources generically while still performing provider detection from the source data itself. That moves provider handling down to parsing, where it belongs, instead of letting provider identity dictate the top-level command structure.

Drive gets its own major simplification too. The branch replaces the older REST/httpx approach with `google-api-python-client` and `google-auth-httplib2`, which gives a cleaner way to resolve folders, iterate JSON-like files, and download bytes or paths.

### 5. Pull Nix and operational tooling onto the same model

The Nix module emits source config for a single run-oriented service instead of proliferating provider-watch services. That is the right reflection of the new product shape. The same unification shows up in health/export/state support and later additions like latest-render output, state reset, and completions.

## Breaking changes

This branch is intentionally disruptive. The main user-visible ones are:

- config moves to a source-list model
- provider-specific sync/import/render command families are replaced by the unified `run`-centric surface
- older internal modules (`local_sync`, provider-specific CLI submodules, older validation/settings/redaction paths, legacy branching stack) are deleted
- search/index standardize on the new canonical store and FTS path

The app migrates config forward where it can, but scripts and docs that targeted the old command taxonomy will need to move.

## Testing

The branch replaces the old suite with a new set of core tests around the unified stack while it is still incubating, then promotes those tests along with the code. The coverage includes:

- CLI and run-path tests for the new surface
- config tests for the new schema
- ingest/render tests over the canonical storage model
- Drive ingest tests for the new Drive API path
- search/health tests against the new DB-backed model

During the `v666` phase there are parallel test files that validate the new implementation before it becomes the default. That is exactly why I chose the side-by-side incubation approach.

## Notes

I deliberately kept the scope focused on architectural unification, not on preserving every old feature verbatim. Some things are simplified aggressively in service of the new model—for example, search/index standardization and command-surface collapse. The branch also includes later polish around latest-render behavior, state reset, completions, incremental indexing, and rerender-on-metadata-changes because those are part of making the new default surface usable once the old one is gone.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 3 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 6 finding(s) confirmed fixed/stale.

PR #50, opened 2026-04-07
refactor: split CLI, pipeline, and importers into packages

Sinity:
## Summary

I split the unified-pipeline implementation into smaller packages with clearer responsibilities, and I used that decomposition to add a few capabilities that fit naturally once the boundaries existed: a FastAPI server, vector search via Qdrant, configurable render roots/templates, and a richer library/repository layer for programmatic consumers. The three main monoliths from the previous step—`click_app.py`, `run.py`, and `source_ingest.py`—were each trying to do too much at once. The CLI mixed parsing, command logic, and formatting; the pipeline mixed planning, IDs, execution, and result models; and the ingest module mixed provider detection with every parser implementation. Pulling those apart made the architecture easier to extend and made new surfaces like `serve`, templated rendering, and library models feel like additions to existing packages instead of new branches on a giant central file. The branch also tightens storage performance by moving to native UPSERTs and better connection reuse, and it upgrades packaging/test infrastructure so the new module boundaries are actually exercised in isolation.

## Motivation

The previous branch deliberately unified the product around a single pipeline, but it also concentrated a lot of logic into a few files. That is a reasonable tradeoff during a rewrite, but it becomes a maintenance problem if left in place.

Three files were especially overloaded:

- `polylogue/cli/click_app.py`
- `polylogue/run.py`
- `polylogue/source_ingest.py`

Each one contained multiple layers of concern that wanted to evolve independently. For example, server-related functionality and vector search are natural product additions once there is a canonical storage model, but they do not belong in the same module that parses CLI flags. Likewise, large-export streaming improvements belong in importer modules, not in a file that also owns provider detection and orchestration concerns.

This branch therefore has two goals:

1. decompose the unified stack into packages with clearer ownership
2. take advantage of those new boundaries to add features that were awkward before, especially server/library/vector-search capabilities

There is one subtle motivation worth calling out too: performance work is easier to do correctly once the boundaries are cleaner. Native UPSERTs, reusable connections, bounded ingestion parallelism, and streaming parse logic are all easier to verify when the call graph is not routed through a small number of oversized files.

### 1. Split the CLI into command modules and support layers

The CLI decomposition is more than mechanical file movement. `polylogue/cli/click_app.py` now acts like an entrypoint should: it wires the group and delegates the real work to command modules plus a few support modules (`formatting.py`, `helpers.py`, `types.py`).

The new command package makes several things easier:

- command-specific tests no longer need to import a giant app file
- formatting and environment helpers are reusable without circular imports
- new commands like `serve` and `sources` can land without inflating the entrypoint again

Search also gets a small but useful capability boost here: `--source` and `--since` filters make it easier to target the new source-centric archive model.

### 2. Pull the pipeline into an actual package

`polylogue/pipeline/` is the architectural heart of the branch. I pulled out:

- `ids.py` for hashing / ID generation
- `ingest.py` for preparation logic
- `models.py` for `PlanResult` / `RunResult`
- `runner.py` for execution

`run.py` remains as a compatibility facade without disappearing outright, which softens the move for code that already imports it. That is an important distinction: the branch is trying to improve internal cohesion without needlessly breaking every existing import site.

Added bounded concurrent ingestion in the runner. The “bounded submission” part matters just as much as the concurrency itself; it keeps the ingestion queue from growing without bound when the source set is large.

### 3. Turn source ingestion into a provider package instead of a mega-parser

The importer split is what made the streaming and parsing cleanup tractable. Once each provider has its own module under `polylogue/importers/`, shared pieces can live in `base.py` and the provider-specific logic can be explicit.

The big behavior change here is the layered `ijson` parsing strategy. Instead of defaulting to `json.loads()` for everything, the importer path can try:

- root-list streaming
- conversations-list streaming
- full-load fallback

That makes the unified archive model much friendlier to large exports while preserving the ability to recover in odd cases.

### 4. Add a core utility layer for things that kept reappearing

Hashing, timestamp parsing, JSON handling, and logging had started to duplicate themselves across pipeline and ingest modules. Extracting `polylogue/core/` keeps those behaviors consistent and removes a lot of small copy/paste drift.

This also helps the new server and library layers, which have a natural place to import those low-level concerns from.

### 5. Extend the platform: render roots, templates, server, and vector search

Once render output and archive storage were separated conceptually, it made sense to expose that in config. `render_root` becomes a first-class setting, and `template_path` allows custom Jinja2 templates for HTML rendering.

That same separation made the FastAPI server and the repository/lib layer feel natural. The server is not reaching into raw parser state; it is serving the same canonical conversation model the CLI uses. The repository improvements and rich `Conversation` / `Message` / `Attachment` models then become useful both to the server and to programmatic consumers.

Qdrant fits this pattern too. `index_qdrant.py` gives the archive a vector-search path without contaminating the core FTS flow with backend-specific details.

### 6. Tighten storage and packaging around the new boundaries

The storage changes are deliberately low-drama but high-impact:

- native UPSERTs instead of select-then-update cycles
- connection-context helpers
- optional connection injection into store/index helpers
- repository query improvements for bulk fetches

Packaging/test work follows the same theme. The branch adds the runtime dependencies required by the new server/logging/vector surfaces, patches the RE2 overlay, restores flake checks, and hardens completion generation so packaging does not fail just because shell completion generation has a transient issue.

## Testing

The decomposition would not mean much without tests that prove the seams hold. The branch adds or strengthens:

- `tests/test_cli.py`
- `tests/test_run.py`
- `tests/test_source_ingest.py`
- `tests/test_retries.py`
- `tests/test_qdrant.py`
- `tests/test_server.py`
- `tests/test_integration_flow.py`
- shared factories / sample fixtures

That set is intentionally broader than a pure refactor suite because the branch adds new externally visible surfaces as well as moving code.

## Notes

One interesting thread in the branch history is the temporary removal of explicit pydantic dependencies before the server work made Pydantic models central again via FastAPI and richer package boundaries. I did not preserve that temporary simplification in the final architecture because the decomposition plus server/library work made typed models more valuable than the shorter dependency list. The final shape favors cohesive packages and extensibility over minimizing imports at all costs.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 2 finding(s) confirmed fixed/stale.

PR #52, opened 2026-04-07
refactor: stage ingestion writes and add typed projections

Sinity:
## Summary

I rebuilt the archive pipeline around explicit validate / parse / persist stages, tightened the storage path so concurrent work cannot corrupt attachment state or duplicate equivalent message content, and added a projection-oriented read layer for analysis and safer search results. This branch is partly architectural and partly defensive: the staged pipeline gives us clearer write boundaries, while the projection layer gives higher-level code something better than raw row-shaped data to reason about. Used the branch to harden several correctness and security edges that become more important once the archive has richer query surfaces: FTS escaping, ZIP bomb rejection, encoding fallbacks, asset extraction races, shell-quoted viewer launches, and stricter dependency expectations. Provider parsing improves in parallel so the new semantic layer has better source signals for thinking/tool/context classification. The net result is a pipeline that writes more deliberately, reads more expressively, and fails more safely.

## Motivation

The previous architecture already had a unified pipeline, but it still had two important limitations.

The first was that writes were still too close to “single pass, trust the input.” Even with canonical storage, it was still too easy for malformed input, duplicate-equivalent content, or concurrent attachment activity to put pressure on the write path in ways that were harder to reason about than they should be.

The second was that the read/query side was still fundamentally row-shaped. Once the archive started to support richer browsing, server views, and library consumption, it became obvious that callers wanted to ask semantic questions:

- show only substantive messages
- skip tool chatter
- search within a projected subset
- reason about thinking traces or context dumps
- operate on conversation slices without hand-writing SQL or row filters

That is not a good fit for raw dict-like search rows and basic repository fetches. The archive needed a more expressive read model.

The branch therefore attacks both sides of the same seam: make writes staged and safer, then make reads more typed and composable.

### 1. Make the write path explicit and staged

The pipeline is structured as validate → parse → persist rather than one broad ingest pass that gradually discovers whether the input is acceptable. That has a few concrete benefits.

Validation can reject malformed JSON and provider-schema mismatches before any database state changes happen. Parsing can then build a richer in-memory tree with parent/branch checks. Persistence becomes the only stage that actually mutates storage, and it does so under one `_WRITE_LOCK` so concurrent work does not interleave attachment ref-count updates or partial batches.

That single write lock is paired with more careful attachment bookkeeping and SAVEPOINT-style atomicity around ref counts. I wanted the storage behavior to be boringly predictable even when the rest of the pipeline is concurrent.

The content-hash changes matter here too. NFC normalization makes semantically identical Unicode payloads hash the same way, which is essential if content hash is going to remain a meaningful identity surface without reflecting encoding accidents.

### 2. Add projection-oriented read models

`polylogue/lib/models.py`, `polylogue/lib/projections.py`, and the repository layer now provide higher-level access patterns over conversations and messages. The important shift is not just “more classes”; it is that callers can compose filters and views without dropping back to raw storage details immediately.

The semantic classification model gives messages categories like:

- thinking
- tool use / tool result
- context dump
- noise
- substantive content

Projections then let higher-level code compose operations like:

- messages by role
- substantive-only slices
- time-range filtering
- projected search

This makes the library/server/search layers much easier to evolve because they have a real analysis vocabulary instead of ad hoc helper predicates scattered across modules.

### 3. Preserve more structure from source payloads

The new projection layer would be much weaker if the importers continued flattening everything too early. That is why the branch also introduces structured `content_blocks` and richer importer behavior.

Claude parsing now distinguishes nested tool-use/tool-result pairs and thinking-like content more carefully. ChatGPT import paths normalize canvas / DALL-E attachment references into the shared attachment model. Drive parsing preserves useful metadata like Gemini thinking markers. Claude Code parsing becomes more faithful too.

This is a key design point: richer read models are only credible if the ingest side preserves enough signal for them to classify meaningfully.

### 4. Harden the unsafe edges

Several fixes in this branch are small individually but important collectively:

- FTS5 query escaping now quotes unsafe special-character input instead of letting syntax injection leak into search
- ZIP-bomb checks reject obviously dangerous archive inputs early
- JSON/JSONL ingest has better encoding fallbacks and null-byte cleanup
- asset extraction uses safer temp-file handling
- viewer launches quote editor/browser paths before invoking subprocesses
- metadata decoding / markdown rendering are cleaned up so the server and library layers see consistent shapes

These are the kinds of issues that become more pressing once the archive exposes richer CLI and server surfaces. I wanted them addressed before the new projections and verify behavior became the normal way to consume the data.

### 5. Tighten runtime expectations and test like we mean it

The branch stops pretending that optional fallbacks are a viable long-term story for orjson and structlog. They are required runtime dependencies, and the code stops carrying parallel codepaths for “if the modern dependency is missing.”

Reintroduce explicit Pydantic and add Hypothesis-backed test capability where the new abstractions justify it. Property-based tests are a particularly good fit here because hashing, ID generation, projection composition, and query escaping all benefit from larger input coverage than a few hand-picked examples.

## Testing

The suite additions are broad because the branch changes both architecture and safety posture. The highlights are:

- property-based coverage for hashing, IDs, query escaping, and projections
- importer tests for Claude / ChatGPT / Codex / Drive behavior
- projection and message-classification tests
- concurrency coverage for the pipeline
- verify, assets, and CLI view safety tests
- integration/server flow coverage that exercises the staged pipeline against real archive operations

This is one of the few cases where “lots of tests” is the whole point. The new abstractions need to earn trust from both correctness and safety angles.

## Notes

One thing I intentionally did not preserve is the older “best effort if fancy deps are missing” posture. Once the archive has typed projections, richer server output, and stronger verification/safety behavior, the maintenance burden of fallback codepaths outweighs the startup convenience. The stricter dependency model is part of the architecture here, not an unrelated cleanup.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 6 finding(s) confirmed fixed/stale.

PR #54, opened 2026-04-07
refactor: split archive layers behind protocols and query CLI

Sinity:
## Summary

I split Polylogue's archive surfaces into explicit contracts and moved the implementation details behind those seams. Before this branch, the CLI, pipeline, and storage code all reached into the same grab-bag of top-level modules: `polylogue/db.py`, `polylogue/search.py`, `polylogue/store.py`, `polylogue/drive_client.py`, and a handful of rendering helpers. The result was that every new feature had to know too much about SQLite, too much about which module happened to own initialization, and too much about how rendering and indexing were wired together. This branch introduces protocol boundaries for storage, search, vector search, and rendering; a dependency-injection container to centralize startup wiring; and a query-first CLI path that lets plain positional arguments execute archive searches without a subcommand. Used the refactor to add a single `Polylogue` facade, schema-v5 metadata storage, and matching service/container/backend test coverage.

## Motivation

The feature branch started as a sequence of "Priority 1/2/3/4" architecture cleanups, but the repeated pain point was the same every time: implementation modules were doubling as the public integration surface. `run`, `search`, and `view` each knew slightly different versions of how to open the database, rebuild the index, render output, and instantiate importers. That made experimentation expensive. If I wanted to add richer filters or expose the same archive operations through MCP, I first had to unwind direct imports and implicit initialization order.

I considered doing this as a thin-wrapper pass where I would leave the old modules in place and add new facades on top. The branch history pushed me away from that. Once I started moving search, store, and ingest code into dedicated packages, the compatibility layer became the noisy part of the design: duplicated imports, alias exports, and unclear ownership. I ended up preferring a more opinionated split—`storage/`, `ingestion/`, `pipeline/services/`, `rendering/`, `cli/`, and `core/facade.py`—with compatibility removed once the new surfaces were exercised.

### Protocols and backend ownership

The biggest mechanical change is the introduction of explicit runtime-checkable boundaries in `polylogue/protocols.py`.

```python
class StorageBackend(Protocol):
    def save_conversation(...) -> dict[str, int]: ...
    def get_conversation(...) -> ConversationRecord | None: ...
    def iter_conversations(...) -> Iterator[ConversationRecord]: ...
```

That contract let me move the storage implementation out of top-level modules and into `polylogue/storage/`. The branch deletes `polylogue/db.py`, `polylogue/search.py`, `polylogue/index.py`, `polylogue/index_qdrant.py`, and `polylogue/store.py`, then recreates those responsibilities under `storage/` with one clear owner per concern. `storage/backends/sqlite.py` is the concrete backend that handles schema-v5 migrations, metadata JSON, nested transactions/savepoints, and thread-local connections. `storage/repository.py` becomes the write coordinator, not a loose collection of helper functions.

The effect is not just packaging. Once callers only depend on `StorageBackend` and `SearchProvider`, the rest of the system can stop assuming "SQLite plus FTS5 plus whatever renderer happens to be imported today."

### Dependency injection and service layer

I added `polylogue/container.py` with an `ApplicationContainer` plus `create_container()` factory so the process wiring happens in one place instead of in every CLI command and helper.

```python
container = create_container()
repo = container.storage()
ingestion = container.ingestion_service()
```

The pipeline has explicit service objects in `polylogue/pipeline/services/`:

- `IngestionService` owns bounded concurrent ingest and result aggregation
- `IndexService` owns FTS/Qdrant update, rebuild, and status logic
- `RenderService` owns parallel output generation and renderer selection

This is the line that mattered architecturally: `pipeline/runner.py` now orchestrates services, rather than knowing how to construct low-level dependencies itself.

### Query-first CLI and fluent filtering

I wanted the CLI to treat the common case—"search the archive for these words"—as the default path, not as a second-class subcommand. `QueryFirstGroup.parse_args()` in `polylogue/cli/click_app.py` now intercepts positional terms and routes them into query execution while leaving named commands (`sync`, `check`, `reset`, `mcp`, `auth`) intact.

On the data side, `polylogue/lib/filters.py` introduces `ConversationFilter`, which gives both the CLI and library callers the same fluent query vocabulary:

```python
convs = (
    p.filter()
     .provider("claude")
     .contains("timeout")
     .since("2025-01-01")
     .limit(10)
     .list()
)
```

`polylogue/cli/query.py` then becomes the single place where CLI flags are translated into filter chains, output destinations, and modifier actions. That is a better long-term place to add features like metadata changes, semantic ranking, or export variants than the previous "one-off code path per command" model.

### Rendering, metadata, and MCP

This branch also lands a new `ConversationFormatter` in `polylogue/rendering/core.py`. Before that, markdown and HTML rendering each reimplemented message ordering and attachment presentation. Now both renderers consume the same structured formatter output, which should make future render bugs much easier to isolate.

Used the storage rewrite to add schema-v5 metadata support. Conversations have a JSON metadata column for user-editable title, summary, and tags. That is an important enabling change for later query modifiers because the storage layer finally has a first-class place to keep archive-local annotations separate from provider metadata.

On the integration side, `polylogue/mcp/server.py` exposes stdio tools for listing, searching, and fetching conversations. Those tools are intentionally thin wrappers around the same repository/facade layer the CLI uses; I did not want to create an MCP-only data path.

Finally, `polylogue/core/facade.py` adds the `Polylogue` class as the programmatic entry point. The new facade is deliberately small—`get_conversation()`, `search()`, `ingest_file()`, `filter()`, `stats()`—but it gives external callers one stable object instead of a pile of repositories, services, and helpers.

## Testing

I added and expanded the tests that specifically anchor the new seams:

- `tests/test_backend_protocol.py` and `tests/test_backend_sqlite.py` cover protocol conformance and SQLite backend behavior
- `tests/test_container.py` and `tests/test_cli_container.py` exercise container wiring and CLI integration
- `tests/test_pipeline_services.py`, `tests/test_pipeline_services_ingestion.py`, and `tests/test_indexing_service.py` cover the service split
- `tests/test_facade.py`, `tests/test_filters.py`, and `tests/test_rendering_core.py` cover the new user-facing programmatic and rendering surfaces
- `tests/test_pipeline_concurrent.py` and the new `tests/benchmarks/*` suite cover concurrency and cache-sensitive behavior

## Notes

Legacy import paths are intentionally not the main story of this branch. Some compatibility code was carried briefly while the feature branch was in flight, but the final state favors clear ownership over old alias exports.

Two behavior changes are worth calling out for reviewers:

1. the database path moves from state storage to data storage, because the archive is durable data, not disposable runtime state
2. positional CLI arguments now enter query mode, so argument parsing behavior in `click_app.py` deserves close review alongside shell completion behavior


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 4 finding(s) confirmed fixed/stale.

PR #58, opened 2026-04-07
test: add golden rendering checks and broaden archive health

Sinity:
## Summary

I used this branch to turn a loosely covered archive surface into something that produces sharper regressions. The work lands in four layers: optional git hooks for local workflow hygiene, expanded `polylogue check` coverage for empty conversations and FTS drift, a proper golden-test framework for rendering, and much broader semantic/provider test coverage. Hardened the SQLite setup and a few parser edges that were making concurrent and corpus-backed tests noisier than they needed to be. The common theme is not "more tests" in the abstract; it is making the failures that matter show up earlier and in a more legible form. That is especially important after the previous architectural refactors, because rendering and parser behavior now spans more seams and needed stronger contracts.

## Motivation

There were two related gaps in the test story.

First, rendering and semantic behavior were mostly verified through small assertions embedded in broader tests. That made regressions hard to diagnose. If markdown ordering changed, or Unicode started being mangled, or a tool-use block rendered differently, the failure usually appeared as a surprising string mismatch somewhere downstream rather than as a snapshot diff that told me exactly what changed.

Second, the health/check surface was too optimistic. I could create empty conversations while debugging, or blow away the FTS table during index work, and `polylogue check` still lacked a targeted opinion about those states. For a tool that relies heavily on FTS-backed queries, not noticing index drift is a real blind spot.

The branch also inherited some fragility from earlier refactors: import paths had shifted, concurrent tests occasionally tripped SQLite lock errors, and parser failures could disappear into swallowed JSON exceptions. Fixing those was part of making the new tests trustworthy.

### Optional workflow hooks

I added local git workflow aids rather than repository-enforced policy:

- `.gitconfig` now includes a handful of history/review aliases and better diff/conflict defaults
- `.githooks/commit-msg` warns on non-conventional commit subjects but does not reject the commit
- `.githooks/pre-push` warns on WIP/messy subjects and asks for confirmation before pushing

The important implementation choice is that these are soft checks. They are meant to catch avoidable mistakes in local development, not to become a second CI system or a source of surprise failures. `.githooks/README.md` documents installation and bypass so contributors can opt in explicitly.

### Health surface checks the states we actually debug

`polylogue/health.py` gains two checks that were missing in practice:

- `empty_conversations`: conversations that exist without any messages
- `fts_sync`: missing `messages_fts` table or message/FTS row-count drift

That makes `polylogue check` substantially more useful during storage/index work. A healthy database is not just "foreign keys and duplicates look fine"; it is also "the content the query layer depends on is present and in sync."

Made the supporting plumbing a little sturdier:

- `storage/backends/sqlite.py` creates parent directories before connect and sets `PRAGMA busy_timeout = 30000`
- `storage/search_providers/fts5.py` stops calling `connection_context()` with the stale signature
- `ingestion/source.py` no longer swallows `JSONDecodeError` in the final fallback path

That last change is subtle but important: if a source file is invalid, the system should count and report it, not quietly limp past it.

### Golden tests and semantic API coverage

The new golden framework is intentionally simple:

```text
tests/golden/<case>/
  input/
  expected/
  metadata.yaml
```

`tests/test_golden_outputs.py` exercises the renderer against cases that matter to humans reading archive output: simple ChatGPT dialogue, Claude thinking blocks, JSON tool-use formatting, empty-message elision, Unicode preservation, attachment formatting, and timestamp ordering. `tests/scripts/update_golden.py` provides the update path when rendering intentionally changes.

Added `tests/test_semantic_api_comprehensive.py`, which is effectively the contract test suite for the semantic/domain layer. It covers things like:

- `DialoguePair` validation
- message metadata extraction (`cost_usd`, `duration_ms`)
- conversation aggregates (`total_cost`, tags, display title)
- iteration helpers (`pairs`, `thinking`, `substantive`, `dialogue`)
- string/render helpers (`to_text`, `to_clean_text`)

That suite is valuable because it verifies archive semantics independently of the CLI and independently of raw provider parsing.

### Real-data parser tests and manual verification

`tests/test_providers_real_data.py` adds skip-gated parser tests for ChatGPT and Claude production exports. These are not always available in every environment, so they are gated rather than mandatory, but they are still worth keeping in-tree because they let us validate real export shapes when the corpus is present.

Added `tests/manual/verify_database.sh`, `verify_attachments.sh`, and `verify_filesystem.sh` for targeted local spot checks. These are not meant to replace automated tests; they exist because archive tooling sometimes benefits from a quick "inspect the actual on-disk outcome" loop.

### Test-suite cleanup and concurrency stability

The first commit in the branch unblocked a very large set of tests after import-path fallout. From there I cleaned up some patterns that were creating avoidable noise:

- import updates to `polylogue.storage.backends.sqlite`
- removal of nested connection-depth assertions that no longer matched the backend behavior
- explicit `conn.commit()` calls and early WAL setup in a few concurrent tests
- replacement of some `caplog` assertions with stdout/stderr capture where structlog formatting had made log-text checks brittle
- addition of `pytest-xdist` so the suite can scale better locally

## Testing

The work is directly exercised by:

- `tests/test_golden_outputs.py`
- `tests/test_semantic_api_comprehensive.py`
- `tests/test_providers_real_data.py`
- updated `tests/test_health.py`, `tests/test_check.py`, `tests/test_db.py`, and `tests/test_pipeline_concurrent.py`
- manual verification scripts under `tests/manual/`

## Notes

The real-data tests are intentionally skip-gated. They provide value when the production-shaped fixtures are available, but I did not want this branch to make the open-source/dev test path depend on private corpora.

For reviewers, the most important question is whether the new golden cases capture the renderer semantics we actually care about. If they do, future output changes should become much easier to review and much harder to miss.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,9 @@ build/
 # Local tool state
 .claude/
 
+# Local test datasets (real data, review before committing)
+tests/datasets/
+
 # Local shell env files
 .env
 
@@ -46,3 +49,9 @@ build/
 data
 sinex_md/
 polylogue.egg-info/
+
+# Test datasets (review before committing)
+tests/test-datasets/*.json
+tests/test-datasets/*.jsonl
+tests/test-datasets/*.md
+tests/test-datasets/*.txt
--- a/polylogue/health.py
+++ b/polylogue/health.py
@@ -170,6 +170,72 @@ def run_health(config: Config) -> HealthReport:
             )
         )
 
+        # Empty conversations (conversations with no messages)
+        empty_conv = conn.execute(
+            """
+            SELECT COUNT(*) FROM conversations c
+            WHERE NOT EXISTS (SELECT 1 FROM messages m WHERE m.conversation_id = c.conversation_id)
+        """
+        ).fetchone()[0]
+        checks.append(
+            HealthCheck(
+                "empty_conversations",
+                VerifyStatus.OK if empty_conv == 0 else VerifyStatus.WARNING,
+                count=empty_conv,
+                detail="No empty conversations" if empty_conv == 0 else f"{empty_conv} conversation(s) with no messages",
+            )
+        )
+
+        # FTS sync check (verify messages_fts table exists and is in sync)
+        try:
+            # Check if messages_fts table exists
+            fts_exists = conn.execute(
+                "SELECT name FROM sqlite_master WHERE type='table' AND name='messages_fts'"
+            ).fetchone()
+
+            if fts_exists:
+                # Count messages in both tables
+                msg_count = conn.execute("SELECT COUNT(*) FROM messages").fetchone()[0]
+                fts_count = conn.execute("SELECT COUNT(*) FROM messages_fts").fetchone()[0]
+
+                if msg_count == fts_count:
+                    checks.append(
+                        HealthCheck(
+                            "fts_sync",
+                            VerifyStatus.OK,
+                            count=fts_count,
+                            detail=f"FTS index in sync ({fts_count:,} messages indexed)",
+                        )
+                    )
+                else:
+                    checks.append(
+                        HealthCheck(
+                            "fts_sync",
+                            VerifyStatus.WARNING,
+                            count=abs(msg_count - fts_count),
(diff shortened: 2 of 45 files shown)


Sinity:
@coderabbitai full review

Sinity:

PR #60, opened 2026-04-07
refactor: extract typed provider models and declarative extraction layer

Sinity:
## Summary

Rearchitect how provider message formats are parsed and analyzed by introducing a two-tier extraction system: typed Pydantic models matching each provider's exact export shape, and a unified extraction layer that normalizes everything into `HarmonizedMessage` with rich cross-provider viewport types. Simultaneously overhaul the test suite to use parametrization, shared builders, and property-based testing — cutting test count from ~500 to ~150 while expanding actual coverage.

## Motivation

The importers currently do double duty — they parse provider exports into `ParsedConversation` for storage, but any analysis that needs tool calls, reasoning traces, or token counts has to re-parse the raw JSON with per-provider conditional logic scattered across query, projection, and rendering code. There's no shared contract for "what did this message contain semantically?" which means every new feature that touches message content (filtering by tool use, stripping thinking blocks, computing costs) requires touching every provider path.

The test situation compounds the problem. The test suite grew organically with one-assertion-per-function style, making refactoring terrifying — renaming a field in the Claude importer means updating 65 individual test functions. Tests are also tightly coupled to implementation details rather than behavioral contracts.

I considered two alternatives before landing on this approach:

1. **Extending ParsedConversation** with optional semantic fields. Rejected because it mixes archival structure (what the provider gave us) with analytical structure (what we derive from it). The two evolve at different rates.

2. **Provider-agnostic extraction only** (no typed models). Rejected because the typed models serve as executable documentation of each provider's format and enable schema drift detection — we can generate JSON schemas from real exports and validate future exports against them.

## Provider models

Each provider gets a Pydantic model in `polylogue/providers/` that mirrors the export format:

```python
class ClaudeCodeRecord(BaseModel):
    model_config = ConfigDict(extra="allow")
    type: str
    message: Optional[MessageContent] = None
    costUSD: Optional[float] = None
    durationMs: Optional[int] = None
    # ...

    def extract_reasoning_traces(self) -> list[ReasoningTrace]:
        """Pull thinking blocks from content array."""

    def extract_tool_calls(self) -> list[ToolCall]:
        """Pull tool_use blocks with category classification."""
```

The `extra='allow'` configuration is critical — when providers add new fields (which happens constantly), the models absorb them without breaking. The schema inference module (`schema_inference.py`) can then detect when a provider has added new fields by comparing generated schemas against stored baselines.

## Unified extraction

The `HarmonizedMessage` type in `polylogue/schemas/unified.py` is the analysis contract:

```python
class HarmonizedMessage(BaseModel):
    role: str
    text: str
    timestamp: Optional[datetime]
    reasoning_traces: list[ReasoningTrace] = []
    tool_calls: list[ToolCall] = []
    content_blocks: list[ContentBlock] = []
    tokens: Optional[TokenUsage] = None
    cost: Optional[CostInfo] = None
    duration_ms: Optional[int] = None
```

Each provider has a dedicated extractor (`_extract_claude_code()`, `_extract_chatgpt()`, etc.) that knows the raw format. The public entry point `extract_harmonized_message(raw_dict, provider)` dispatches to the right one. A secondary entry point `extract_from_provider_meta()` bridges from database-stored provider_meta back to `HarmonizedMessage` for post-ingest analysis.

Alongside the imperative extractors, `extractors.py` provides glom-based declarative specs for simpler field-level mapping — useful for the common case where extraction is just "reach into nested dict, rename, maybe parse a timestamp."

## Viewport types

The `viewports.py` module defines the cross-provider vocabulary:

- `ToolCall` carries name, input, output, category (`ToolCategory.FILE_WRITE`, `SHELL`, `GIT`, etc.), and computed properties like `is_file_operation` and `affected_paths`
- `ReasoningTrace` carries the raw thinking text and optional summary flag (for Claude Code compaction events)
- `ContentBlock` with `ContentType` enum handles the heterogeneous content arrays all providers use differently
- `classify_tool()` maps 30+ tool names from Claude Code, ChatGPT, Codex to categories

## Importer improvements

The Claude importer gets the biggest expansion. `parse_sessions_index()` reads Claude Code's `sessions-index.json` to enrich conversations with better titles, git branch, and project path. New semantic extractors pull structured data from content blocks:

```python
def extract_thinking_traces(content_blocks: list[dict]) -> list[dict]:
    """Extract thinking/reasoning blocks from Claude content arrays."""

def detect_context_compaction(messages: list[dict]) -> list[int]:
    """Identify summary messages that indicate context window compaction."""
```

The Codex importer is a complete rewrite — the old prompt/completion format has been replaced by an envelope format with `session_meta` and `response_item` records. The importer handles both formats and deduplicates session metadata (child sessions include parent meta).

Security gets a dedicated pass: `ParsedAttachment.sanitize_path()` blocks `../` traversal and symlink following, `sanitize_name()` strips control characters.

## Hybrid search

`HybridSearchProvider` combines FTS5 full-text search with vector similarity search using Reciprocal Rank Fusion:

```python
def reciprocal_rank_fusion(
    ranked_lists: list[list[tuple[str, float]]],
    k: int = 60,
) -> list[tuple[str, float]]:
    """Fuse multiple ranked result lists into a single ranking."""
```

The `k=60` default follows the original RRF paper. The provider is created via factory function `create_hybrid_provider()` which wires up FTS and vector backends.

## Test redesign

Three pillars:

**Shared builders** (`tests/helpers.py`): `ConversationBuilder` and `MessageBuilder` provide a fluent API for constructing test data. Export-level builders (`ChatGPTExportBuilder`, `ClaudeExportBuilder`) generate realistic file structures.

**Hypothesis strategies** (`tests/strategies/`): Custom strategies generate provider-specific messages, conversations, filters, and adversarial inputs for property-based testing. This catches edge cases that hand-written fixtures miss.

**Aggressive parametrization**: Tests that varied only in input data are collapsed into `@pytest.mark.parametrize` calls. The code detection suite goes from 101 individual tests to 15 parametrized ones. The Claude importer tests go from 65 to 12. Total test function count drops from ~500 to ~150 with equal or better behavioral coverage.

## Infrastructure

- Batch `executemany` for message, attachment, and ref insertion replaces individual INSERT loops
- Double-check lock pattern fixes async schema initialization race
- `delete_conversation()` now cleans FTS entries before removing the conversation
- Health repair system (`run_all_repairs()`) fixes orphaned messages, empty conversations, dangling FTS entries
- CLI gains `--transform`, `--dry-run`, `--force` flags and YAML/plaintext output


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 2 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #62, opened 2026-04-07
refactor: switch to singleton services and branch-aware storage

Sinity:
## Summary

I removed the dependency-injector container, replaced it with a tiny services module, and used the same branch to make conversation lineage a first-class storage concern. The storage layer can preserve raw provider payloads, session continuations, sidechains, and message branches; the repository can traverse that tree; and the domain model no longer has to load every message eagerly just to answer summary-level questions. This also changes the ingest flow: raw bytes are stored first via a new acquisition stage, and parsing/replay happens from `raw_conversations` without requiring the original provider source to still be present. The CLI follows that shift by consolidating staged execution under `run` and removing the old `sync` alias. Across all of that, I tightened role normalization and expanded test/fuzz coverage around the new storage and pipeline behavior.

## Motivation

Two architectural problems had started to reinforce each other.

The first was process wiring. The container solved the "where do dependencies come from?" question, but by this point it was more ceremony than clarity. Most commands wanted the same small set of objects, and the storage repository was already the effective center of gravity. Carrying dependency-injector everywhere was no longer buying enough structure to justify the extra layer.

The second was that the archive model still flattened away too much history. ChatGPT edit branches, Claude sidechains, and Codex session continuations were all observable in raw data, but the stored representation had only a weak place to keep that lineage. At the same time, large conversations were getting expensive to load eagerly when many callers only needed counts, summaries, or a stream.

The branch history reflects that progression. It starts with test-isolation fixes and explicit branching metadata, then pushes those fields through ingest, then adds raw conversation capture and staged acquisition, and finally removes the container once the storage/repository model had become simpler and more self-sufficient.

### Module-level services instead of the DI container

`polylogue/services.py` owns the runtime singletons:

```python
from polylogue.services import get_backend, get_repository, reset

repo = get_repository()
```

The goal here is not "global state everywhere." It is a very small, explicit composition module for the common process-local objects the application always needs. `reset()` exists so tests can clear the singleton cache and get a fresh process model.

That let me delete `polylogue/container.py` entirely and trim `polylogue/protocols.py` back to the pieces that still make sense as protocols: search, vector search, and rendering. Storage had become simpler to understand as a concrete backend plus repository than as a very large abstract protocol.

### Branch-aware storage and raw payload preservation

The biggest capability change is in `polylogue/storage/backends/sqlite.py`. Schema migrations now add:

- `parent_conversation_id`
- `branch_type`
- `parent_message_id`
- `branch_index`
- `raw_conversations`

Together, those let the archive preserve both conversation-level lineage (continuations, sidechains) and message-level lineage (edited branch siblings). Raw provider bytes are stored alongside the parsed archive data so later parsing/replay can happen from the database itself rather than from the original provider source.

That is an important shift. It means the ingest story becomes:

1. acquire raw bytes
2. persist them
3. parse from stored raw bytes
4. materialize archive rows

rather than "read provider export again and hope it is still there."

Added `iter_messages()` and `get_conversation_stats()` to the backend so callers can work at summary/stream granularity without forcing full conversation loads.

### Unified repository and lazy messages

`polylogue/storage/repository.py` and `polylogue/lib/repository.py` are consolidated into one `ConversationRepository` model that can serve both write and read use cases. New traversal helpers like `get_parent()`, `get_children()`, `get_root()`, and `get_session_tree()` make the branching data queryable instead of just stored.

On the domain side, `polylogue/lib/messages.py` introduces `MessageCollection`, and `Conversation.messages` is typed as that lazy collection rather than a plain eager list. The idea is:

```python
conv = repo.get(conversation_id)
for msg in conv.messages:
    ...
```

without paying the memory cost of loading the full message list up front unless the caller really needs to. For those cases, the eager path still exists.

This is one of the bigger review points in the branch because it changes the performance model of the domain object deliberately. Most archive consumers iterate; fewer need random indexing or immediate serialization of the whole message list.

### Strict role normalization and branch-aware importers

`polylogue/lib/roles.py` owns the canonical `Role` enum plus `normalize_role()`. Empty strings raise `ValueError` instead of silently collapsing to a default. That change forces importers to be explicit about missing-role fallbacks, which is healthier than letting invalid role data masquerade as a valid user/assistant event.

The provider importers also emit lineage data:

- ChatGPT extracts `parent_message_provider_id` and `branch_index` from mapping children arrays
- Claude carries forward `parentUuid` and marks sidechains
- Codex detects continuation lineage from multiple `session_meta` records

`prepare_ingest()` in `pipeline/ingest.py` resolves those provider IDs into internal archive IDs and persists them through `ConversationRecord` and `MessageRecord`.

### Acquisition stage and CLI `run`

I added `polylogue/pipeline/services/acquisition.py` so the pipeline can explicitly acquire and store raw provider payloads before parsing. `pipeline/services/ingestion.py` then reads from `raw_conversations` rather than from the provider source directly.

This also fits the CLI cleanup. The branch includes the staged `run` command and removes the deprecated `sync` alias. That is a breaking CLI change, but it aligns the command surface with the actual pipeline stages much better than carrying two names for the same evolving behavior.

### QA, fuzzing, and corpus-backed tests

This branch also adds a lot of verification around the new behavior:

- `tests/test_branching.py` and `tests/test_branching_pipeline.py`
- `tests/test_raw_conversations.py`
- `tests/test_pipeline_services_acquisition.py`
- `tests/test_cli_run.py`
- `tests/test_mcp_server.py`
- parser contract fixtures and seeded databases
- `tests/fuzz/*` plus `mutmut.toml`

The fuzzers target the sorts of inputs that become more important when you are preserving raw payloads and doing stricter role/path handling.

## Testing

The new storage/pipeline model is exercised by branching, acquisition, raw-conversation, run-command, and MCP suites, plus fuzz coverage for FTS escaping, path handling, JSON importers, and timestamp parsing. Kept corpus-backed fixtures in the mix so the importer changes are not only validated against tiny synthetic samples.

## Notes

Three reviewer-sensitive parts of this branch deserve extra attention.

First, confirm that the lazy `MessageCollection` contract is clear enough for callers that previously assumed a plain list.

Second, confirm that `raw_conversations` is storing enough metadata to make replay trustworthy without turning the database into a second, ambiguous source of truth.

Third, check the CLI migration from `sync` to `run` carefully. The intent is to make staged execution more explicit, but it does remove a long-standing command alias.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 5 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 2 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #64, opened 2026-04-07
feat: consolidate sources and ship sqlite-vec static archives

Sinity:
## Summary

This branch finishes three architectural moves that were all blocked on each other: source handling lives under one `polylogue/sources/` domain, semantic search no longer depends on an external Qdrant service, and archives can be published as a static HTML site without requiring a running FastAPI process. Used the migration to tighten the CLI surface, add SQL pushdown and batch-query improvements, and fix a long tail of parser/output bugs that showed up once the new surfaces were exercised against real data. The result is a more self-contained product: one source package, one SQLite database for both FTS and vector search, and one offline-friendly browsing path. The branch log is noisy because a lot of manual-testing-driven fixes landed after the big moves, but the through-line is consistent—reduce external moving parts and make the archive itself the deployable artifact.

## Motivation

Three pain points had converged.

First, source code ownership was split across `importers/`, `providers/`, and `ingestion/`. Those packages all existed to answer variants of the same question—how do we get provider data into the archive?—but they were organized by implementation detail rather than domain. It was too easy to make a parser change and then discover the corresponding provider or ingestion glue lived somewhere else entirely.

Second, vector search still depended on a Qdrant server. That was workable in development, but awkward in CI, awkward in packaging, and fundamentally at odds with the "single archive, local-first" direction of the project. If the rest of the archive already lives in SQLite, pushing vector state there too is a much better fit.

Third, browsing required a live FastAPI server. That is a poor match for an archive product. Users should be able to render their archive once and host it from a file URL or any static server.

The long sequence of fix commits later in the branch is also telling. Once I made the big package/search/site changes, real-corpus and manual testing exposed parser type assumptions, output escaping issues, recursion hazards, WAL quirks, and stale-search-index problems. I left that history reflected in the final shape instead of pretending the first pass was enough.

### Source package consolidation

The package move is simple conceptually:

```text
polylogue/importers/*   -> polylogue/sources/parsers/*
polylogue/providers/*   -> polylogue/sources/providers/*
polylogue/ingestion/*   -> polylogue/sources/*
```

That lets everything related to "finding provider data, validating it, and parsing it" live under one domain. Absorbed adjacent single-purpose modules while I was here:

- analytics moves into `polylogue/cli/analytics.py`
- schema validation moves into `polylogue/schemas/validator.py`
- the old server package is removed

The CLI follows suit. `serve` is deleted, `dashboard` stays as the interactive local browsing surface, and `site` is added for static generation. Trimmed and rationalized some query flags/exports along the way so the command surface matches the new package ownership more closely.

### sqlite-vec replaces Qdrant

`polylogue/storage/search_providers/sqlite_vec.py` introduces `SqliteVecProvider`, which stores Voyage embeddings in SQLite vec0 tables. The provider handles:

- binary float32 serialization via `_serialize_f32()`
- batched embedding calls
- KNN queries with provider filtering
- extension availability checks

The relevant configuration also changes. `IndexConfig` drops Qdrant-specific fields and gains:

- `voyage_model`
- `voyage_dimension`
- `auto_embed`

That aligns the config with the actual remaining moving pieces: the embedding service and whether we should embed eagerly.

On the storage side, schema migrations move from v9 to v11. The important pieces are:

- `embeddings_meta` and embedding-status tracking
- vec tables when sqlite-vec is available
- FTS UPDATE/DELETE triggers so edited or deleted messages stop leaving stale rows behind
- bulk helpers like `count_conversations()` and `get_conversations_batch()`

This is one of the most valuable behavior changes in the branch. Full-text and vector state live inside the same SQLite deployment story instead of needing a second service.

### Static site builder and HTML rendering

`polylogue/site/builder.py` adds `SiteBuilder`, and the CLI gets a dedicated `site` command:

```bash
polylogue site --search-provider pagefind --dashboard
```

The builder generates archive index pages, provider pages, conversation pages, and search assets for either pagefind or lunr. That gives the archive an offline/publication path that does not require an application server.

Rewrote the HTML renderer to use Pygments for code highlighting and to share theme behavior with the broader HTML/static-site surfaces. That means code blocks rendered in per-conversation HTML and code blocks rendered in the generated site behave more consistently, which should make future output changes easier to reason about.

### Query and performance improvements

The branch is not only packaging and new features. It also tightens the query path and some large-archive performance behavior.

`ConversationFilter._sql_pushdown_params()` pushes provider/date/title constraints into SQL instead of filtering everything in memory. The backend picks up bulk fetch helpers, and the ingestion service batches raw records more aggressively. These are modest changes individually, but together they make the bigger archive flows a lot less wasteful.

### Parser and runtime hardening

A large share of the branch is correctness work discovered by real-corpus and manual testing:

- ChatGPT parser guards against `author` being a plain string
- Claude Code parser stops assuming cost fields are numeric
- Codex role handling now normalizes raw role variants
- `sources/source.py` caps recursive parse depth and adds JSONL fallback handling
- plain-text output paths stop leaking Rich markup
- site builder path handling is sanitized
- MCP limit validation is made explicit
- FTS/index state is kept more honest after edits/deletes
- timestamp/logging/error-reporting edges are surfaced instead of disappearing silently

I am calling this out because it explains the branch shape. Once sqlite-vec and the site builder existed, the code got exercised in new ways and the weaker assumptions showed up quickly.

### Tests, fuzzing, and benchmarks

The testing work is intentionally broad here because the branch changes a lot of foundation:

- `tests/test_sqlite_vec.py` covers the new vector provider
- site-builder and dashboard tests cover static/archive UI generation
- `tests/fuzz/` adds atheris fuzzing for FTS escaping, path sanitization, importer parsing, and timestamp handling
- `tests/benchmarks/` adds performance coverage
- parser regression tests and CLI/unit updates cover the long correctness tail

## Testing

The branch adds or expands coverage for sqlite-vec, site generation, dashboard/TUI behavior, parser edge cases, security-critical editor/path handling, and benchmark/fuzz scenarios. Used the manual-testing cycle to land targeted regressions rather than leaving those findings undocumented in commit history.

## Notes

The most important review areas are:

1. the source-package move itself—whether ownership is genuinely clearer now
2. sqlite-vec integration and its failure behavior
3. the static site builder's output and path handling
4. the config/API fallout from removing Qdrant-specific assumptions

This is also intentionally a breaking branch for some surfaces. `serve` is gone, Qdrant config is gone, and old import paths under `importers/`, `providers/`, and `ingestion/` are no longer the supported integration path.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 3 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #66, opened 2026-04-07
feat: expose tag inventory and make metadata edits atomic

Sinity:
## Summary

I wanted tag management to be discoverable instead of only writable. At the start of this branch Polylogue could add and remove tags on conversations, but there was no cheap way to ask "what tags already exist?" and the storage path behind those edits was still a plain JSON read/modify/write cycle with no explicit protection against competing writers. I fixed that by pushing tag aggregation into SQLite with `json_each(...)` and by routing metadata/tag mutations through a transactional helper that uses `BEGIN IMMEDIATE` at the outer boundary and `SAVEPOINT` when the caller is already inside a write. Once that work was in place, I used the same branch to harden the other archive surfaces that depend on sparse message records: query output, repair commands, renderers, provider parsers, concurrent ingestion, and sqlite-vec error paths. The diff looks broad, but the work is actually pretty coherent: the new tag surface forced me to tighten the places where missing role/text fields or non-atomic metadata could still produce misleading or fragile behavior.

## Motivation

I considered landing `polylogue tags` as a very small CLI feature and leaving the rest for later, but that would have put a nicer read surface on top of metadata writes I still did not trust. `add_tag()`, `remove_tag()`, `update_metadata()`, and `delete_metadata()` all revolved around the same pattern: fetch the JSON blob, mutate it in Python, and write it back. That is tolerable when there is exactly one writer and no nesting, but it gets shakier once the archive supports concurrent ingestion, batch mutation, repair commands, and repository helpers that may already be inside a transaction.

The later commits kept reinforcing the same theme. Every time I wrote or expanded a regression, I ran into one of two classes of bug:

1. sparse provider data: `role=None`, `text=None`, empty content blocks, or odd timestamp fallbacks
2. archive-scale operations that were still written in the convenient but expensive SQL shape

Did not want tag discovery to materialize the entire archive just to answer a count question. If the command is going to be useful in the CLI, it should be backed by SQL aggregation, not a Python loop over conversations.

### Tag discovery and metadata mutation

I added a dedicated `polylogue tags` command in `polylogue/cli/commands/tags.py` and wired it into `polylogue/cli/click_app.py`. The command stays intentionally narrow: list tags with counts, optionally scope by provider, optionally emit JSON, optionally truncate to the top `N`.

The backing query lives in `SQLiteBackend.list_tags()` and uses SQLite's JSON1 support:

```sql
SELECT tag.value AS tag_name, COUNT(*) AS cnt
FROM conversations,
     json_each(json_extract(metadata, '$.tags')) AS tag
WHERE metadata IS NOT NULL
  AND json_extract(metadata, '$.tags') IS NOT NULL
GROUP BY tag.value
ORDER BY cnt DESC
```

That gave me a cheap read path, but the more important work is underneath it. Metadata mutation runs through an explicit transaction boundary in `_metadata_read_modify_write()`. The helper takes the outer write lock when necessary and falls back to a savepoint if a caller is already inside a transaction:

```python
BEGIN IMMEDIATE  # outer metadata write
...
SAVEPOINT metadata_rmw  # nested metadata write
...
RELEASE / ROLLBACK TO SAVEPOINT
```

That is what makes tag and metadata changes safe to expose more aggressively from the CLI. It also gives the storage tests a single place to pin the semantics without scattering the same correctness expectation across `add_tag()`, `remove_tag()`, and key-value metadata updates.

### Query UX and machine-readable output

A lot of the user-visible polish here is individually small, but the cumulative effect matters. `cli/query.py` has a lightweight `--count` path, clearer no-results output through `_describe_filters()` and `_no_results()`, full-content JSON for single conversations via `_conv_to_json()`, and cleaner CSV behavior for list mode.

Switched the stdout path to `click.echo()` so machine-readable output is not coupled to the Rich console implementation. That interacts with another branch fix: `PlainConsole` now strips markup before printing, so CI logs and plain-mode output do not leak `[bold]`, `[green]`, or other Rich tags.

Bulk operations changed in a deliberate way as well. Instead of requiring callers to know ahead of time that they need `--force`, the delete and modifier flows now present an interactive confirmation path. That is safer in interactive use and also made the code easier to reason about because preview and confirmation sit on the same branch.

### Repair paths and archive-scale operations

`health.py` had two separate weaknesses. Some repair queries were written with `NOT IN`, which becomes the wrong shape once the tables get large or `NULL`s become relevant, and the dry-run logic for FTS synchronization was doing more work than needed. I replaced those with correlated `NOT EXISTS` forms and a row-count-based dry-run comparison.

Tightened the role repair logic. The previous path could collapse any unknown role into `tool`, which is too blunt for Claude Code archives that have progress or system-like records with incomplete chat-role information. The new repair path inspects provider metadata through `json_extract(...)` and maps more precisely based on the source record type.

The branch also makes sqlite-vec failure modes explicit. Extension-load problems are no longer silently swallowed, and `httpx` exception text is sanitized so caller-visible errors do not echo full URLs or credentials.

### Sparse record handling across renderers, parsers, and ingestion

By the middle of the branch I had fixed essentially the same null-handling bug in markdown, org, HTML, the static site, the MCP server, and the TUI. I made the policy explicit everywhere: missing roles serialize as `"unknown"`, missing text is skipped rather than rendered as an empty section, and empty stream headers are suppressed.

On the parser side I tightened several provider-specific cases:
- ChatGPT/Gemini text extraction now tolerates `None` and structured parts more defensively.
- ISO timestamp fallback runs through the shared parser instead of assuming a single shape.
- expected provider validation skips get logged at DEBUG not surfacing as noisy warnings.

In `pipeline/services/ingestion.py` I also stopped one failed future from poisoning the rest of the concurrent batch. That bug is easy to miss until you run parallel imports across mixed-quality data, but once it appears the failure mode is bad: a single parse error silently prevents the rest of the futures from being observed.

## Testing

This branch is very regression-driven. The main additions are concentrated in:

- `tests/test_cli_run.py` for the new `tags` command and embed-path regressions
- `tests/test_storage.py` for tag aggregation, metadata atomicity, DB path handling, and SQL edge cases
- `tests/test_health.py` for repair functions and dry-run behavior
- `tests/test_pipeline_concurrent.py` for failed-future isolation
- `tests/test_provider_models.py` and related parser tests for sparse provider payloads
- `tests/test_ui_wrapper.py` for plain-console markup stripping

Where possible I tested the seam where the bug actually lived instead of only driving it through the public CLI. For example, the metadata work is primarily pinned at the storage helper boundary because that is where correctness lives.

## Notes

A few things are intentionally out of scope. I did not add tag normalization or hierarchy semantics; the new command reports exactly what is stored. `PRAGMA integrity_check` remains opt-in behind `--deep` because it is still the right tool, just not the right default. And while the renderers now agree on missing role/text handling, there is still room to reduce duplication further by pushing more of that policy into shared helpers later.

### DIFF
--- a/polylogue/cli/click_app.py
+++ b/polylogue/cli/click_app.py
@@ -117,6 +117,7 @@ def _handle_query_mode(ctx: click.Context) -> None:
     has_filters = any(
         params.get(k)
         for k in (
+            "conv_id",
             "contains",
             "exclude_text",
             "provider",
@@ -144,8 +145,18 @@ def _handle_query_mode(ctx: click.Context) -> None:
         )
     )
 
-    # Stats mode: no query terms, no filters, and no explicit output mode
-    if not query_terms and not has_filters and not has_output_mode:
+    # Modifier flags that require query execution
+    has_modifiers = any(
+        params.get(k)
+        for k in (
+            "add_tag",
+            "set_meta",
+            "delete_matched",
+        )
+    )
+
+    # Stats mode: no query terms, no filters, no output mode, and no modifiers
+    if not query_terms and not has_filters and not has_output_mode and not has_modifiers:
         _show_stats(env, verbose=params.get("verbose", False))
         return
 
--- a/polylogue/cli/click_app.py
+++ b/polylogue/cli/click_app.py
@@ -138,6 +138,7 @@ def _handle_query_mode(ctx: click.Context) -> None:
             "list_mode",
             "limit",
             "stats_only",
+            "count_only",
             "stream",
             "dialogue_only",
         )
@@ -207,6 +208,7 @@ def _show_stats(env: AppEnv, *, verbose: bool = False) -> None:
 @click.option("--fields", help="Fields for list/json: id, title, provider, date, messages, words, tags, summary")
 @click.option("--list", "list_mode", is_flag=True, help="Force list format")
 @click.option("--stats", "stats_only", is_flag=True, help="Only statistics, no content")
+@click.option("--count", "count_only", is_flag=True, help="Print matched count and exit")
 @click.option(
     "--stats-by",
     "stats_by",
@@ -260,6 +262,7 @@ def cli(
     fields: str | None,
     list_mode: bool,
     stats_only: bool,
+    count_only: bool,
     stats_by: str | None,
     open_result: bool,
     transform: str | None,
--- a/polylogue/cli/click_app.py
+++ b/polylogue/cli/click_app.py
@@ -22,6 +22,7 @@ from polylogue.cli.commands.mcp import mcp_command
 from polylogue.cli.commands.reset import reset_command
 from polylogue.cli.commands.run import run_command, sources_command
 from polylogue.cli.commands.site import site_command
+from polylogue.cli.commands.tags import tags_command
 from polylogue.cli.formatting import announce_plain_mode, should_use_plain
 from polylogue.cli.types import AppEnv
 from polylogue.ui import create_ui
@@ -299,6 +300,7 @@ def cli(
         polylogue run       Ingest/render/index pipeline
         polylogue check     Health check and repair
         polylogue embed     Generate vector embeddings
+        polylogue tags      List tags with counts
         polylogue site      Build static HTML archive
         polylogue sources   List configured sources
         polylogue mcp       Start MCP server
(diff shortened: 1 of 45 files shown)


Sinity:
@coderabbitai full review

Sinity:
@codex review

PR #68, opened 2026-04-07
fix: extract structured thinking and guard sparse archive paths

Sinity:
## Summary

I followed a regression sweep here instead of a single feature thread, because the failures all came from the same pattern: real exports contain `None` text, typed provider objects, odd `parts` shapes, mixed timestamp types, and security-sensitive filenames that the happy-path code never exercised. The biggest behavioral change is in `Message.extract_thinking()`: it now treats structured `content_blocks`, Gemini `isThought`, and ChatGPT thinking content as first-class reasoning traces not only looking for XML-like tags embedded in text. Around that I fixed the support gaps that were making those richer records unusable in practice: dropped Claude Code token/model metadata, lost `tool_result` content, site-generation crashes, hidden async limits, corrupt-cache fallbacks, and path-safety issues in Drive and site output. The branch is test-heavy because these were all edge conditions that only stayed visible once I pinned them down with direct regressions. The net result is less glamorous than a new feature, but it makes the archive much more honest about the data providers actually emit.

## Motivation

The starting point was output coverage. `_conv_to_json()` and `_conv_to_csv_messages()` were not wrong on normal conversations, but once I started writing explicit tests around sparse records the branch widened quickly. A few examples:

- ChatGPT content extraction assumed `parts` was a list and could iterate it safely.
- `ClaudeCodeRecord.to_meta()` only looked at dict-backed message payloads, so typed assistant messages could silently lose token usage and model information.
- `parse_code()` could emit `text=None` and also drop the actual payload from `tool_result` blocks, which made later consumers see a shell of the event, not the data that mattered.
- the site builder could die on a single bad conversation and take the rest of the build with it.
- facade sorting and async list calls were making hidden assumptions about timestamp types and implicit limits.

Once I had enough failing tests in place, it stopped making sense to treat these as isolated bugs. They are all symptoms of the same issue: the archive layer still expects provider data to look cleaner and more uniform than it actually does.

### Structured thinking extraction

The most important change is in `polylogue/lib/models.py`. `Message.extract_thinking()` used to be heavily biased toward one legacy representation: XML-style thinking tags embedded in the message text. That misses a lot of current provider output.

I changed the extraction order so that structured data wins first:

```python
if content_blocks:
    ...  # collect type == "thinking"
if gemini_is_thought:
    return msg.text or ""
if chatgpt_content_type in {"thoughts", "reasoning_recap"}:
    return msg.text or ""
# only then fall back to XML-ish text tags
```

That matters because once a provider already tells us "this block is thinking", searching the text for ad-hoc delimiters is the wrong abstraction. Join multiple structured thinking blocks with double newlines so the output remains readable when a model emits more than one reasoning segment.

In the same area I tightened role fallback logic. Whitespace-only roles now normalize to `"unknown"` instead of surviving as an empty string that later code has to special-case.

### Provider extraction and metadata preservation

There are two important provider-specific fixes here.

First, `extract_chatgpt_text()` in `polylogue/schemas/unified.py` now guards against non-list `parts` values. That sounds tiny, but it closes a nasty class of bugs: integers, strings, or `None` values in `parts` should not trigger `TypeError` or accidental character-by-character iteration. They should degrade to a reasonable string or empty value.

Second, `ClaudeCodeRecord.to_meta()` now extracts token usage and model information from typed `ClaudeCodeMessageContent` instances as well as raw dict payloads. Without that, typed assistant messages look valid at the object level but still arrive at the archive layer missing their cost/model metadata.

### Claude parser and site hardening

`polylogue/sources/parsers/claude.py` now preserves `tool_result.content` and `tool_result.is_error` inside `content_blocks`, and it guarantees `text` is a string by falling back to `""`.

That combination matters more than it sounds. A tool-result record without its `content` field is almost useless for later inspection, and `text=None` has a habit of surfacing far from the parser that emitted it. The new parser behavior makes tool execution records much easier to render, serialize, and re-analyze later.

The site builder change is deliberately defensive rather than fancy. Each conversation page generation is wrapped in `try/except`, failed pages are counted, and the build continues. I would rather ship a site with explicit skipped-page warnings than fail the whole archive because one conversation contains a weird record.

### Facade, health, acquisition, and vector helpers

This branch also closes a handful of correctness gaps that became obvious once the tests widened:

- `AsyncPolylogue.list_conversations()` no longer imposes a hidden `50` default when the caller passes `limit=None`.
- facade sorting uses a UTC-aware fallback instead of letting mixed datetime/string records explode at comparison time.
- health-cache deserialization falls back to a fresh check when the cache payload is corrupt.
- acquisition progress callbacks are fired outside the per-record `try/except`, so progress still advances even when individual records fail.
- sqlite-vec availability is probed lazily and cached, which avoids unnecessary connection work in edge paths.

None of these changes are especially large in isolation. Together they remove a lot of quiet inconsistency from the archive API.

### Path safety and credential handling

Folded in the path-safety fixes because they touch the same sparse/edge-data boundary. The site builder and Drive download path now sanitize provider-derived filename components, and the Drive token file is written with `0600` permissions. That is not a deep auth redesign; it is the minimum hardening needed so that provider-derived names and cached credentials do not take unnecessary liberties with the filesystem.

## Testing

Most of the branch weight is in explicit regressions. The main additions are:

- `tests/test_cli_query.py` for `_conv_to_json()` / `_conv_to_csv_messages()`
- `tests/test_importers_unit.py` for `parse_code()` content and `text` guarantees
- `tests/test_storage.py` plus `tests/test_sqlite_backend.py` for metadata operations and SQLite edge paths
- `tests/test_click_app_routing.py` for query-routing regressions
- `tests/test_claude_code_record.py` for typed Claude Code metadata extraction
- `tests/test_sqlite_vec_operations.py` for vector-provider behavior
- `tests/test_filters_advanced.py` and `tests/test_repository_operations.py` for composition and traversal behavior
- `tests/test_unified_extraction_edge_cases.py` for provider extraction oddities

The test philosophy here is to hit the function that actually owned the bug. For example, I did not only verify "thinking shows up somewhere" through a high-level path; I verified `extract_thinking()` precedence directly.

## Notes

I did not try to normalize every possible provider irregularity in one branch. The goal was narrower: make the archive stop lying by omission or crashing on data that is already common enough to be worth testing. There is still room to reduce duplication across parser/model helpers, and the site builder could eventually expose skipped-page details more explicitly than just logging a warning count. For now, the important thing is that the edge paths are visible, testable, and no longer surprising.

### DIFF
--- a/polylogue/async_facade.py
+++ b/polylogue/async_facade.py
@@ -177,7 +177,7 @@ class AsyncPolylogue:
         """
         conv_records = await self._backend.list_conversations(
             provider=provider,
-            limit=limit or 50,
+            limit=limit,
         )
 
         # Fetch messages for each conversation in parallel
--- a/polylogue/cli/helpers.py
+++ b/polylogue/cli/helpers.py
@@ -143,7 +143,7 @@ def print_summary(env: AppEnv, *, verbose: bool = False) -> None:
     ui.summary("Polylogue", lines)
 
     # Show analytics visualization (compatible with plain mode too)
-    if True:
+    if ui:
         try:
             from polylogue.cli.analytics import compute_provider_comparison
 
--- a/polylogue/facade.py
+++ b/polylogue/facade.py
@@ -30,6 +30,7 @@ Example:
 from __future__ import annotations
 
 import logging
+from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -336,10 +337,12 @@ class Polylogue:
             for conv in convs:
                 print(f"{conv.id}: {conv.title}")
         """
-        # Use repository's list method with filters
+        # When source filter is active, we need all candidates (source is in
+        # provider_meta, not a SQL-pushable column).  Otherwise respect limit.
+        fetch_limit = limit if not source else None
         all_conversations = self._repository.list(
             provider=provider,
-            limit=limit or 50,
+            limit=fetch_limit,
             offset=0,
         )
 
@@ -478,9 +481,11 @@ class Polylogue:
             total_words += sum(m.word_count for m in conv.messages)
 
         # Get recent conversations (top 5 by date)
+        # Use datetime.min as fallback to avoid mixed datetime/str TypeError
+        _epoch = datetime.min
         recent = sorted(
             conversations,
-            key=lambda c: c.updated_at or c.created_at or c.id,
+            key=lambda c: c.updated_at or c.created_at or _epoch,
             reverse=True,
         )[:5]
 
--- a/polylogue/facade.py
+++ b/polylogue/facade.py
@@ -30,7 +30,7 @@ Example:
 from __future__ import annotations
 
 import logging
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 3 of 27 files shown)


Sinity:
@coderabbitai full review

Sinity:
@codex review

PR #72, opened 2026-04-07
fix: shrink health scans and harden sparse archive reads

Sinity:
## Summary

I started this branch by chasing two annoyances — flaky stderr handling in tests and a couple of sparse-record parser misses — but it quickly converged on a clearer goal: archive inspection paths should stay cheap on large databases and predictable on incomplete data. The highest-impact work is in `polylogue/health.py`: the health check no longer counts against the FTS virtual table directly and it narrows orphan detection before counting dependent rows, which removes the worst scan pattern from large archives. Around that I added `display_date` fallbacks for conversations that only have `created_at`, pruned derived directories out of source iteration, preserved Claude Code system-record text, made stream mode honest about incompatible flags, and fixed the logging/service-reset behavior that was making tests and redirected stderr brittle. The test delta is unusually large because this branch also raises the confidence floor: the new behavior is pinned with enough coverage to justify a 90% threshold. So while the diff contains a lot of tests, the branch is not "tests for their own sake"; it is a maintenance pass aimed at making the inspection and repair surfaces dependable under real archive conditions.

## Motivation

There were two concrete problems driving this work.

The first was performance. The health-check path was doing some things the easy way rather than the scalable way, especially around FTS row counting and orphan detection. Those choices are easy to miss on small developer databases and painful on real archives. An archive maintenance command should be one of the safest things to run on a big database, not the path most likely to make a user wait.

The second was sparse-data behavior. Several parts of the system still assumed the presence of fields that are often absent in real provider data: `updated_at`, `message`, or text in Claude Code system records. Logging had a similar hidden assumption: structlog was effectively capturing whatever `sys.stderr` looked like at configuration time, which made redirected handles and test isolation much harder than they needed to be.

Once the branch picked up momentum, it became obvious that the fixes needed to land together. A faster health check without better diagnostics still leaves corrupt rows hard to understand. Better parser fallbacks without coverage still makes the next edge case easy to regress.

### Health-check query shape

The biggest behavioral change is in `polylogue/health.py`. I replaced the expensive count against the FTS virtual table with a count against `messages_fts_docsize`, and I changed orphan detection to a two-step process:

1. find the distinct orphan conversation ids
2. count messages only for those ids

That sounds modest, but it changes the shape of the work from "scan the whole thing and compare everything" to "identify the broken subset first and then inspect it." That is exactly the right tradeoff for a maintenance command.

Kept deep integrity scanning behind an explicit flag. The goal here is not to remove expensive verification; it is to make the default path reflect the questions users usually ask first.

### Sparse dates and parser fidelity

The next cluster of fixes is about records that are valid enough to store but too sparse for the display path that reads them back.

I added `display_date` to both `Conversation` and `ConversationSummary` so the display layer can fall back from `updated_at` to `created_at` without exploding on a `NoneType`. That change is deliberately narrow: I did not want every formatter to grow its own fallback logic.

On the parser side, Claude Code system records can carry text in a top-level `content` field even when there is no `message` object. The old extraction path treated those as empty. The new property path preserves that text, which makes compact-boundary and local-command records visible instead of silently blank.

Tightened JSON parse errors so the exception includes the field and record context. Corrupt rows are rare, but once one shows up the caller needs to know which blob actually failed.

### Logging, service reset, and stream honesty

The stderr fix has two parts. `configure_logging()` runs early in the CLI callback, and the logger factory writes through a tiny proxy object that delegates to the current `sys.stderr` on every call. That avoids the stale-handle problem where a cached logger keeps writing to an already-redirected stream.

`services.reset()` now closes an open backend before clearing global references. This is mostly a test-isolation fix, but it also makes the singleton reset path behave like a real teardown instead of a reference drop.

In `cli/query.py` I made `--stream` stricter and more honest. It respects `--limit`, and it warns when `--transform` or non-stdout `--output` values will be ignored. I would rather make stream mode slightly noisier than let it silently do something other than what the caller asked.

### Source iteration and TUI resilience

Source traversal now prunes directories that are clearly not raw conversation inputs: `analysis`, `__pycache__`, `.git`, and `node_modules`. That keeps ingestion focused on archive material instead of wandering into derived or irrelevant trees.

On the TUI side, `Browser.load_tree()` now tolerates repository/database errors instead of blowing up the whole tree load. The intent is not to hide real failures; it is to let the UI degrade into an error state rather than crash outright.

### Coverage work

A huge amount of this branch is coverage, but the coverage has a purpose. It lands in the places where the bugs actually were:

- `tests/test_mcp_server.py` and provider-edge suites
- source iteration and schema inference tests
- CLI integration and formatting tests
- repository stats and Drive integration coverage
- embed/run command coverage
- query execution and print-summary analytics
- UI facade and parser behavior

That test breadth is what made it reasonable to raise `fail_under` to 90. Without it, the maintenance work would still be fragile.

## Testing

The changed behavior is primarily covered by:
- `tests/test_mcp_server.py`
- `tests/test_provider_edge_cases.py`
- `tests/test_source_iteration.py` and related ingestion/schema tests
- `tests/test_click_app_integration.py`, query execution tests, and CLI helper suites
- repository, Drive, embed, and run command coverage
- UI facade/browser coverage

Verified that the new logging setup behaves correctly under stderr redirection, because that was one of the underlying pain points that started the branch.

## Notes

I did not try to rewrite the health subsystem from scratch. The goal was to change the expensive/default query shapes, improve diagnostics, and make sparse records readable. There is still room to do more around cache invalidation policy and end-to-end TUI recovery behavior, but the critical maintenance path is much saner now: large archives do not pay unnecessary scan costs, sparse records still display a date, and redirected stderr no longer corrupts logging behavior.

### DIFF
--- a/polylogue/cli/click_app.py
+++ b/polylogue/cli/click_app.py
@@ -23,6 +23,7 @@ from polylogue.cli.commands.reset import reset_command
 from polylogue.cli.commands.run import run_command, sources_command
 from polylogue.cli.commands.site import site_command
 from polylogue.cli.commands.tags import tags_command
+from polylogue.lib.log import configure_logging
 from polylogue.cli.formatting import announce_plain_mode, should_use_plain
 from polylogue.cli.types import AppEnv
 from polylogue.ui import create_ui
@@ -321,6 +322,9 @@ def cli(
 
     Run `polylogue <command> --help` for subcommand details.
     """
+    # Set up logging early so all output goes to stderr
+    configure_logging(verbose=verbose)
+
     # Set up environment
     use_plain = should_use_plain(plain=plain)
     env = AppEnv(ui=create_ui(use_plain))
--- a/polylogue/cli/query.py
+++ b/polylogue/cli/query.py
@@ -530,11 +530,14 @@ def _output_stats_by(env: AppEnv, results: list[Conversation], dimension: str) -
         if dimension == "provider":
             key = conv.provider or "unknown"
         elif dimension == "month":
-            key = conv.updated_at.strftime("%Y-%m") if conv.updated_at else "unknown"
+            dt = conv.display_date
+            key = dt.strftime("%Y-%m") if dt else "unknown"
         elif dimension == "year":
-            key = conv.updated_at.strftime("%Y") if conv.updated_at else "unknown"
+            dt = conv.display_date
+            key = dt.strftime("%Y") if dt else "unknown"
         elif dimension == "day":
-            key = conv.updated_at.strftime("%Y-%m-%d") if conv.updated_at else "unknown"
+            dt = conv.display_date
+            key = dt.strftime("%Y-%m-%d") if dt else "unknown"
         else:
             key = "all"
         groups[key].append(conv)
@@ -639,7 +642,7 @@ def _format_list(
     else:
         lines = []
         for conv in results:
-            date = conv.updated_at.strftime("%Y-%m-%d") if conv.updated_at else "unknown"
+            date = conv.display_date.strftime("%Y-%m-%d") if conv.display_date else "unknown"
             raw_title = conv.display_title or conv.id[:20]
             title = (raw_title[:47] + "...") if len(raw_title) > 50 else raw_title
             msg_count = len(conv.messages)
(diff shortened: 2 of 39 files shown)


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 4 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550).

Sinity:
@codex review

PR #74, opened 2026-04-07
fix: route filtered streams correctly and guard provider edges

Sinity:
## Summary

I closed the routing and provider bugs that showed up under audit — filtered `--stream` calls bypassing their own filters, `--delete` allowing unscoped full-archive wipes, title searches treating `%` and `_` as wildcards, attachment ref counts surviving cascade deletes, provider parsers leaking metadata noise into text, and role/timestamp normalization drifting across modules — and then used the same branch to consolidate the test surface that was supposed to catch those issues. The result is a branch with two visible threads: a correctness pass across CLI/storage/parser paths, and a substantial reduction in test duplication through parametrization and file consolidation. I kept those threads together on purpose. The audit bugs were hard to reason about partly because the coverage was scattered across many tiny files with overlapping fixtures and duplicated constructors. Pulling the shared production utilities into one place and reshaping the tests made the fixes easier to express and much harder to regress. So although the diff is large, the through-line is simple: make edge-case routing deterministic, make provider normalization consistent, and give the test suite a shape that can actually defend those guarantees.

## Motivation

The branch began with a focused audit and quickly exposed a pattern I have run into before in Polylogue: the code is usually right on happy-path inputs, but the wrong thing happens in small, quiet ways when the inputs are slightly more literal or slightly more provider-specific than expected.

A few examples that drove the work:
- `--stream --latest -p claude` could ignore the filter and stream the globally latest conversation, not the latest Claude conversation.
- `--delete` had no hard requirement that at least one filter be present, which is a bad shape for a destructive archive command.
- title filters used `LIKE` without escaping wildcard characters, so a literal search for `100%` or `100_` did not actually mean what the caller typed.
- ChatGPT multimodal/tether payloads could stringify dict parts not extracting the text-bearing fields.
- `normalize_role()` and timestamp parsing logic had drifted between providers, so one provider would preserve an odd source string while another normalized the same shape to `"unknown"`.

The audit also made a tooling problem obvious. A large chunk of the tests covered the right things, but in a highly fragmented way. It was too easy to fix a bug in one place and leave a near-duplicate behavior unreviewed in another file. If I wanted the routing and normalization fixes to stick, I needed to collapse the test surface into something denser and more legible.

### Query routing and destructive-guard fixes

The most user-visible bug here is in `polylogue/cli/query.py`. Stream mode resolves through the active filter chain when filters are present. That means `--stream` no longer has a private shortcut to "global latest" when the caller has already constrained the result set.

The logic is intentionally explicit:

```python
elif params.get("latest"):
    summaries = filter_chain.list_summaries()
elif _describe_filters(params):
    summaries = filter_chain.sort("date").limit(1).list_summaries()
```

That is the right behavior because stream mode is not a different query language. It is just a different output mode. Once the caller has supplied provider, tag, date, or other filters, the stream target needs to come from the same filtered selection path as non-stream output.

Added the `--delete` safety guard. If there is no filter, the command now exits with an error without proceeding toward a full-archive delete path. That is one of those cases where a strict guard is better than a clever prompt. The command should refuse the request before it even gets to the confirmation layer.

The YAML fix in the same file is smaller but important: field selection checks comma-separated field names exactly, so `messages_count` does not accidentally satisfy the `"messages"` membership test and pull full bodies into YAML output.

### Literal storage semantics and cleanup correctness

`polylogue/storage/backends/sqlite.py` picks up several low-level correctness fixes.

The first is title matching. The `title_contains` path now escapes `%`, `_`, and `\` before building the `LIKE... ESCAPE '\'` clause. Without that, literal titles behave like patterns:

```sql
title LIKE ? ESCAPE ''
```

That is not just a correctness footnote. The CLI advertises a literal title substring filter, so the backend has to honor the user's exact input instead of turning some characters into SQL wildcards.

The second is attachment cleanup after conversation deletion. The old path could delete the conversation, rely on CASCADE, and then remove attachments based on stale `ref_count` values. That is fine until an attachment is shared. The new path collects affected attachment ids, recalculates the counts from `attachment_refs`, and only deletes the ones that are actually orphaned.

Used the branch to reduce duplication in row mapping and filter building. `_row_to_conversation()`, `_row_to_message()`, `_row_to_raw_conversation()`, and `_build_conversation_filters()` do not add new user-visible behavior by themselves, but they matter because the old duplication was part of why subtle edge behavior drifted between sync/async paths and different list/count helpers.

### Provider parsing and shared normalization

A lot of the audit bugs were really normalization bugs.

In the ChatGPT parser I now ignore dict parts that are metadata-only (for example image pointers) and extract `.text` from tether-quote-like dicts instead of stringifying the whole object. That keeps extracted message text faithful to what a user would actually read.

In `polylogue/sources/source.py` I tightened two detection paths:
- Claude AI ZIP handling only processes `conversations.json` rather than consuming metadata files and other archive noise.
- path-based provider detection recognizes both `claude-code` and `claude_code`.

Fixed `_iter_json_stream()` so it does not retry after partially succeeding and end up yielding duplicate objects.

Role and timestamp normalization are more centralized. `normalize_role()` returns `"unknown"` for truly unrecognized roles instead of lowercasing arbitrary source strings, and all providers move toward the shared timestamp parser. That is a subtle but important contract change: normalized values should be comparable across providers, not lightly processed source strings.

### Shared utilities and repair scaffolding

The audit showed the same kind of small duplication in utility code that I saw in the test suite. I pulled `sanitize_path()` into `polylogue/lib/security.py` instead of maintaining similar implementations in more than one place. Extracted `_run_repair()` in `health.py`, which cuts down the repeated scaffolding around the repair operations.

Migration code changed shape for a similar reason. I replaced `executescript()` calls with explicit `execute()` calls in the migration path so failures are easier to localize and rollback behavior is more predictable. That matters more than it first appears, because migration failures are exactly the kind of rare edge case where "which statement failed?" is the first question a maintainer needs answered.

### Test consolidation and helper motion

The second half of the branch is largely about making the audit coverage sustainable.

I moved test-only helpers such as `store_records()` and related builders out of production code and into `tests.helpers`, which is where they belong. Consolidated a large number of tiny `_coverage` or near-duplicate test files into denser parent modules and replaced many repetitive test bodies with parametrized tables.

The important point here is not cosmetic line-count reduction. The important point is that the new correctness rules live in a suite that is easier to search, easier to extend, and less likely to split the same behavior across five files with subtly different fixtures.

## Testing

This branch is almost as much about test structure as production code. The changed behavior is covered in consolidated suites such as:

- `tests/test_storage.py`
- `tests/test_search_index.py`
- `tests/test_provider_models.py`
- `tests/test_models_messages.py`
- `tests/test_repository_operations.py`
- `tests/test_drive_coverage.py`
- `tests/test_cli_commands.py`
- `tests/test_click_app_integration.py`
- `tests/test_audit_fixes.py`

Raised the coverage floor to 90% in `pyproject.toml`, but that is a consequence of the new coverage shape, not the point of the branch.

## Notes

There are a couple of deliberate behavior changes reviewers should keep in mind. `normalize_role()` is stricter now, so callers that relied on normalized roles preserving arbitrary provider strings should read the raw/provider metadata instead. And `--delete` is intentionally less permissive: the command now requires a real filter instead of trusting the caller not to ask for a global delete.

I did not try to make every change "small" here, because the smaller path had already failed: the bugs were scattered across query routing, storage cleanup, provider normalization, and utility duplication, and the tests around them were just as scattered. The better maintenance move was to tighten all of those seams in one coherent pass.


Sinity:
@coderabbitai full review

Sinity:

Sinity:
@codex review

PR #76, opened 2026-04-07
feat: replace manual MCP dispatch with FastMCP and richer metadata

Sinity:
## Summary

I replaced the hand-rolled MCP server with a FastMCP-based implementation, moved color/theme decisions into a shared module, and expanded the provider extraction layer so Polylogue keeps far more of the source metadata without flattening everything to text. That feature work also gave me an excuse to remove duplicated filter/storage plumbing, package compressed schemas for drift validation, wire schema validation into ingestion, add Pagefind-backed site search, and fix the vector-distance ordering bug that could return worse matches ahead of better ones. The branch looks broad because the user-facing parts depend on that cleanup: once the filter pipeline, row mapping, and theme tokens were shared, the MCP/site/parser changes stopped needing one-off implementations in four places. The most important outcome is not only "Polylogue talks FastMCP now"; it is that the archive becomes a much richer, more faithful substrate for downstream tools. MCP clients, static-site consumers, and provider-aware analysis code now see data that previously disappeared during normalization.

## Motivation

There were three separate ceilings I wanted to break through here.

The first was protocol and integration. The existing MCP server worked, but it was still a hand-rolled JSON-RPC surface with custom dispatch, custom error handling, and a partial imitation of what the official SDK already knows how to do. That creates maintenance work for no real benefit. If Polylogue is going to expose tools, resources, and prompts as an MCP server, it should do so through the supported FastMCP interface rather than preserving a homegrown protocol layer.

The second was metadata fidelity. Polylogue already did a good job preserving core message text across providers, but it still left a lot of useful source data on the floor: ChatGPT citations and tool execution context, Claude Code cwd/git/model/token usage, Gemini safety/grounding/code execution state, Codex session context, and typed Claude AI records. For an archive project, flattening that metadata away is not neutral. It narrows what later analysis can ask.

The third was surface consistency. Color choices, CSS tokens, row-mapping helpers, filter pipelines, site rendering decisions, and schema assets were all slightly duplicated. That duplication was making it harder to add richer MCP/site/parser behavior without copying more code into more places.

### FastMCP migration

The center of the branch is `polylogue/mcp/server.py`. I replaced the manual request-dispatch path with a FastMCP server built from decorators for tools, resources, and prompts. That gives Polylogue a standards-aligned construction path instead of another layer of custom JSON-RPC glue.

Conceptually, the old code looked like this:

```python
request = json.loads(line)
response = _handle_request(request, repo)
print(json.dumps(response))
```

The new shape is closer to:

```python
server = FastMCP("polylogue")

@server.tool()
def search(...): ...

@server.resource("polylogue://stats")
def stats_resource(...): ...
```

That shift matters for three reasons:
1. protocol compliance moves to the SDK
2. tool/resource/prompt registration becomes declarative instead of living in one big request router
3. the server surface is easier to test because the interesting behavior is in callable endpoints, not in a monolithic dispatcher

Removed the 1000-character message truncation from full conversation resources. Truncation made sense when the server surface was mostly a quick inspection aid, but it is the wrong default once MCP clients are consuming Polylogue as an archive backend.

### Richer provider extraction

A lot of the branch value is in the parser/provider layer.

For ChatGPT I keep model slug, citations, code-execution details, thinking content blocks, gizmo metadata, archived state, and end-turn markers. For Claude Code I extract cwd, git branch, version, model, and token-usage information per message, then aggregate working directories and models at the conversation level. Claude AI gets a typed parsing path through `ClaudeAIConversation.model_validate()` with fallback to the older untyped route when validation fails. Codex picks up session-level git context and system instructions. Gemini gets a typed extraction path that keeps finish reason, thinking budget, safety ratings, grounding metadata, branch topology, executable code, execution results, error messages, and edit markers.

This is the part of the branch where "archive fidelity" becomes concrete. None of that data is decorative. It is exactly the context downstream analysis and tooling would want once the archive stops being just a text dump.

### Shared theme system

`polylogue/lib/theme.py` is new and intentionally central. Before this branch, provider colors and role colors were effectively scattered across CLI formatting, HTML output, static site templates, and TUI code. That made it much harder to keep the surfaces visually consistent.

The new module defines provider and role color dataclasses, shared theme dictionaries, and helpers for both Rich and CSS output. That lets the CLI and site speak the same visual language without duplicating the token definitions.

Used the same pass to improve actual rendering surfaces:
- Rich tables for CLI output
- role/provider-colored conversation display
- light-mode tokens for the static site
- `prefers-color-scheme` support
- Markdown rendering with code highlighting in conversation pages

The theme system is the glue that makes those changes coherent rather than ornamental.

### Filter, storage, and ranking cleanup

The feature work depended on simplifying some internals first.

In `polylogue/lib/filters.py` I genericized the dual conversation/summary paths so they share common filter application, fetch logic, sort logic, and pipeline execution. That removes a class of drift where one path would pick up a bug fix or a behavior change and the other would not.

In the SQLite backends I pulled common row mappers and vec-table DDL into shared helpers/constants, then reused them from both sync and async code. Corrected the vector ranking behavior. Similarity was effectively being ordered in the wrong direction by taking the wrong aggregate and reversing the sort. The new path treats lower distance as more similar and ranks accordingly.

There is also a small but important timestamp fix here: values below `86400` are no longer treated as epoch seconds automatically. That avoids misreading year-like values such as `"2025"` as timestamps.

### Static site and schema validation

The site builder picks up three meaningful improvements:
- incremental generation based on mtime
- Pagefind search UI wired into the index
- Markdown/code rendering that is closer to the CLI/HTML output

Delegated source iteration through `iter_source_conversations_with_raw(capture_raw=False)` so the two iteration modes stop diverging.

On the validation side, provider schemas are compressed `*.schema.json.gz` assets that ship with the package, and `SchemaValidator` is wired into ingestion. That moves schema drift checking from a "developer convenience" status into a real packaged capability.

## Testing

The branch updates and expands coverage in the places the behavior changed:

- `tests/test_mcp_server.py` for the FastMCP-based surface
- parser/provider tests for the richer metadata extraction
- sqlite-vec and storage tests for ranking/table-management behavior
- schema validation coverage around packaged compressed schemas
- site/UI tests around the theme and rendering paths

The important thing is that the tests now reflect the richer contract. For example, they do not just ask whether a provider message has text; they ask whether the source metadata that makes the message analyzable survives extraction.

## Notes

There are some real API and packaging changes here. The hand-rolled MCP helpers are no longer the primary construction path, some dead model helpers were removed, and schema assets ship compressed. I have documented those as API changes rather than pretending this is a drop-in internal refactor.

Did not try to make the branch "small" by splitting the cleanup from the features. That would have obscured the dependency chain. The FastMCP migration, richer extraction, shared theming, packaged schemas, and site work all got easier and safer once the duplicated filter/storage/theme plumbing was gone. Keeping them together makes the architectural move much clearer to review.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 5 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #78, opened 2026-04-07
feat: add async pipeline, schema registry, and synthetic corpus

Sinity:
## Summary

I am turning the archive runtime, contract layer, and test corpus into one coherent surface: the pipeline has async acquisition/parsing/rendering/indexing services and an async runner, provider schemas are versioned and annotated strongly enough to drive synthetic wire-format generation, and the CLI/MCP/export layers share more of the same formatting and automation plumbing. The biggest user-visible additions are the `polylogue demo` command, a much broader MCP server, async facade parity, branch-aware HTML rendering, and CI that exercises lint, tests, typechecking, and Nix builds together. Split the README into focused docs pages so the landing page can stay product-facing while the command, API, configuration, data-model, and MCP references live under `docs/`. The net effect is that parser work, automation work, and test work move against explicit contracts, not a pile of hand-maintained fixtures and sync-only entrypoints.

## Motivation

This branch started with schema-registry/MCP/TUI/test work, but the commit arc made the deeper problem obvious pretty quickly. I first cleaned up the test tree, fixed skips, and renamed the ingestion vocabulary to parsing so the code and CLI were at least speaking the same language. Once that was in place, the real blocker surfaced: too much of the repo still assumed a synchronous runner, static path bindings, duplicated formatting code, and real-export fixtures that were expensive to review and still too narrow to cover the provider edge cases we actually care about.

I considered taking the smaller path and just wrapping the old runner in `asyncio.to_thread()` while continuing to patch the fixture suite. That would have left stage boundaries, run bookkeeping, and watch-mode side effects coupled to the CLI, and it would have kept the parser contract story implicit. Instead I leaned into a structural move: split the pipeline into async services, make schema artifacts richer and versioned, then use those same schema annotations to generate provider-shaped synthetic corpora. That pivot is what let the async migration, the demo/showcase work, and the test reorganization land as one coherent change instead of three separate partial fixes.

### 1. Pipeline and facade parity

The largest architectural change is the new async pipeline path in `polylogue/pipeline/async_runner.py` and `polylogue/pipeline/services/async_*.py`. The old sync runner bundled source selection, parsing, rendering, indexing, drift tracking, and run recording together. The new layout makes those concerns explicit:

```python
result = asyncio.run(async_run_sources(
    config=config,
    stage="all",
    plan=None,
    ui=None,
    source_names=None,
))
```

That split only works if the library surface can consume it directly, so I also extended `polylogue/async_facade.py` instead of leaving it as a thin read-only wrapper. `AsyncPolylogue` exposes `search()`, `parse_file()`, `parse_sources()`, `rebuild_index()`, `filter()`, `config`, and `archive_root`, and its `stats()` method returns the same `ArchiveStats` object model the sync facade uses. In `polylogue/facade.py` I renamed the sync parsing entrypoints from `ingest_*` to `parse_*` so the vocabulary matches the actual staged runtime.

The rename carries through the rest of the stack. `polylogue/pipeline/prepare.py` replaces `ingest.py`, `prepare_ingest()` becomes `prepare_records()`, and `IngestionService` becomes `ParsingService`. I wanted the code to stop mixing "ingest" as a generic umbrella term with "parse" as the actual stage that converts raw provider bytes into normalized records.

### 2. Evented watch mode and safer automation boundaries

Once the pipeline became more explicit, the CLI no longer needed to own every watch-mode side effect itself. `polylogue/pipeline/events.py`, `polylogue/pipeline/event_bus.py`, and `polylogue/pipeline/watch.py` move notifications, shell hooks, and webhooks behind typed handlers and a reusable watch loop. `polylogue/cli/commands/run.py` now composes those handlers without hardcoding side effects inline.

That paid off in two ways. First, the watch loop is reusable outside the CLI. Second, it gave me a clean place to harden the automation boundaries. The webhook handler validates scheme/host resolution to avoid obvious SSRF targets, and the exec handler rejects shell metacharacters before dispatching through `shell=False`.

```python
handler = CompositeSyncHandler([
    NotificationHandler(),
    ExecHandler(exec_cmd),
    WebhookHandler(webhook_url),
])
runner = WatchRunner(sync_fn=_sync_once, handler=handler, interval=60)
```

Used this pass to replace import-time path constants with lazy accessors in `polylogue/paths.py`. That change looks small, but it removes a lot of hidden coupling in tests and seeded workspaces because `archive_root()`, `render_root()`, and `db_path()` resolve environment overrides when called, not when imported.

### 3. Versioned schemas and synthetic provider corpora

The contract layer is the other major shift. `polylogue/schemas/registry.py` introduces `SchemaRegistry` and `SchemaDiff`, giving provider schemas stable version metadata and a place to compare or promote regenerated artifacts. `polylogue/schemas/schema_inference.py` now annotates inferred schemas with `x-polylogue-*` metadata such as enum-like values, frequency, ranges, detected formats, reference targets, and array-length hints. Added sanitization so those annotations keep structural signals without leaking conversation content into committed schema artifacts.

That contract metadata is what unlocks `polylogue/schemas/synthetic.py`. Instead of hand-writing or committing real exports, the repo can generate provider-valid JSON/JSONL corpora from the schemas themselves, including tree-shaped formats like ChatGPT and Claude Code:

```python
corpus = SyntheticCorpus.for_provider("chatgpt")
raw_items = corpus.generate(count=3, seed=42)
```

This was a deliberate pivot from the earlier fixture cleanup commits. I could have kept reorganizing the old samples forever, but the schema annotations were already telling me enough about value ranges, formats, and field frequency to generate better test data than the committed fixtures were providing. That is why `tests/integration/test_workflows.py` and the parser contract tests move to synthetic sources instead of guarded real-data skips.

Folded optional tree-sitter language detection into `polylogue/schemas/code_detection.py` and wired the extras through `pyproject.toml`. Regex detection still works as the fallback, but when the grammars are installed the code can disambiguate close language matches with actual parser error ratios.

### 4. Shared formatting, richer MCP surface, and HTML branch rendering

A lot of smaller diffs are really one consolidation move. `polylogue/lib/formatting.py` centralizes conversation export formatting, and `polylogue/rendering/renderers/html.py` exposes `render_conversation_html()` so the HTML surface and the CLI/MCP export paths are not maintaining parallel serializers. The HTML renderer also gains `_attach_branches()` and matching template support in `polylogue/rendering/templates/conversation.html`, which means branch siblings render as grouped alternatives instead of disappearing into a flat mainline transcript.

On the automation side, `polylogue/mcp/server.py` becomes a much fuller product surface. I kept the basic read tools, but I added metadata/tag mutation, summary/session-tree/stat-by reads, health and index maintenance hooks, export formatting, new resources, and extra prompts. `_safe_call()` and `_error_json()` give the server a consistent structured failure path without leaking tracebacks to clients.

```python
server._tool_manager._tools["export_conversation"].fn(
    id="claude:abc123",
    format="markdown",
)
```

The MCP work is also where the shared formatter mattered most. I did not want CLI output, HTML exports, and MCP exports to keep drifting on field selection and content rendering.

### 5. Demo/showcase surface, docs split, packaging, and cleanup

The user-facing capstone is `polylogue/cli/commands/demo.py`. `polylogue demo --corpus` writes raw provider-format samples for inspection, `--seed` builds a full demo archive with seeded environment variables, and `--showcase` exercises the CLI surface and records a cookbook/report. That command is the practical payoff for the schema-driven corpus work: onboarding, debugging, and demos no longer depend on private archives.

```bash
eval $(polylogue demo --seed --env-only)
polylogue dashboard
```

Around that, I moved the packaging and docs to match the active codebase. `pyproject.toml` switches to `hatchling`, adds project metadata and the optional tree-sitter extras, and the repo gains GitHub Actions for lint, tests, mypy, coverage, and Nix builds. `flake.nix` adds `hatchling` and `vhs`, while the obsolete NixOS modules and the stale PR template are removed. The README becomes the product overview and quick start, while detailed reference material moves into `docs/cli-reference.md`, `docs/library-api.md`, `docs/data-model.md`, `docs/configuration.md`, and `docs/mcp-integration.md`.

## Testing

I treated the test reorganization as part of the feature work, not as cleanup after the fact. The suite is grouped under `tests/unit`, `tests/integration`, and `tests/infra`, with benchmarks/fuzz/manual helpers separated cleanly. The highest-signal coverage for this change is in:

- `tests/unit/test_async_facade.py`, `tests/unit/pipeline/test_async_helpers.py`, and `tests/unit/pipeline/test_async_services.py` for async facade and staged pipeline behavior.
- `tests/unit/core/test_schema.py` for validation/inference annotations and synthetic round-trips across providers.
- `tests/unit/cli/test_demo.py` and `tests/integration/test_workflows.py` for demo seeding/corpus generation and provider end-to-end workflows on synthetic data.
- `tests/integration/test_mcp_mutations.py`, `tests/integration/test_mcp_exports.py`, and `tests/integration/test_mcp_resources.py` for the expanded MCP surface.
- `tests/unit/rendering/test_branch_rendering.py` and `tests/unit/security/test_events_security.py` for branch-aware HTML and automation hardening.

The CI workflow enforces lint, formatting, strict mypy, and a pytest matrix with `--cov-fail-under=90`, plus a separate Nix workflow.

## Notes

A few things are intentionally conservative in this pass. Tree-sitter support is optional and the regex detector still remains the baseline path. Drive/OAuth behavior is still mocked in test coverage; I did not try to force live credentials into CI. The HTML branch view is sibling-grouped rather than a full interactive tree explorer, because I wanted a readable export that reuses the existing transcript layout. And the old `ingest_*` / path-constant API names are not preserved as compatibility shims, because keeping both vocabularies alive would have undercut the point of the surface cleanup.


Sinity:
@codex review

PR #80, opened 2026-04-07
refactor: make Polylogue async-first across query and render

Sinity:
## Summary

I moved the remaining high-traffic archive surfaces onto the async storage path so the library facade, filter chain, CLI query flow, MCP server, and rendering layer all stop maintaining parallel sync implementations. The biggest architectural change is that `ConversationFilter` is async at the terminal edge, which lets the same query object drive CLI, MCP, and library code without creating a shadow sync repository just to execute filters. Made the canonical `Polylogue` facade async and reintroduced it as the main library export after the branch briefly used an `AsyncPolylogue` alias as a bridge. On the rendering side, I deleted the dead sync `RenderService`, pushed `OutputRenderer.render()` into the async contract, and aligned the pipeline services with that boundary. The result is much less storage-path drift, fewer connection storms in higher-level APIs, and a smaller surface area for future pipeline work.

## Motivation

Before this branch, the codebase had already started migrating to async SQLite, but the top-level surfaces were still split. `ConversationFilter` built queries against a sync repository, the CLI query path mixed async and sync access depending on which option was active, MCP still had handlers that assumed synchronous execution, and rendering still carried a dead sync service plus sync protocol signatures. That duplication had two costs: correctness drift and maintenance drag. A filter bug or ordering tweak had to be fixed twice, and docs/examples were already starting to disagree on whether `Polylogue` and filter terminals were synchronous.

I tried the smallest possible bridge first by exporting `AsyncPolylogue` as `Polylogue` from `polylogue.__init__`. That removed some call-site friction, but it still left `polylogue/facade.py` and the rendering stack in an awkward half-migrated state. The more durable move was to restore `polylogue/facade.py` as the canonical async facade, update the MCP and CLI query call sites to await their filter terminals directly, and then remove the shims that only existed to keep sync code alive.

### Query and filter execution

`polylogue/lib/filters.py` now treats repository callbacks as awaitables. The core fetch helpers (`_fetch_generic()`, `_fetch_candidates()`, `_fetch_summary_candidates()`) were changed to async, and every terminal method now awaits the repository without assuming an in-process sync call.

```python
results = await (
    ConversationFilter(repo)
    .provider("claude")
    .contains("error")
    .limit(10)
    .list()
)
```

That change let me simplify `polylogue/cli/query.py`. The command now enters `_async_execute_query()` immediately, constructs the filter chain on an async repository, awaits count/summary/full-list paths consistently, and keeps sync access only where existing streaming or mutation helpers still expect it. This also let me clean up the public negative-filter naming by renaming:

```python
.exclude_text("warning")
.exclude_provider("gemini")
.exclude_tag("internal")
```

instead of the older `no_*` methods.

### Library facade and storage boundary

The old sync facade was removed and then rebuilt as an async `Polylogue` in `polylogue/facade.py`. That reintroduced a single library entrypoint with async context management, async `close()`, async `stats()`, async `search()`, and async batch retrieval helpers. Switched batch fetches to repository or backend batch paths instead of per-conversation round trips, which matters a lot for `get_conversations()`, `list_conversations()`, and stats generation on larger archives.

A notable pivot in the branch was export wiring. Early on, `polylogue.__init__` temporarily aliased `Polylogue = AsyncPolylogue` so the rest of the migration could land incrementally. Once `polylogue/facade.py` was restored as the canonical async facade, I removed that bridge and exported the real `Polylogue` class directly again.

### MCP and rendering

`polylogue/mcp/server.py` now awaits repository access in tools/resources/prompts rather than wrapping sync handlers around async internals. Search and list endpoints build the same async filter chains as the CLI, and conversation/detail resources fetch through the async repository API. That removes one of the last places where query behavior could drift from the CLI.

On the rendering side, I deleted `polylogue/pipeline/services/rendering.py`'s dead sync service, changed `OutputRenderer.render()` in `polylogue/protocols.py` to an async method, and completed the async migration in `polylogue/rendering/core.py` plus the Markdown/HTML renderers. This made the pipeline rendering stage a simple awaitable service instead of a sync wrapper around async database work.

### Cleanup that fell out of the migration

A lot of small but important cleanup came along with the boundary change: shared `DB_TIMEOUT` in `polylogue/storage/backends/connection.py`, renamed async index helpers in `polylogue/storage/async_index.py`, batch-oriented facade and analytics fixes to stop connection churn, progress output for render/index stages, provider-enum cleanup, and a long pass over docs/examples/tests so they describe the async contract the code now actually implements.

## Testing

I updated the integration coverage around the surfaces most affected by the migration:

- `tests/integration/test_facade.py` now exercises the async `Polylogue` API end to end, including parse, list, stats, and context-manager behavior.
- `tests/integration/test_mcp.py`, `test_mcp_exports.py`, and `test_mcp_resources.py` cover async MCP tool/resource/prompt handlers.
- `tests/unit/core/test_filters.py` and `test_filters_adv.py` were converted to await terminal filter methods and expanded around delete, summaries, and picker paths.
- `tests/unit/cli/test_query_exec.py` covers the async query execution path and count/stream behavior.
- `tests/unit/test_protocol_conformance.py` plus the rendering tests verify the new async renderer contract.

## Notes

This is intentionally a breaking library change. Filter terminal methods are async, the high-level `Polylogue` facade is async, and the negative filter methods use the `exclude_*` naming. I left the branch’s follow-on cleanup commits in place rather than squeezing them into a fake “big bang” diff because they capture the actual migration sequence: move the contracts first, then remove the shims, then clean up the fallout where connection patterns, docs, and tests still assumed the old world.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 3 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 1 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #82, opened 2026-04-07
perf: add incremental ingest tracking and pipeline SQL fast paths

Sinity:
## Summary

I used this release-packaging branch to land the performance and conformance work that makes the 0.1.x/0.2.x line behave well on real archives instead of just passing on small fixtures. The core theme is avoiding unnecessary work: unchanged files are skipped by mtime and parse tracking, parsing moves lightweight dedup checks out of the write lock, render/export paths stream not materializing the whole archive, and stats/count operations use SQL aggregation without loading full conversations. Advanced the schema/storage layer to support those fast paths with `sort_key`, parse-state tracking, and the `subagent` branch migration. On top of the hot-path work, I tightened release packaging with the missing `pyyaml` dependency and substantially expanded protocol/adversarial test coverage. The result is less peak memory, faster reruns, and a branch that is much safer to tag.

## Motivation

The release branch had two separate problems.

The first was operational. On a medium or large archive, rerunning the pipeline still did too much work: acquisition reread files it had already seen, parsing spent too long inside write-locked sections, render/export code materialized large intermediate structures, and simple user requests like `--count` or `--stats` could pull far more data than they needed. Those issues are survivable in development, but they become release blockers when the advertised workflow is “drop exports in a folder and rerun safely.”

The second was packaging drift. Some surfaces already documented or implied behavior that the packaged build did not fully support or validate. The YAML formatter needed its dependency declared, the eager-vs-lazy message model still carried code paths that bypassed the rest of the architecture, and the release line needed stronger protocol conformance tests so regressions would be caught before tags were cut.

I deliberately kept the branch ordered the way the work happened: early cleanup and conformance fixes first, then a focused performance campaign across ingest/render/export/storage, then schema/migration fixes and test hardening once the new fast paths were in place.

### Incremental ingest and lower parse contention

`polylogue/pipeline/services/parsing.py`, `polylogue/pipeline/services/acquisition.py`, `polylogue/pipeline/prepare.py`, and `polylogue/storage/backends/async_sqlite.py` carry most of the release-critical work.

The branch adds parse tracking columns on `raw_conversations`, exposes helpers such as `mark_raw_parsed()` and `get_known_source_mtimes()`, and teaches acquisition to skip unchanged files based on source-path mtime rather than re-reading everything. Parsing now treats “what changed?” as a cheap question.

Moved the light dedup/content-hash check outside the write lock and added `PrepareCache` for bulk lookups. That means parse batches can preload existing conversation/message state once and then do minimal locked work while writing.

```python
cache = await PrepareCache.load(backend, candidate_cids)
...
if existing and existing.content_hash == content_hash:
    return cid, skipped_counts, False
```

That one shift matters a lot under load because it prevents the slow “compute + compare + maybe write” path from serializing the whole parse stage.

### Render, export, and embed hot paths

The second group of changes is about peak memory and repeated CPU work.

`polylogue/export.py` no longer builds giant `defaultdict` maps for every message and attachment before writing JSONL. It streams one conversation at a time. `polylogue/cli/commands/embed.py` reuses a single event loop for batch embedding and resolves a conversation through `view()` so partial IDs and fetched messages stay in sync. `polylogue/rendering/renderers/html.py` caches templates and lexers, and the rendering services/pipeline runner scale workers based on the actual workload.

The cumulative effect is that the expensive archive-wide surfaces now behave more like bounded streaming jobs than “load everything, then serialize.”

### Storage, ordering, and SQL aggregation

The storage layer changes are what make the user-visible fast paths feel predictable.

I added numeric `sort_key` columns to conversations and messages, updated ordering queries to use them, and advanced migrations so existing databases can adopt the new indexes without manual intervention. That lets ordered reads avoid repeated timestamp parsing and helps both renderers and search-style listing paths.

Added SQL-backed aggregate stats and early stats/count exits. In practice that means query mode can answer things like `--count` and `--stats` from lightweight grouped queries rather than loading complete conversation bodies just to total message counts.

### Release hardening

This branch also removes the lazy `MessageCollection` / `MessageSource` detour, drops `Conversation.from_lazy()`, expands protocol conformance coverage, and declares `pyyaml` in both `pyproject.toml` and `flake.nix`. I treated those as release work because they close the gap between the formatter/protocol surface we expose and the package we actually ship.

There are two tiny README-only commits in the branch as well. I left them intact because they are genuine release polish: trimming the license line and removing heavy hero images that were not helping the docs story.

## Testing

The release hardening is covered from several angles:

- `tests/unit/test_protocol_conformance.py` exercises SearchProvider, VectorProvider, OutputRenderer, ConsoleLike, and parser module contracts.
- `tests/unit/pipeline/test_async_services.py`, `tests/unit/pipeline/test_services.py`, and `tests/unit/pipeline/test_prepare.py` cover the new incremental/streaming service behavior.
- `tests/unit/storage/test_sqlite_ops.py` and schema tests cover parse tracking, `sort_key`, and migration behavior.
- `tests/unit/cli/test_query_exec.py` verifies SQL-backed count/stats behavior.
- `tests/integration/test_mcp.py`, `test_mcp_resources.py`, `test_workflows.py`, and the generalized contract modules keep the optimized paths exercised from end to end.
- Adversarial suites were expanded to catch edge cases around provider metadata, security-sensitive inputs, and archive mutations.

## Notes

There are a couple of intentional tradeoffs here.

I did not try to turn every optimization into a generic abstraction. In several places the right answer was just to move work from “N per conversation” to “one batch query” and keep the code obvious. Kept the schema work tightly scoped to what the fast paths needed for release packaging: parse tracking, `sort_key`, and the `subagent` branch migration. That keeps the branch focused on “make the packaged release fast and dependable” instead of smuggling in unrelated storage redesign.


Sinity:
@coderabbitai full review

Sinity:
@codex review

PR #84, opened 2026-04-07
feat: add schema-validated ingest and synthetic showcase fixtures

Sinity:
## Summary

I pulled schema validation, provider normalization, synthetic corpus generation, and operator QA flows onto the same ingest spine without letting them keep evolving as parallel side paths. The biggest structural change is that ingestion is no longer just acquire → parse: `ParsingService` now orchestrates acquire → validate → parse, persists validation state on `raw_conversations`, and can expose validation without parsing through a dedicated `validate` stage. Alongside that, I extracted shared raw-payload decoding and provider inference so runtime parsing, schema verification, and showcase tooling all answer “what is this payload?” in the same way. Added two operator-facing surfaces on top of that work: `polylogue check --schemas` for raw-corpus verification and `polylogue qa` for snapshotting QA output directories with hashes and indexes. Finally, `polylogue demo --showcase` can seed itself from schema-driven synthetic fixtures, which makes the showcase path much more useful for exercising provider formats without relying only on packaged static files.

## Motivation

This branch started from the “synthetic corpus/showcase” feature request, but it quickly exposed a deeper architectural problem: the code that parsed real data, the code that verified schemas, and the code that generated demo/showcase data did not actually share one canonical payload path.

Runtime parsing had its own decode/provider-detection flow. Schema tooling had a different view of provider names and JSONL handling. Showcase/demo data could exercise the CLI, but it did not necessarily prove that schema verification and parser dispatch were using the same rules. That is exactly the kind of split that produces false confidence: tests pass, the showcase looks good, but the raw-corpus validation path and the runtime ingest path disagree on edge cases.

The fix was not to bolt synthetic generation on top. It was to give the pipeline an explicit validation phase, unify payload decoding/provider identity, and make the operator tools reuse that same machinery.

### Canonical payload decoding and provider identity

I added `polylogue/lib/raw_payload.py` with `decode_raw_payload()` and `infer_payload_provider()`. That gave both runtime ingest and schema verification one shared place to handle JSON vs JSONL, malformed JSONL-line counting, and path-aware provider inference.

I paired that with `polylogue/lib/provider_identity.py`, which distinguishes runtime provider names from schema provider names while still canonicalizing aliases. That matters because we have legitimate cases like runtime `"claude"` mapping to schema `"claude-ai"`.

```python
provider = infer_payload_provider(
    payload,
    source_path=raw_record.source_path,
    fallback_provider=raw_record.provider_name,
)
```

Using the same helper in both validation and parse flows means the pipeline no longer validates one provider identity and parses another.

### Acquire → validate → parse orchestration

`polylogue/pipeline/services/parsing.py` is the core of the branch. It exposes `ingest_sources()` and returns an `IngestResult` that carries acquisition, validation, and parse outputs explicitly. Validation lives in the new `polylogue/pipeline/services/validation.py`, which persists `validated_at`, `validation_status`, `validation_error`, drift counts, provider, and mode on raw records. The runner was updated so `run_sources()` understands `validate` as a real stage and reports validation counts alongside parse counts.

Added `polylogue/pipeline/services/ingest_state.py` to make the phase transitions explicit and guarded. That gives the branch a typed state machine for “what raw IDs are allowed to move from acquired to validation candidates to parse candidates,” which is useful both for preview/planning work and for catching bad orchestration regressions in tests.

### Schema verification and operator tooling

The new `polylogue/schemas/verification.py` adds a non-mutating verification pass over the full `raw_conversations` corpus. `polylogue/cli/commands/check.py` exposes that as `polylogue check --schemas`, with provider scoping, per-record sample depth, record windowing, and optional quarantine of malformed payloads.

That turns schema verification into an operator workflow instead of a behind-the-scenes utility. It also means schema validation is no longer “advisory log spam” during parse; it has a place to be run deliberately and reported clearly.

### Synthetic showcase data and QA snapshots

`polylogue/showcase/runner.py` and `polylogue/cli/commands/demo.py` support synthetic showcase seeding. If `--showcase-data synthetic` is selected, the runner generates provider fixtures through `SyntheticCorpus` instead of copying only packaged static files. That is a much better fit for a branch trying to prove that our schema/runtime pathways are actually unified.

Added a dedicated `polylogue qa` command that snapshots QA directories, computes SHA-256 hashes, writes an `INDEX.md`, and maintains a `latest` pointer. That gives us a lightweight artifact-management surface for release and support workflows.

## Testing

This branch adds coverage where the architectural seams moved:

- `tests/unit/pipeline/test_services.py` covers validation candidate selection and parse-ready ID selection.
- `tests/unit/pipeline/test_ingest_state.py` locks the ingest state-machine transitions and guards.
- `tests/unit/core/test_provider_identity.py` verifies runtime/schema provider canonicalization.
- `tests/unit/cli/test_check.py`, `test_demo.py`, and `test_qa.py` cover the new operator commands and showcase options.
- `tests/unit/core/test_filters.py` and `test_models.py` add no-raw harmonization regression coverage so message classification still works when only extracted provider metadata is present.

## Notes

A deliberate constraint here is that synthetic/showcase work is downstream of the ingest refactor, not the other way around. I did not want a separate “demo-only” provider-detection path. The synthetic corpus only becomes valuable if it is exercising the same decode, provider-identity, validation, and parse rules that real data will hit in production.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 3 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 8 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #86, opened 2026-04-07
refactor: unify run planning, search ranking, and MCP payloads

Sinity:
## Summary

I pulled several parallel control planes back into a single typed architecture: run preview and runtime execution share one planning service, notification hooks moved onto explicit run observers, search ranking is built from one canonical FTS query spec, and MCP stopped hand-assembling dictionaries in favor of typed payload models. The branch also tightens a bunch of boundaries that were still stringly typed or singleton-driven, especially around provider names, branch types, and query intent. In practice, that means the CLI, repository layer, facade, and MCP server are asking the same questions with the same objects without rebuilding slightly different versions of the same logic. Used the branch to delete dead event-bus code, fix one partial-write failure mode in the pipeline, and replace a large block of example-based JSON tests with property laws that better fit the new typed contracts. The result is a cleaner async pipeline foundation for later phases without adding another abstraction stack on top of the old one.

## Motivation

This branch came out of the “phases 0–8” async pipeline work, but the real problem was not “we need more phases.” It was that several layers were still carrying duplicate orchestration and serialization logic.

`run --preview` and real execution were not driven by one canonical planner. Notification and post-run hooks were split across legacy event modules. CLI and MCP both knew how to construct conversation filters, but they did it independently. Search ranking existed in multiple forms, which made deterministic ordering and tie-breaking harder to reason about. MCP payloads were assembled ad hoc, and a surprising number of values—provider names, roles, branch types, query params—were still crossing module boundaries as unvalidated strings.

That kind of duplication is expensive in a phase-based async system. Every new rule around source scoping, parse readiness, ranking, or serialization has multiple places it can drift. I wanted to reduce those seams before building more behavior on top.

### Canonical planning and run lifecycle

The new `polylogue/pipeline/services/planning.py` introduces `PlanningService` and `IngestPlan`. Preview and runtime execution can derive their work from the same source-scoped database state instead of hand-building separate lists of raw IDs or conversation IDs.

At the same time, I replaced the old notification/event hook model with `polylogue/pipeline/observers.py`. We have a small observer interface:

```python
class RunObserver:
    def on_progress(self, amount: int, desc: str | None = None) -> None: ...
    def on_completed(self, result: RunResult) -> None: ...
    def on_idle(self, result: RunResult) -> None: ...
    def on_error(self, exc: Exception) -> None: ...
```

That let me remove the dead `pipeline/event_bus.py` and later `pipeline/events.py` path, while still supporting desktop notifications, webhooks, and exec hooks on completed runs.

### Search ranking and query intent

`polylogue/storage/search.py` owns the canonical ranked conversation search query through `build_ranked_conversation_search_query()`. The backend and repository layers reuse that builder instead of each reinterpreting FTS ranking. The key behavior is one best hit per conversation, ordered deterministically by relevance and recency.

Introduced `polylogue/lib/query_spec.py` with `ConversationQuerySpec`. The CLI and MCP now construct the same typed selection intent and hand it to the filter layer instead of manually reassembling provider/tag/title/since conditions in multiple places.

```python
spec = ConversationQuerySpec(
    query_terms=("error",),
    providers=(Provider.CLAUDE,),
    since="2026-03-01",
    limit=25,
)
results = await spec.build_filter(repo).list()
```

That simplifies maintenance because there is one place to evolve query semantics.

### Typed MCP payloads and enum boundaries

`polylogue/mcp/payloads.py` replaces hand-assembled dict payloads with typed Pydantic models for summary/detail conversations, archive stats, mutation status, metadata, tag counts, stats-by payloads, and errors. The server helpers just serialize those models, which makes it much easier to review the wire contract and keep tools/resources/prompts consistent.

This also pairs with the typed-boundary cleanup elsewhere in the branch: `BranchType` moved into its own enum, provider-related inputs now flow through the typed `Provider` surface, role helpers stringify through `Role`, and storage/query/formatting code was updated to work with those types directly instead of assuming raw strings.

### Failure handling and contract tests

The branch includes two correctness hardening steps that mattered once the boundaries were tightened.

First, I fixed a pipeline failure mode where attachment materialization could partially mutate state before a save failed. That path now rolls back correctly. Second, I stopped swallowing harmonized/indexing exceptions quietly and log them instead, which should make future edge cases much easier to diagnose.

Finally, I replaced a large pile of example-based JSON serialization tests with property laws in `tests/unit/core/test_json_laws.py`. For typed payloads and enums, that is a much better long-term fit than adding more fixtures every time a model changes.

## Testing

The branch’s regression surface is deliberately broad:

- `tests/unit/pipeline/test_services.py` and `test_pipeline.py` cover source scoping, planning, and rollback behavior.
- `tests/unit/storage/test_fts5.py` and `test_repository.py` verify ranked-search semantics and deterministic best-hit ordering.
- `tests/integration/test_mcp.py` and `test_facade.py` cover the shared query/payload behavior through real higher-level surfaces.
- CLI tests were updated around the run command, subprocess harness, demo command, and auth/dashboard/query behavior.
- `tests/unit/core/test_json.py` and `test_json_laws.py` replace example duplication with property-based serialization contracts.

## Notes

A couple of things are intentionally modest here. I did not try to invent a generic event framework after deleting the old bus; the observer layer is deliberately small and run-result focused. Kept the typed-boundary cleanup close to the code actually using it, rather than introducing a giant “types” rewrite in one commit. The point of the branch is not more abstraction for its own sake. It is fewer ways for preview/runtime/query/MCP behavior to drift while the async pipeline keeps evolving.


Sinity:
@codex review

PR #88, opened 2026-04-07
refactor: make content blocks first-class and precompute analytics

Sinity:
## Summary

I reset the storage line around structured content blocks and precomputed
analytics so the archive stops rediscovering tool/thinking state by scanning
provider blobs at read time. The big architectural move here is making
`content_blocks` the canonical stored representation of rich message structure,
then hanging fast analytics and query predicates off persisted counts instead of
hydrated conversations. In the same branch I tightened schema-inference privacy
because the new storage shape makes schema generation a more prominent workflow,
and I replaced a lot of broad example-heavy tests with property laws and
snapshots so the new invariants are explicit. I considered layering this on top
of the existing migration chain, but the schema shape changed enough that a
fresh reset was simpler and less risky than carrying forward half-legacy,
half-block-native rows. The result is a branch that is half storage rewrite and
half test/QA realignment around that rewrite.

## Motivation

Before this change, several important paths were paying an avoidable tax:

- message structure lived mostly as provider-shaped blobs, so detecting things
 like tool use or thinking blocks often meant scanning nested JSON instead of
 answering from first-class rows;
- analytics and query helpers had to decide between loading whole conversations
 or approximating answers from incomplete metadata;
- schema inference could still promote values that looked structural in a small
 sample but were really sensitive or one-off content;
- a lot of existing tests were broad examples that exercised code paths without
 pinning the invariants I actually cared about for a schema reset.

The storage rewrite and the QA rewrite belong together because the storage work
changes what “correct” means. It is much easier to trust a reset like this when
the suite checks laws such as idempotence, mutual exclusion, and semantic
contracts, not a stack of brittle string matches.

### 1. First-class content blocks and analytics rows

`polylogue/storage/backends/schema.py` now defines a clean-schema v1/v2 world
where `content_blocks` are stored independently and `conversation_stats` carries
rollups that were previously recomputed or inferred. `polylogue/storage/store.py`
gains the row-model support needed to move those blocks through the pipeline.

At the repository/backend layer, the important shift is that message reads now
reconstruct structured blocks directly from storage rather than expecting to
rediscover them from `provider_meta`. The save path also grows a first-class
block persistence stage.

```python
await repo.save_conversation(
    conversation,
    messages,
    attachments,
    content_blocks=content_block_records,
)
```

That makes later consumers less provider-specific: they can ask “what blocks are
here?” instead of “which provider blob do I have, and how do I recurse it?”

### 2. Precomputed message and conversation analytics

I pushed analytics closer to ingest. `polylogue/pipeline/prepare.py` emits the
rows needed for block storage, and `polylogue/storage/backends/async_sqlite.py`
maintains precomputed message flags and `conversation_stats` rollups. The CLI
analytics surface then uses fast summary/SQL paths where possible and only falls
back to full message loading when the question really needs content.

This is also where the large-archive wins land. `query --latest` and `stats-by`
stop behaving like ad hoc report generators over hydrated conversations and
start behaving like indexed queries over durable counters.

### 3. SQL pushdowns instead of post-filter scans

`ConversationFilter`, `_build_conversation_filters()`, `ConversationQuerySpec`,
the Click surface, and the MCP server now agree on the same pushdown-capable
filter set:

```python
ConversationFilter(repo)     .has_tool_use()     .has_thinking()     .min_messages(4)     .min_words(200)
```

That same shape is exposed in CLI params and MCP query specs, which means there
is one place to reason about predicate semantics and one place to optimize them.

### 4. Privacy guards in schema inference

The schema-inference changes are not just “more denylist items.” I added three
defenses that change what values can graduate into enums:

- a cross-conversation threshold so one-off values do not look structural;
- a broader content-field denylist for things like `body`, `message`, `input`,
 and `output`;
- rejection of private/internal TLD-like values in addition to public domains.

The goal is to keep inferred schemas useful without turning them into a lossy
catalog of user content.

### 5. QA realignment

A large part of the branch deletes tests on purpose. Example tables and smoke
assertions were replaced with:

- property-law suites for messages, parsing, FTS5 escaping, and hybrid search;
- snapshot coverage for rendering output;
- tiered showcase exercises and report/audit-trail checks;
- privacy-specific schema inference tests.

That gives me stronger regression signal for a storage reset than keeping the
older, broader examples around.

## Testing

I exercised the storage and analytics work through a mix of focused and
cross-cutting tests:

- `tests/unit/pipeline/test_prepare.py` and related helpers for ingest record
 preparation;
- `tests/unit/storage/test_backend.py`, `test_sqlite_ops.py`, `test_search_misc.py`,
 and `test_render.py` for the storage read/write consequences of the new row
 model;
- `tests/unit/cli/test_analytics.py` and `tests/unit/cli/test_query_exec.py` for
 analytics/query behavior on the new fast paths;
- `tests/unit/core/test_schema_privacy.py` for the privacy guards;
- `tests/unit/core/test_message_laws.py`,
 `tests/unit/storage/test_fts5_laws.py`,
 `tests/unit/storage/test_hybrid_laws.py`,
 `tests/unit/sources/test_parse_laws.py`, and
 `tests/unit/sources/test_harmonization_contracts.py` for the new law-based
 coverage;
- `tests/unit/rendering/test_output_snapshots.py` for renderer regressions.

## Notes

- This is intentionally a fresh-schema change. There is no migration path from
 the old line to v2.
- The showcase, snapshot, and mutation-tooling changes are in this branch
 because I used them to validate the new invariants before and during the
 storage reset, not because I wanted to bundle unrelated product work.
- The main thing I want reviewed closely is not raw line count, but whether the
 new canonical authority is clear: `content_blocks` and precomputed analytics
 should now be the place higher-level surfaces read from.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 14 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 10 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #90, opened 2026-04-07
feat: classify content blocks with semantic types at ingest

Sinity:
## Summary

This branch adds a semantic layer to stored content blocks without treating
every block as just raw `type` plus text/tool fields. `content_blocks` now carry
a `semantic_type`, the storage layer indexes it, and `prepare_records()` classifies
blocks as they are ingested so later consumers can answer higher-level questions
without reparsing provider payloads. Pulled that logic into a dedicated
`polylogue/pipeline/semantic.py` module rather than letting `prepare.py` grow a
pile of provider-specific conditionals. The storage change is a clean schema v3
step with no migration path, because the value of the branch is in making the
semantic column authoritative from the start. Although the branch metadata points
at documentation work, the actual diff is a very tight two-step storage and
ingest feature.

## Motivation

After the content-block rewrite, Polylogue could tell that a block was
`tool_use`, `thinking`, or `code`, but it still could not answer the more useful
questions that product surfaces actually want to ask:

- was this tool invocation a file read, file write, git command, or subagent
 spawn?
- if a code block was stored without metadata, what language should downstream
 renderers and analyzers assume?
- can we make those distinctions once at ingest instead of teaching every query,
 renderer, and analysis path to rediscover them?

The answer here is “yes, if storage has a durable semantic column and ingest is
responsible for filling it.”

### 1. Schema v3 adds a semantic axis to content blocks

`polylogue/storage/backends/schema.py` now adds `semantic_type` to
`content_blocks` and indexes it both globally and per conversation. This is the
minimal storage primitive needed for semantic predicates such as file ops, git
ops, and subagent activity.

`polylogue/storage/store.py` extends `ContentBlockRecord` accordingly, and the
backend save path in `polylogue/storage/backends/async_sqlite.py` persists and
updates the field on conflict.

```python
ContentBlockRecord(
    ...,
    type="tool_use",
    semantic_type="git",
)
```

Added `_topo_sort_messages()` to the async SQLite backend so message
batches still insert parents before children when parent-message foreign keys are
present. That is adjacent to the semantic work, but it is the kind of storage
hygiene I wanted in place before leaning harder on block-level persistence.

### 2. Ingest-time classification instead of read-time rediscovery

The new `polylogue/pipeline/semantic.py` module is the canonical home for
semantic extraction. It classifies tools with `classify_tool()`, parses git
commands into structured metadata, extracts file/subagent details, keeps the
older raw-dict helper APIs available where tests/internal callers still need
them, and gives `prepare_records()` a single semantic dependency.

`polylogue/pipeline/prepare.py` then writes the results once:

```python
if block.type == "tool_use" and block.tool_name:
    category = classify_tool(block.tool_name, block.tool_input or {})
    semantic_type = category.value
    tool_meta = extract_tool_metadata(block.tool_name, block.tool_input or {})
```

Thinking blocks get a durable `"thinking"` semantic tag, and code blocks without
metadata pick up detected language information at ingest time.

### 3. Test surface

I added explicit coverage for both layers:

- `tests/unit/storage/test_crud.py` proves the new column persists and
 round-trips cleanly;
- `tests/unit/pipeline/test_prepare_semantic.py` exercises tool classification,
 git parsing, file/subagent metadata extraction, thinking tagging, and code
 language detection.

## Testing

The branch is small enough that the tests are correspondingly focused:

- `tests/unit/storage/test_crud.py`
- `tests/unit/pipeline/test_prepare_semantic.py`

Together they cover the schema/persistence path and the ingest classification
path. I did not wire the new semantic fields into the full query surfaces here;
that follow-up belongs in the next branch so this one stays focused on the data
model and ingest authority.

## Notes

- This is a fresh-schema v3 change; there is no migration path from the earlier
 schema line.
- The important architectural decision is not the column itself. It is that
 `prepare.py` becomes the place semantic meaning is attached, which lets later
 filters and analytics behave like storage queries instead of reparsers.

### DIFF
--- a/polylogue/pipeline/prepare.py
+++ b/polylogue/pipeline/prepare.py
@@ -8,6 +8,9 @@ from typing import TYPE_CHECKING
 
 from polylogue.lib.log import get_logger
 from polylogue.lib.json import dumps as json_dumps
+from polylogue.lib.viewports import ToolCategory, classify_tool
+from polylogue.pipeline.semantic import extract_tool_metadata
+from polylogue.schemas.code_detection import detect_language
 from polylogue.pipeline.ids import (
     attachment_content_id,
     conversation_content_hash,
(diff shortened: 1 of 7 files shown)


Sinity:

Sinity:
@codex review

PR #92, opened 2026-04-07
perf: expose indexed semantic filters and benchmark hot paths

Sinity:
## Summary

This branch finishes the semantic-type work by making semantic filters real
product surfaces instead of storage-only potential. File-op, git, and subagent
predicates now flow from `ConversationFilter` through query specs, repository
list/count methods, the Click CLI, and the MCP server, and they execute as SQL
predicates over `content_blocks.semantic_type` rather than read-time enrichment.
At the same time I removed the last parser/model assumptions that raw provider
blobs would be available on parsed messages, because schema v3 made stored
content blocks the canonical source of message structure. Replaced the
old standalone benchmark scripts with `pytest-benchmark` coverage and explicit
latency budgets so I can measure these hot paths repeatably. A small but
important mid-branch fix corrected SQL scoping for semantic `EXISTS` predicates
when they run alongside stats joins; without that, the benchmark numbers would
have been measuring the wrong query.

## Motivation

After the previous branch, the database could store semantic block types, but
the rest of the stack still had three gaps:

1. semantic queries were not fully exposed through the filter/query surfaces;
2. some model and parser paths still assumed raw provider blobs were the place
 to recover semantics, even though ingest-time content blocks had become the
 canonical representation;
3. the performance story was still half manual because the benchmark scripts in
 `tests/benchmarks/` were not reliable enough to act as regression signal.

I wanted one branch that closed those gaps in order: make the semantics
authoritative, expose them, then measure them.

### 1. Remove read-time semantic fallback as the primary path

The early commits do cleanup that is important for both correctness and
performance:

- semantic extraction helpers move out of `polylogue/sources/parsers/claude.py`
 and into `polylogue/pipeline/semantic.py`;
- `polylogue/sources/parsers/base.py` and related parser tests stop assuming
 `ParsedMessage.provider_meta["raw"]` is the normal path for semantic recovery;
- `polylogue/lib/models.py` prefers stored `content_blocks` for
 `is_tool_use`, `is_thinking`, and `extract_thinking()`.

That means the read path can trust what ingest already decided instead of
replaying provider-specific logic.

### 2. Productize semantic filters

The core interface additions are simple:

```python
ConversationFilter(repo)     .has_file_operations()     .has_git_operations()     .has_subagent_spawns()
```

Those flags now travel through:

- `polylogue/lib/query_spec.py`
- `polylogue/storage/backends/connection.py`
- `polylogue/storage/repository.py`
- `polylogue/cli/click_app.py`
- `polylogue/mcp/server.py`

and end up as `EXISTS` predicates over `content_blocks.semantic_type`.

That is also why `polylogue/pipeline/enrichment.py` could be deleted. Semantic
enrichment is no longer a read-side patch-up step; it is part of ingest.

### 3. Fix the SQL shape before trusting measurements

When I first wired the semantic filters, there was a scoping problem around
combining semantic `EXISTS` filters with the stats-join path. The fix in
`polylogue/storage/backends/connection.py` makes the conversation ID reference
switch correctly depending on whether the query has already introduced the
`conversation_stats` join. I added focused tests for that path in
`tests/unit/core/test_filters_adv.py` and related UI/CLI tests because it is the
kind of bug that looks fine until you combine the predicates.

### 4. Replace dead benchmark scripts with a real harness

The old benchmark scripts were large, bespoke, and difficult to trust. I deleted
them, added `pytest-benchmark`, and rewrote the perf coverage as normal tests:

- `tests/benchmarks/test_pipeline.py`
- `tests/benchmarks/test_search_filters.py`
- `tests/benchmarks/test_storage.py`

Tightened the budget-style checks in `tests/unit/storage/test_scale.py`
so common hot paths have explicit expectations in CI-adjacent tests.

```python
@pytest.mark.benchmark
def test_bench_filter_count(...):
    ...
```

This gives me a split between microbench timing and “must stay under this
latency budget” assertions.

## Testing

The branch is covered by three main clusters:

- parser/model cleanup: `tests/unit/sources/test_claude.py`,
 `tests/unit/sources/test_parsers_base.py`,
 `tests/unit/sources/test_parsers_props.py`,
 `tests/unit/sources/test_extraction.py`;
- semantic filter behavior: `tests/unit/core/test_filters_adv.py`,
 `tests/unit/ui/test_tui.py`, `tests/unit/cli/test_run_int.py`,
 `tests/unit/storage/test_scale.py`;
- performance harness: `tests/benchmarks/test_pipeline.py`,
 `tests/benchmarks/test_search_filters.py`,
 `tests/benchmarks/test_storage.py`.

## Notes

- This branch does not change the database schema version again; it is the
 productization and perf follow-through for the semantic-type storage introduced
 earlier.
- The biggest reviewer question is whether every surface now agrees that stored
 `content_blocks` are the semantic authority. If any path still needs raw
 provider blobs to answer normal questions, that is a bug.
- The benchmark suite intentionally lives in pytest now. I wanted the same test
 runner, markers, and fixture system to own both correctness and performance
 signal.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 4 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550).

Sinity:
@codex review

PR #100, opened 2026-04-07
test: concentrate mutation owner contracts and split test infra

Sinity:
## Summary

I collapsed the mutation-sensitive source, parser, CLI query, and repository surfaces into explicit owner suites, refreshed the corresponding mutmut baselines, and then finished the pass by splitting `tests/infra/helpers.py` into focused infra modules. The resulting tree is much less ambiguous about where provider sniffing, harmonization fallbacks, Drive parser behavior, repository hydration rules, and query entrypoint seams are supposed to be asserted. To make those owner files viable, I also extracted a handful of small support seams in `polylogue/schemas/unified.py`, `polylogue/storage/repository.py`, `polylogue/cli/query.py`, and `polylogue/sources/parsers/drive.py`. The overall effect is not a new product feature; it is a large test-ownership cleanup that makes the mutation campaigns readable again and removes a lot of cross-domain leakage from the suite.

## Motivation

The branch history made the underlying problem pretty clear. The source-detection, providers-semantics, sources-parse, cli-query, and repository campaigns were all still getting signal from overlapping legacy files, so a mutation kill or survivor did not tell me which owner surface actually mattered. The same helper logic was being asserted from storage tests, parser tests, and source-iteration tests at once, while `tests/infra/helpers.py` had grown into a catch-all module for builders, assertions, and storage seeding. That shape made every follow-up rerun noisy: I could improve a campaign, but the suite layout still hid the real contract boundary.

I deliberately kept the branch in a repeated rhythm: tighten one owner surface, rerun the campaign, record the new baseline, and move on to the next cluster. That is why the diff alternates between test concentration commits and `docs/mutation-testing-baseline.md` updates. Once the source and repository contracts were concentrated enough, a few internal refactors became obvious because they gave the new owner files stable seams to target without routing through duplicated call paths.

### Source detection, provider semantics, and parser ownership

The biggest concentration work happens under `tests/unit/sources/`.

I moved provider detection, cursor-state bookkeeping, mixed ZIP/source iteration, grouped JSONL raw-capture behavior, and mtime skipping into `tests/unit/sources/test_source_laws.py`, and deleted the scattered `test_providers.py`, `test_raw_capture.py`, and `test_source_ops.py` surfaces that used to share those concerns. Collapsed provider viewport and fallback harmonization coverage into `tests/unit/sources/test_unified_semantic_laws.py`, which let me remove `tests/unit/sources/test_provider_viewport_laws.py` and trim duplicated assertions from `test_extraction.py`.

That concentration is supported by a small internal rewrite in `polylogue/schemas/unified.py`:

```python
_ADAPTER_BUILDERS = {
    Provider.CLAUDE_CODE: _validate_claude_code_record,
    Provider.CLAUDE: _validate_claude_ai_message,
    Provider.CHATGPT: _validate_chatgpt_message,
    Provider.GEMINI: _validate_gemini_message,
    Provider.CODEX: _validate_codex_record,
}
```

and the matching `_FALLBACK_EXTRACTORS` map. The behavior does not widen the public surface, but it makes the dispatch table explicit enough that the new law suite can assert adapter selection and fallback routing without duplicating provider-specific test scaffolding.

Narrowed `tests/unit/sources/test_parsers_drive.py` to concrete Drive/Gemini parser contracts and changed `polylogue/sources/parsers/drive.extract_text_from_chunk()` to accept arbitrary objects and return `None` for non-dict chunks. That removes a brittle `AttributeError` path and matches what the concentrated parser contract is actually asserting.

### Query and repository seam cleanup

The query campaign was carrying a lot of formatting and presentation noise that did not belong in the mutation owner file. I pulled vector-provider construction into `polylogue/cli/query._create_query_vector_provider()` and rewrote `tests/unit/cli/test_query.py` around the async execution seam: config failures, query-plan failures, vector-provider setup, and argument flow into the filter builder. The important behavior here is still `_async_execute_query()`, but the helper extraction makes that route much easier to test directly.

Repository ownership got a similar cleanup. In `polylogue/storage/repository.py` I extracted `_provider_conversation_id()` so canonical ids are stripped in one place, and `_resolve_optional_vector_provider()` so the similarity/search paths stop inlining default-provider creation. That pairs with the move from broad `test_repository.py` coverage into `tests/unit/storage/test_repository_laws.py`, where the suite owns context-manager cleanup, concurrent empty reads, attachment pruning, and bundle resave behavior directly.

### Shared test infrastructure split

The last structural change in the branch is the removal of `tests/infra/helpers.py`. By the end of the earlier concentration passes it had become obvious that the helper module itself was now the biggest source of cross-domain leakage.

I replaced it with three focused modules:

```python
from tests.infra.source_builders import make_claude_chat_message
from tests.infra.storage_records import ConversationBuilder, db_setup
from tests.infra.assertions import assert_messages_ordered
```

- `tests/infra/source_builders.py` owns provider export builders and source payload factories.
- `tests/infra/storage_records.py` owns storage-record builders, DB seeding helpers, and attachment ref utilities.
- `tests/infra/assertions.py` holds the small rendered-output assertion helpers.

The import churn across CLI, core, pipeline, storage, parser, integration, and UI tests is large, but the effect is simple: source builders are no longer pulled in through storage helpers, and storage helpers are no longer the home for generic text assertions.

### Campaign definitions and baseline refreshes

I updated `devtools/mutmut_campaign.py` as each owner surface moved so the campaign definitions point at the concentrated suites instead of the old umbrella files. After each meaningful shift, I refreshed `docs/mutation-testing-baseline.md` so the branch records the new kill/survivor shape for source-detection, providers-semantics, sources-parse, cli-query, and repository. That documentation is not decorative here; it is the evidence that the ownership changes are actually changing the mutation profile rather than just moving tests around.

## Testing

The main coverage additions are:

- `tests/unit/sources/test_source_laws.py` for provider sniffing, JSON stream fallback, cursor state, raw iteration, ZIP/grouped byte preservation, and source iteration failure handling.
- `tests/unit/sources/test_unified_semantic_laws.py` for adapter dispatch, fallback extraction, token/cost extraction, overlay behavior, and provider content-block semantics.
- `tests/unit/sources/test_parsers_drive.py` for chunk text extraction, metadata preservation, reasoning/tool-result assembly, attachments, and synthetic Gemini payload acceptance.
- `tests/unit/cli/test_query.py` for `execute_query()`, `_create_query_vector_provider()`, query-plan failures, and vector-provider routing into filter creation.
- `tests/unit/storage/test_repository_laws.py` for repository context handling, empty concurrent reads, bundle resave pruning, and search/similarity forwarding.
- `tests/infra` import migrations and focused infra helpers are exercised by the existing callers that now import from `source_builders.py`, `storage_records.py`, and `assertions.py`.

## Notes

The user-facing runtime changes are intentionally small. The two behavior changes I do expect are that Drive parser helpers now fail closed on non-dict chunk inputs, and query execution now centralizes best-effort vector-provider setup instead of repeating that logic inline. The mutation baselines also make it clear that there is still follow-up work left in `polylogue.schemas.unified`, source ZIP/provider sniffing paths, and repository hydration/archive-stat coverage, but this branch gets the ownership boundaries into a shape where those next passes should be much easier to reason about.


Sinity:
@coderabbitai full review

Sinity:

Sinity:
@codex review

PR #102, opened 2026-04-07
feat: add benchmark campaign tooling and phase owner suites

Sinity:
## Summary

I split the next mutation wave into explicit owner suites for storage/repository, schema/pipeline, CLI, and integration/MCP/security, then added `devtools/benchmark_campaign.py` so benchmark regressions can be recorded and compared as durable artifacts, not one-off `pytest-benchmark` runs. The net effect is that the branch no longer treats “testing work” as one undifferentiated surface: the mutmut campaigns now point at the files that actually own the behavior they mutate, and the benchmark domain has a first-class developer tool with `list`, `run`, `compare`, and `index` entrypoints. The one notable production-code simplification outside devtools is in `polylogue/pipeline/services/acquisition.py`, where I removed the duplicate scan/store side paths so the pipeline service surface matches the staged model the new owner suites are exercising. Followed up with two tooling fixes that became obvious once the new workflow existed: rooting mutmut artifacts under `ROOT` and insulating benchmark runs from repo-level `pytest` defaults.

## Motivation

The mutation ledger had reached the point where “more tests” was not the bottleneck. The problem was ownership. The campaigns were still pointing at wide umbrella files like `tests/unit/core/test_schema.py`, `tests/unit/pipeline/test_services.py`, `tests/integration/test_click.py`, and large integration/MCP/security suites that mixed too many concerns. That made the baseline tables harder to interpret, because a survivor or timeout cluster did not tell me which owner surface was weak; it only told me that some big domain file still needed attention.

Benchmarks had a similar problem. The benchmark suite existed, but the workflow around it was still mostly a set of manual pytest incantations copied into comments. That was good enough for a local spot-check, but not good enough for durable before/after comparisons across branches. I wanted the benchmark domain to look more like the mutation domain: named campaigns, deterministic artifact locations, machine/commit metadata, explicit regression thresholds, and a lightweight compare step.

The branch history reflects that progression. I started by tightening storage/repository ownership, then repeated the same phase pattern across schema/pipeline, CLI, and integration surfaces. Once those owner boundaries were in place, it became worthwhile to formalize the benchmark workflow. The two small follow-up fixes at the end are exactly the kind of issues that only show up once the workflow is exercised repeatedly: path anchoring for mutmut artifacts and `pytest` default isolation for benchmark runs.

### Storage and repository owner files

The first pass focuses on storage and repository behavior.

I deleted broad legacy surfaces such as `tests/unit/storage/test_async.py`, `test_crud.py`, `test_render.py`, and `test_sqlite_ops.py`, and moved the durable behavior into tighter owners:

- `tests/unit/storage/test_crud_laws.py`
- `tests/unit/storage/test_repository_laws.py`
- `tests/unit/storage/test_store_ops.py`
- the focused async/backend infrastructure contracts in `tests/unit/storage/test_backend.py`

That reorganization makes the repository campaign much easier to aim. Instead of mutating `polylogue/storage/repository.py` against a mixed grab-bag of tests, the campaign now points at the law suite that actually owns repository conversion, pruning, search forwarding, context cleanup, and save/resave behavior.

The only runtime-adjacent coverage I intentionally kept outside those files is the HTML escaping contract in `tests/unit/ui/test_rendering.py`. That behavior matters, but it belongs to renderer/UI ownership rather than the repository law suite.

### Phase 4: schema and pipeline owners

The next concentration pass is larger and introduces the main production code change in the branch.

In `polylogue/pipeline/services/acquisition.py` I removed `ScanResult.records`, `AcquisitionService.scan_sources()`, and `AcquisitionService.store_records()`. Those paths duplicated the staged acquisition flow that `visit_sources()` and the normal pipeline already expose. Keeping them around made the service surface broader than the owner tests actually needed.

The resulting API is simpler:

```python
result = await service.visit_sources(
    sources,
    progress_callback=progress_callback,
    on_record=handle_record,
)
```

instead of holding onto a second “scan now, persist later” path inside the service itself. That lines up much better with the owner files I added for the pipeline layer:

- `tests/unit/pipeline/test_acquisition_service.py`
- `tests/unit/pipeline/test_async_index.py`
- `tests/unit/pipeline/test_parsing_service.py`
- `tests/unit/pipeline/test_planning_service.py`
- `tests/unit/pipeline/test_prepare_records.py`
- `tests/unit/pipeline/test_render_service.py`
- `tests/unit/pipeline/test_run_sources.py`
- `tests/unit/pipeline/test_runner_history.py`
- `tests/unit/pipeline/test_runner_preview.py`
- `tests/unit/pipeline/test_validation_service.py`

I did the same thing on the schema side by splitting the old monolithic coverage into `tests/unit/core/test_schema_annotations.py`, `test_schema_generation.py`, and `test_schema_validation.py`. That separation matters because the mutation baseline is telling us that validation, annotation, and inference heuristics fail in different ways, and the suite should make those boundaries obvious.

Once the owner files existed, I updated `devtools/mutmut_campaign.py` so the phase 4 campaigns point at them directly, and refreshed the baseline ledger in `docs/mutation-testing-baseline.md`.

### Phase 5: CLI owners

The CLI pass follows the same pattern but with different surfaces.

I replaced the older umbrella CLI files with focused owners such as `tests/unit/cli/test_click_app.py`, `test_command_surfaces.py`, `test_formatting.py`, and `test_source_selection_helpers.py`, while moving true end-to-end subprocess behavior into `tests/integration/test_cli_query_mode.py`. Added `tests/infra/test_storage_records.py` and `tests/unit/core/test_lazy_exports.py` where that phase needed explicit ownership for shared storage test helpers and lazy-export surfaces.

The point here is not just deletion count. The new owner split means I can tell whether a regression belongs to CLI app wiring, command surface resolution, formatting, or subprocess query behavior without reading through one giant file first.

### Phase 6: integration, MCP, and security owners

For the integration-heavy surfaces, I added shared deterministic corpora and MCP helpers first:

- `tests/infra/adversarial_cases.py`
- `tests/infra/mcp.py`

Then I moved concrete owner behavior into dedicated unit-level files:

- `tests/unit/core/test_facade_api.py`
- `tests/unit/core/test_health_core.py`
- `tests/unit/mcp/test_server_surfaces.py`
- `tests/unit/mcp/test_tool_contracts.py`
- `tests/unit/security/test_attachment_security.py`
- `tests/unit/storage/test_query_security.py`

That let me delete or dramatically shrink the older integration-heavy files that had been carrying too many concerns at once. The new shape should give much clearer signal for MCP surface regressions, health/facade behavior, and attachment/query security rules.

### Benchmark campaign tooling

After the owner-file passes were in place, I added `devtools/benchmark_campaign.py` as a proper benchmark workflow.

The new tool defines named campaigns (`search-filters`, `storage`, `pipeline`), runs them with explicit thresholds, writes JSON and Markdown artifacts under `docs/benchmark-campaigns`, records git and machine metadata, and can regenerate an index README or compare two existing artifacts.

A typical run looks like this:

```bash
python -m devtools.benchmark_campaign run search-filters
python -m devtools.benchmark_campaign compare baseline.json candidate.json
```

instead of copying a fragile `pytest` command out of a comment block. Updated `tests/benchmarks/helpers.py` so the benchmark files use shared wrappers for connection-bound and async repository operations. That keeps the measurements focused on the operation under test rather than repeated setup boilerplate.

The follow-up fix in the next commit is important: benchmark runs now override repo-level `pytest` `addopts` and fail immediately if pytest exits without emitting a JSON artifact. That closes the failure mode where a benchmark “run” looked successful from the script’s perspective but produced nothing durable to compare.

### Tooling and baseline bookkeeping

Fixed `devtools/mutmut_campaign.py` so JSON and Markdown outputs are rooted against `ROOT` when the caller passes relative paths. That makes artifact locations deterministic from any working directory, which matters once the workflow is being scripted and repeated.

Finally, I refreshed the repository and Phase 5 baseline entries in `docs/mutation-testing-baseline.md` and tightened the provider fallback property in `tests/unit/core/test_json.py` so the unknown-provider generator excludes aliases using `normalize_provider_token()`. That keeps the property in line with the actual identity rules the runtime already uses.

## Testing

Coverage in this branch is intentionally distributed across the new owner surfaces rather than added to old umbrella files.

The main additions are:

- `tests/unit/storage/test_crud_laws.py`, `test_repository_laws.py`, and `test_store_ops.py` for storage/repository contracts
- `tests/unit/core/test_schema_annotations.py`, `test_schema_generation.py`, and `test_schema_validation.py` for schema phase 4 ownership
- `tests/unit/pipeline/test_*service.py`, `test_prepare_records.py`, `test_run_sources.py`, `test_runner_history.py`, and `test_runner_preview.py` for pipeline staging and orchestration
- `tests/unit/cli/test_click_app.py`, `test_command_surfaces.py`, `test_formatting.py`, `test_source_selection_helpers.py`, and `tests/integration/test_cli_query_mode.py` for CLI phase 5 ownership
- `tests/unit/core/test_facade_api.py`, `test_health_core.py`, `tests/unit/mcp/test_server_surfaces.py`, `test_tool_contracts.py`, `tests/unit/security/test_attachment_security.py`, and `tests/unit/storage/test_query_security.py` for phase 6 surfaces
- `tests/benchmarks/test_search_filters.py`, `test_storage.py`, and `test_pipeline.py` run through the shared benchmark helpers and the new benchmark campaign tooling
- `tests/unit/core/test_json.py` adds alias-aware provider fallback property coverage

## Notes

The most important migration concern is `AcquisitionService`: callers should no longer expect a scan-and-store convenience path inside that service. The benchmark tool is also new enough that its core logic is mostly exercised through the benchmark suite shape and its artifact contract rather than through a dedicated unit test module. The refreshed baselines show that timeout-heavy areas still remain, especially around `cli-run` observer/display paths and some pipeline service surfaces, but this branch gets the suite topology and benchmark workflow into a much more usable state for those follow-up passes.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 12 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 1 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #108, opened 2026-04-07
feat: add schema packages, artifact proofs, and site manifests

Sinity:
## Summary

I pulled three previously separate lines of work into one durable evidence path: raw artifacts land in an observation ledger, schema support is reported in package-aware proofs, and site builds publish typed manifests that point back to both the latest run and the proof summary. Before this, raw-corpus verification and the early proof work answered point-in-time questions, but they recomputed too much state and gave the site generator nothing durable to attach to published output. At the same time, provider schemas were starting to accumulate richer evidence, but the on-disk layout still treated flat provider schema files as the primary artifact instead of versioned packages with catalogs. The result was a mismatch: proofing talked about package versions, site generation wanted publication metadata, and storage only had raw blobs plus conversations. This change closes that gap by making artifacts, schema packages, and site publications first-class stored objects.

## Motivation

The immediate driver was operator visibility. I wanted `polylogue check` to answer not just “is the raw corpus valid?” but also “which raw artifacts are contract-backed, why did they resolve that way, and which Claude sidecars are still orphaned?” That question is expensive and awkward if every run has to rediscover artifact identity from raw bytes. Once I made that durable, the next friction point was publication: site output had no typed manifest or database record, so a successful build still left no durable pointer to the run or proof state that produced it. The schema package work pushed in the same direction. As soon as packages became the real unit of evidence, keeping flat provider schema files as the de facto authority made the rest of the reporting path feel second-class.

I considered keeping proofing ephemeral and only expanding the CLI report, but that would have left cohorting and site publication as separate implementations over the same derived facts. It was simpler to persist the observation ledger once and let proof, listing, cohorting, and publication all read from it.

### Durable artifact observation ledger

The core of the change is `polylogue/storage/artifact_observations.py`. It owns artifact identity via `artifact_observation_id()`, raw inspection via `inspect_raw_artifact()`, upsert/backfill via `ensure_artifact_observations()`, row listing, and cohort aggregation. The storage schema grows a durable `artifact_observations` table plus typed read models, and the acquisition path writes enough metadata to populate that ledger consistently.

That lets `polylogue/schemas/verification.py` separate two jobs more cleanly:

- `verify_raw_corpus()` still validates raw records against provider schemas.
- `prove_raw_artifact_coverage()` reports support status, package resolution, element kinds, resolution reasons, decode gaps, and Claude sidecar linkage.

`polylogue/cli/commands/check.py` then wires those facts through four operator surfaces: proof summary, observation rows, cohort summary, and JSON output.

```json
{
  "artifact_proof": {
    "summary": {
      "contract_backed_records": 123,
      "unknown_records": 2
    },
    "providers": {
      "claude-code": {
        "package_versions": {"v4": 17},
        "resolution_reasons": {"bundle_scope": 17}
      }
    }
  }
}
```

### Site publication manifests

I added `polylogue/publication.py` as the typed manifest model layer and taught `polylogue/site/builder.py` to emit and record a `SitePublicationManifest`. `polylogue/storage/backends/queries/publications.py` and the repository methods give the builder a durable publication record instead of relying on filesystem conventions alone.

The key effect is that a build carries its own provenance: outputs, latest run summary, and artifact proof summary travel together. `polylogue/site/publication_support.py` pulls the latest run and proof summaries from the DB so the builder does not need to know about proofing internals.

### Canonical versioned schema packages

The provider schema directories under `polylogue/schemas/providers/*` use `catalog.json` plus `versions/<version>/package.json` as the canonical layout. `polylogue/schemas/registry.py` grows search-root logic for bundled vs. data-home packages and reads and writes package catalogs directly. `polylogue/schemas/packages.py` and `polylogue/schemas/schema_generation.py` enrich those packages with representative paths, exact-structure IDs, bundle-scope counts, and profile-family metadata.

That turns package version reporting from a cosmetic label into an actual durable contract. The later report changes simply surface data that already lives in the package manifests.

### Pipeline boundary cleanup

As a prerequisite, I moved `RecordBundle`, `SaveResult`, and `save_bundle()` into `polylogue/pipeline/prepare.py` and removed them from `polylogue.sources`. The source layer should enumerate and decode exports; it should not own persistence bundle types that the pipeline needs independently. That cleanup makes the package and publication work sit on a cleaner boundary.

## Testing

I added focused coverage around each new surface:

- `tests/unit/core/test_verification.py` covers artifact proof summaries, package version tallies, resolution reasons, and sidecar linkage cases.
- `tests/unit/cli/test_check.py`, `tests/unit/cli/test_site.py`, and `tests/unit/cli/test_deterministic_output.py` cover the new CLI output paths.
- `tests/unit/site/test_builder.py` and `tests/unit/storage/test_publications.py` cover manifest generation and durable publication recording.
- `tests/unit/core/test_schema_registry.py`, `test_schema_registry_versions.py`, and `test_schema_generation.py` cover canonical package layout and enriched evidence metadata.
- `tests/unit/pipeline/test_prepare.py` keeps the moved bundle/save helpers honest after the import-path change.

## Notes

The docs-only commits in the middle of the branch were a temporary detour while I was shaping the artifact cohort model. They do not change the product surface, but they explain why otherwise unrelated planning material appears in the same feature branch.

The intentionally breaking part here is the import-path move for `RecordBundle`, `SaveResult`, and `save_bundle()`. Callers should import those from `polylogue.pipeline.prepare` now, not `polylogue.sources`.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 2 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550).

Sinity:
@codex review

PR #110, opened 2026-04-07
feat: add semantic preservation proofs across read surfaces

Sinity:
## Summary

This branch turns semantic proofing from a one-surface experiment into a first-class archive quality surface. I start by fixing a subtle correctness problem in the artifact proof path so stale durable observations can be refreshed against the current schema registry, then I add semantic proofs for canonical markdown, export surfaces, query summaries, query streams, and MCP read endpoints. Those proofs are surfaced in `polylogue check`, the showcase QA flow, and site publication manifests so they behave like an operator-facing gate instead of a private test helper. Add validation lanes that explicitly exercise the new read-surface contracts. The last few commits clean up the TUI read path so repository-bound screens share one base container and the screen composition remains explicit after dead config plumbing is removed.

## Motivation

The project already had strong snapshot and behavior tests around rendering and query output, but that still left a blind spot: we were asserting exact strings without explicitly stating which semantics a surface must preserve and which losses are intentional. That is fine for one renderer in isolation; it becomes brittle once the same conversation is exposed through canonical markdown, several export formats, streamed query output, and MCP detail/summary payloads. I wanted a single vocabulary that can say “message sections and attachments must survive”, “typed thinking markers are intentionally collapsed”, and “missing timestamps here is a real failure.”

The first proof correctness fix matters because semantic proofing is only useful if the underlying artifact proof is trustworthy. If a durable observation row can stay frozen in an older unsupported state even after the registry learns how to resolve it, the operator-facing surfaces will tell the wrong story. That is why the branch opens with a refresh path rather than jumping straight into new proof surfaces.

Did not want this to stay confined to unit tests. The check command, QA runner, and site publication manifest already carry other archive health signals. Semantic preservation belongs beside them.

### Refresh artifact proof resolution before semantic proofing

`polylogue/storage/artifact_observations.py` now refreshes eligible existing rows when proofing runs, rather than only backfilling missing observations. `polylogue/schemas/verification.py` routes proof through that refreshed ledger, and `tests/unit/core/test_verification.py` covers the stale-resolution case directly.

That change is small in code size but important in effect: semantic proof summaries now rest on the same up-to-date support classification an operator would expect after registry changes.

### Add a semantic proof model layer

The bulk of the change lives in `polylogue/rendering/semantic_proof.py`. I added:

- `SemanticMetricCheck`
- `SemanticConversationProof`
- `ProviderSemanticProof`
- `SemanticProofReport`
- `SemanticProofSuiteReport`

Those types let each surface express metric-level outcomes as one of three states:

- preserved
- declared loss
- critical loss

That distinction is what lets me prove semantics without pretending every surface is lossless. Canonical markdown should preserve message sections, attachment presence, timestamps, and roles. It does not need to preserve typed thinking/tool markers if the loss is documented and intentional.

### Extend proofing beyond canonical markdown

After the canonical markdown pass was in place, I expanded the same contract model across the read stack:

- export JSON-like, CSV, Markdown, Obsidian, Org, and HTML surfaces
- query summary text / JSON / CSV surfaces
- query stream plaintext / Markdown / JSON-lines surfaces
- MCP summary and detail surfaces

That work stays centered in `polylogue/rendering/semantic_proof.py`, but it touches `polylogue/rendering/renderers/html.py`, `polylogue/cli/query_output.py`, and MCP read contracts so the proof evaluators line up with real outputs.

```python
report = prove_export_surface_semantics(conversation, "export_html_v1", rendered)
suite = prove_semantic_surface_suite(projection, formatter, surfaces=["canonical_markdown_v1", "export_html_v1"])
```

The important part is not the helper signature; it is that all of these surfaces now speak the same contract language.

### Surface proof results in operator workflows

`polylogue/cli/commands/check.py` supports `--semantic-proof` plus provider/limit/offset filters and JSON output. `polylogue/publication.py`, `polylogue/showcase/qa_runner.py`, `polylogue/showcase/report.py`, and `polylogue/site/publication_support.py` add semantic-proof summaries to QA and publication metadata.

That means the proof is visible where operators already look:

- local `check` output
- QA JSON and Markdown
- site publication manifests

### Add validation lanes and clean up TUI read screens

I added explicit validation lanes in `devtools/run_validation_lanes.py` and documented them in `docs/test-quality-workflows.md`. The new lanes make query-routing, machine-contract, TUI, chaos, and live read-surface checks visible as named regression targets rather than implicit collections of tests.

At the end of the branch I also cleaned up the TUI screen stack. `polylogue/ui/tui/screens/base.py` now provides `RepositoryBoundContainer` so browser/search/dashboard screens stop duplicating repository acquisition logic. I briefly removed too much tab/config wiring while simplifying that path, and the last commit restores explicit `ScreenSpec` composition in `polylogue/ui/tui/app.py`.

## Testing

Coverage for this work is intentionally wide rather than deep in one module:

- `tests/unit/core/test_semantic_proof.py` covers canonical markdown, export, query stream, query summary, and MCP proof surfaces.
- `tests/unit/cli/test_check.py`, `test_click_app.py`, `test_deterministic_output.py`, `test_query_fmt.py`, and `test_site.py` cover CLI and publication rendering.
- `tests/unit/showcase/test_exercise_catalog.py`, `test_qa_runner.py`, and `test_report.py` cover QA/report integration.
- `tests/unit/ui/test_tui.py` covers the repository-bound TUI screen path and explicit tab composition.
- `tests/integration/test_cli_machine_contract.py`, `test_cli_query_mode.py`, `test_ingestion_chaos.py`, and related validation-lane coverage make the new named lanes real.

## Notes

The branch clearly splits into two phases: semantic proof expansion first, TUI cleanup second. The TUI work is smaller, but it is still relevant because those screens are another read surface over repository data. I kept it in the same branch because the repository-bound container extraction was part of the same “make read surfaces explicit and testable” direction.

There are no intended breaking API changes here. The point is to add proof and validation coverage while keeping existing output surfaces intact.

### DIFF
--- a/devtools/run_validation_lanes.py
+++ b/devtools/run_validation_lanes.py
@@ -1,0 +1,204 @@
+"""Run named validation lanes for the remaining operator frontier.
+
+Usage:
+    python -m devtools.run_validation_lanes --list
+    python -m devtools.run_validation_lanes --lane machine-contract
+    python -m devtools.run_validation_lanes --lane frontier-local --dry-run
+"""
+
+from __future__ import annotations
+
+import subprocess
+import sys
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class LaneConfig:
+    """Configuration for a validation lane."""
+
+    name: str
+    description: str
+    timeout_s: int
+    command: list[str] | None = None
+    sub_lanes: tuple[str, ...] = ()
+
+    @property
+    def is_composite(self) -> bool:
+        return bool(self.sub_lanes)
+
+
+LANES: dict[str, LaneConfig] = {
+    "machine-contract": LaneConfig(
+        name="machine-contract",
+        description="Root CLI JSON success/failure envelopes and runtime-health machine surfaces",
+        timeout_s=180,
+        command=[sys.executable, "-m", "pytest", "-q", "-n", "0", "-m", "machine_contract"],
+    ),
+    "query-routing": LaneConfig(
+        name="query-routing",
+        description="Query-first CLI route planning, integration, and streamed read-surface proofs",
+        timeout_s=240,
+        command=[sys.executable, "-m", "pytest", "-q", "-n", "0", "-m", "query_routing"],
+    ),
+    "tui": LaneConfig(
+        name="tui",
+        description="Textual Mission Control screens and interaction-state coverage",
+        timeout_s=240,
+        command=[sys.executable, "-m", "pytest", "-q", "-n", "0", "-m", "tui"],
+    ),
+    "chaos": LaneConfig(
+        name="chaos",
+        description="Hostility, interruption, and chronology integration coverage",
+        timeout_s=900,
+        command=[sys.executable, "-m", "pytest", "-q", "-n", "0", "-m", "chaos"],
+    ),
+    "scale-fast": LaneConfig(
+        name="scale-fast",
+        description="Fast storage scale budgets",
+        timeout_s=120,
+        command=[sys.executable, "-m", "devtools.run_scale_lanes", "--lane", "fast"],
+    ),
+    "scale-slow": LaneConfig(
+        name="scale-slow",
+        description="Slow local storage scale budgets",
+        timeout_s=360,
+        command=[sys.executable, "-m", "devtools.run_scale_lanes", "--lane", "slow"],
+    ),
+    "long-haul-small": LaneConfig(
+        name="long-haul-small",
+        description="Small reproducible benchmark/long-haul campaign",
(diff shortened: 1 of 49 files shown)


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 5 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 4 finding(s) confirmed fixed/stale.

Sinity:
@codex review

Issue #111, opened 2026-04-07
Archive surface modules are carrying too many responsibilities


Sinity:
## Context

Several core archive modules now serve as catch-all owners for unrelated behavior. `polylogue/cli/query_output.py` formats summary lists, computes grouped stats, and streams transcript output. `polylogue/site/rendering.py` handles conversation pages, provider indexes, dashboards, and template concerns in one place. `polylogue/rendering/semantic_proof.py` mixes metric models, fact extraction, and per-surface evaluators. `polylogue/health.py` mixes archive checks, runtime checks, cache concerns, and shared models. `polylogue/storage/repository.py` owns read, write, vector, metadata, and publication methods in one file. `polylogue/lib/filters.py` combines builder methods, runtime behavior, and type declarations. `polylogue/mcp/server.py` mixes prompt/resource/tool registration with support helpers.

These modules still work, but they are becoming increasingly expensive to review and change.

## Motivation

The problem is not just file size. The larger issue is responsibility mismatch. When one file owns multiple surfaces, a small change to one concern forces a reviewer to reload the entire neighboring subsystem. That increases risk in exactly the places where we are adding the most archive features: query routing, semantic proofing, publication metadata, and health/reporting.

This is also starting to distort testing. In several cases, the tests are surface-specific but the implementation is not. For example, query stream behavior and summary formatting should be testable and reviewable as distinct modes, but they currently live in the same implementation file. Site page generation and template setup have similar separation pressure. Semantic proof facts and semantic proof surfaces are related, but they are not the same responsibility.

The codebase needs smaller modules that map to actual archive surfaces without forcing a public-API rewrite.

## Scope

In scope:

- splitting query output into stream-oriented and summary-oriented helper modules while preserving the existing entry points
- splitting site generation into conversation-page, index-page, template, and publication-flow modules
- separating semantic proof facts/models/surfaces
- separating archive health from runtime health and shared health models
- splitting repository methods into read/write/vector families
- splitting filter builder/runtime/types concerns
- splitting MCP registration into support, prompt, resource, and tool modules
- extracting shared embedding-stats helpers where the same optional SQL logic is repeated

Out of scope:

- changing the behavior of query output formats
- redesigning the site structure or templates
- changing health semantics or repository semantics beyond the minimum needed to split the modules
- introducing new external dependencies

## Acceptance criteria

- Query stream, summary, and grouped-stats behavior can be maintained from focused modules instead of a single multi-mode implementation file.
[...]

PR #112, opened 2026-04-07
refactor: split archive surface workflows into focused modules

Sinity:
## Summary

This is a decomposition pass over several archive “god modules” that had accumulated multiple responsibilities over time. The goal is not to change the product surface; it is to make query output, site generation, semantic proofing, archive health, repository access, filter execution, and MCP registration reviewable in pieces that map to the surface they actually serve. The clearest examples were `polylogue/cli/query_output.py`, `polylogue/site/rendering.py`, `polylogue/rendering/semantic_proof.py`, `polylogue/health.py`, `polylogue/storage/repository.py`, `polylogue/lib/filters.py`, and `polylogue/mcp/server.py`. Each one had reached the point where a small feature change required opening a file that also owned unrelated formatting, orchestration, and helper logic. This branch splits those responsibilities while deliberately preserving the user-facing behavior.

## Motivation

The pattern behind all of these commits is the same: a surface-specific feature becomes hard to change because the implementation lives in a file that also owns three or four adjacent concerns. Query output is a good example. Streaming one conversation to stdout, formatting summary lists, and computing grouped stats all lived in `query_output.py`, even though they have different lifecycles and different tests. Site rendering had a similar problem: one file owned conversation page streaming, provider/index generation, and template setup. The same happened in semantic proofing, health checks, repository methods, and MCP registration.

I did not want to “rewrite” these areas. The safer approach was to extract one focused module at a time, keep the original import surface where it mattered, and let the old file become orchestration glue or a thin façade. That way callers keep working while the code underneath becomes easier to reason about.

### Normalize optional embedding stats

I started small with `polylogue/storage/embedding_stats.py`. The codebase had several places that wanted optional embedding counts, each with slightly different defensive SQL around missing tables or missing `sqlite-vec`. The new `EmbeddingStatsSnapshot`, `read_embedding_stats_sync()`, and `read_embedding_stats_async()` centralize that logic.

`polylogue/cli/commands/embed.py`, `polylogue/storage/repository.py`, and `polylogue/storage/search_providers/sqlite_vec.py` now all read the same helper instead of independently trapping `OperationalError`.

### Split query output by actual output mode

`polylogue/cli/query_output.py` had grown into a catch-all module. I split it into:

- `polylogue/cli/query_stream_output.py` for streamed headers/messages/footers and direct stdout flow
- `polylogue/cli/query_summary_output.py` for summary formatting, CSV generation, and grouped stats
- `polylogue/cli/query_output.py` as the stable façade/orchestrator

That keeps the existing call surface intact while making the implementation match the modes we actually support.

```python
from polylogue.cli.query_output import stream_conversation
# caller path stays the same

# implementation now lives in:
# - query_stream_output.py
# - query_summary_output.py
```

The same “stable façade, smaller implementation modules” pattern repeats through the rest of the branch.

### Split site generation and publication flow

`polylogue/site/rendering.py` now hands off work to:

- `polylogue/site/conversation_pages.py`
- `polylogue/site/index_pages.py`
- `polylogue/site/templates.py`

and `polylogue/site/builder.py` hands publication responsibilities to `polylogue/site/publication_flow.py`.

That separates “how do we stream a conversation page?”, “how do we build index/dashboard pages?”, and “how do we attach publication metadata?” into reviewable units.

### Split semantic proof, showcase reporting, and health layers

`polylogue/rendering/semantic_proof.py` is broken into facts, models, and surface evaluators. `polylogue/showcase/report.py` is split into common/session/file/report modules. `polylogue/health.py` becomes archive/runtime/model/cache layers.

The immediate benefit is that metrics, rendering, and orchestration stop being mixed in the same edit. The long-term benefit is that adding one more proof surface or one more health check no longer requires reopening a monolith.

### Split repository methods, filter runtime, and MCP registration

The lower-level decomposition follows the same rule:

- repository methods move to `repository_reads.py`, `repository_writes.py`, and `repository_vectors.py`
- filter behavior is separated into builder/runtime/types helpers
- MCP registration is separated into prompts/resources/support/tools helpers
- the `check` command uses `check_workflow.py` and `check_rendering.py`

None of those changes are meant to alter semantics. They are meant to make each layer legible on its own.

## Testing

The test story here is mostly regression preservation rather than new capability:

- `tests/unit/storage/test_embedding_stats.py` covers the new optional stats helper directly.
- Existing CLI query tests and snapshots continue to exercise the split query-output path.
- `tests/unit/site/test_builder.py` and related site tests keep the builder/publication flow honest after the extraction.
- `tests/unit/ui/test_tui.py` covers the prompt-plumbing cleanup.
- `tests/unit/cli/test_qa.py`, health tests, repository tests, and MCP tool-contract tests continue to cover the extracted modules via their unchanged public surfaces.

## Notes

This branch is intentionally repetitive. Once the extraction pattern proved safe on query output and site rendering, I applied the same rule to health, repository methods, filters, and MCP registration. There are no meaningful pivots or reverts here; the value comes from making many small, consistent boundary corrections rather than from one headline feature.

I have tried hard to keep import compatibility where that matters. In a few places, the old module now mostly re-exports helpers from the new files; that is deliberate. The refactor is successful only if downstream callers and snapshots keep seeing the same behavior while the implementation becomes easier to change.

Ref #111


Sinity:
@coderabbitai full review

Sinity:
@codex review

PR #114, opened 2026-04-07
perf: batch ingest work and remove raw-parse memory blow-ups

Sinity:
## Summary

The main goal here is to stop ingest from falling over on large raw archives while keeping the pipeline observable enough to tune. The most urgent issue was the parse stage loading too much raw material into memory at once, especially when raw rows are multi-megabyte JSONL payloads. I split the parsing service into batch, workflow, and state modules so the hot path can stay bounded and parallelized. I applied the same pattern to the pipeline runner so stage execution and run-state accumulation are explicit instead of hidden inside one large control-flow function. After that, I moved root query routing and showcase catalog loading behind clearer execution boundaries so the hot path and the QA path are both lighter-weight.

## Motivation

The performance problem is concrete: large raw archives can contain thousands of records whose raw blobs are themselves big JSON or JSONL payloads. Pulling too many of those through one service method is a reliable way to spike RSS and get the process killed before the parser can make progress. Even when the run survives, the current structure makes it hard to see which part of the pipeline needs attention because acquisition, validation, parsing, rendering, and indexing all live too close together in the orchestration layer.

I did not want to chase this only with smaller constants. The real fix is to put the heavy work on boundaries that make batching, caching, and stage accounting explicit. Once I started doing that, it became obvious the runner needed the same treatment, and then the root query front door and showcase catalog had the same “one file owns too much orchestration” smell.

### Bound the parse stage around batches and explicit worker state

`polylogue/pipeline/services/parsing.py` is mostly a service shell. The heavy lifting moved to:

- `polylogue/pipeline/services/parsing_batch.py`
- `polylogue/pipeline/services/parsing_models.py`
- `polylogue/pipeline/services/parsing_workflow.py`

The critical behavior change is that `parse_from_raw()` now stays intentionally batch-oriented. The service walks raw IDs in fixed-size chunks (`RAW_BATCH_SIZE = 50`), batch-loads raw records, precomputes candidate conversation IDs, loads one shared `PrepareCache` for the batch, and feeds prepared work items through an async worker queue.

```python
for batch_start in range(0, total, service.RAW_BATCH_SIZE):
    batch_ids = raw_ids[batch_start : batch_start + service.RAW_BATCH_SIZE]
    await service._process_raw_batch(backend, batch_ids, result, progress_callback)
```

That is the core of the memory fix. We still get concurrency, but we stop pretending the safe unit of work is “the whole raw corpus.”

### Make runner stage outcomes explicit

`polylogue/pipeline/runner.py` had become a dense orchestration file. I split it into:

- `run_stages.py` for stage execution
- `run_state.py` for count accumulation and finalization
- `run_support.py` for source selection, sync helpers, and run-JSON writing
- a slimmer `runner.py` as the top-level coordinator

This is partly cleanliness, but it is also performance work: you cannot tune or reason about a stage if its status, counts, and side effects are interleaved in one long method.

### Converge root query routing on one execution plan

The parse-stage work exposed a similar issue on the read side. Query execution and root CLI routing had enough duplicated planning logic that it was easy to lose control over batching and post-filter fetch size. `polylogue/lib/query_execution.py` owns a `ConversationQueryPlan` that makes SQL pushdown, post-filters, candidate fetch limits, and search behavior explicit. `polylogue/cli/root_request.py` and `polylogue/cli/query_frontdoor.py` route the root mode through that model.

This is not the headline performance win, but it matters because large archive reads need the same bounded behavior as large archive writes.

### Externalize the showcase catalog

Moved the showcase exercise definition to `polylogue/showcase/exercise_catalog.json` with `catalog_loader.py`, `exercise_models.py`, and `cli_boundary.py`. That keeps QA startup and CLI invocation from rebuilding the full exercise catalog from Python code every time.

## Testing

This branch leans on regression coverage more than on one giant benchmark:

- `tests/unit/cli/test_run.py` exercises the runner split.
- `tests/unit/cli/test_query_exec.py`, `tests/unit/cli/test_click_app.py`, and `tests/unit/core/test_filters_props.py` cover the query-front-door and execution-plan changes.
- `tests/unit/mcp/test_tool_contracts.py` keeps root routing and read surfaces aligned with MCP expectations.
- `tests/unit/showcase/test_runner.py` and `test_vhs_generation.py` cover the externalized showcase catalog path.

I did not add a dedicated large-archive benchmark in this branch. The batching constants and orchestration split are motivated by real memory pressure, but the explicit benchmark lane still needs to live elsewhere.

## Notes

The branch is more architectural than the final subject line suggests, but that is intentional. The performance fix depends on making the units of work real. Batching and parallelism are the visible win, but the supporting work—typed ingest models, stage outcome objects, root query planning, and externalized showcase definitions—is what makes the hot path small enough to tune without breaking unrelated behavior.


Sinity:
@codex review

PR #116, opened 2026-04-07
feat: overhaul schema inference and runtime registry boundaries

Sinity:
## Summary

This branch separates runtime schema authority from the heavier tooling layers that sit around inference, clustering, promotion, verification, and semantic proofing. The immediate change is `polylogue/schemas/runtime_registry.py`: runtime callers get a compact package-catalog and payload-resolution surface without importing diffing, promotion, or evidence-manifest helpers. Around that, I split schema observation, generation, synthetic corpus generation, prepare/transform boundaries, artifact verification, and semantic-surface contracts into focused modules that can be composed back together through explicit operator workflows. The end result is not just cleaner architecture; it is a stronger schema product surface. Parsing, emission, proofing, QA, and schema operators share one evidence model without every runtime path paying for the full tooling stack. Narrowed the package-root API so the public import surface better reflects that split.

## Motivation

`polylogue/schemas/registry.py` had become the clearest example of a mixed-responsibility module. Runtime callers needed a safe subset: load package catalogs, resolve payloads, and fetch schemas. Tooling callers needed much more: cluster manifests, schema diffs, promotion helpers, comparison output, evidence introspection, and on-disk rewrite support. Keeping those in the same class made the runtime surface larger than it needed to be and made internal changes harder to reason about.

The same coupling showed up elsewhere. `schema_generation.py` had grown into a monolith that mixed clustering, schema assembly, metadata annotation, and writeback. Synthetic corpus generation, verification, artifact persistence, and semantic-surface proofing were all trending the same way: one file per subsystem, each with a mix of models, orchestration, and low-level helpers. That slows down changes because a targeted improvement to one layer drags unrelated logic into the diff.

I did not want to fracture the product surface into disconnected helpers, though. The right shape is smaller modules internally, with explicit orchestration layers like `operator_workflow.py`, `roundtrip_proof.py`, and `semantic_surface_registry.py` recombining them where the operator experience actually needs a unified story.

### Split runtime schema authority from tooling

The first step is `polylogue/schemas/runtime_registry.py`. It owns:

- canonical schema-provider normalization
- package catalog discovery
- bundled-vs-data-home provider lookup
- package writes and version resolution
- element-schema fetches
- payload resolution

`polylogue/schemas/tooling_registry.py` owns the tooling-specific layer: `SchemaDiff`, `SchemaCluster`, `ClusterManifest`, promotion, comparison, and cluster persistence. `polylogue/schemas/registry.py` becomes the façade that combines the runtime base with `SchemaRegistryToolingMixin`.

That gives runtime paths a smaller dependency surface while keeping the old high-level registry story available for tooling callers.

```python
from polylogue.schemas.runtime_registry import SchemaRegistry  # runtime authority
from polylogue.schemas.registry import SchemaRegistry as ToolingSchemaRegistry  # full façade
```

### Extract schema observation and decompose generation

Schema observation helpers moved to `polylogue/schemas/observation.py`. That file owns schema-unit extraction, bundle-scope derivation, profile clustering, and fingerprint helpers that need to be safe for runtime callers and artifact inspection alike.

I then broke `schema_generation.py` apart into:

- `generation_analysis.py`
- `generation_models.py`
- `generation_support.py`
- `generation_workflow.py`

That split isolates cluster analysis, metadata/evidence assembly, schema annotation, and top-level workflow orchestration. It also makes the generation path easier to test at the level where a failure actually belongs.

### Split synthetic generation and prepare/source boundaries

Synthetic corpus generation gets the same treatment via:

- `schemas/synthetic/builders.py`
- `schemas/synthetic/models.py`
- `schemas/synthetic/runtime.py`
- `schemas/synthetic/selection.py`

The prepare/source boundary is also cleaned up through `prepare_transform.py`, `prepare_enrichment.py`, and `prepare_models.py`, with `sources/source.py` and `source_walk.py` taking on clearer ownership for traversal versus record transformation.

### Converge verification, artifact persistence, and operators

Verification is no longer one mixed module. I split it into:

- `verification_artifacts.py`
- `verification_corpus.py`
- `verification_models.py`
- `verification_requests.py`
- `verification_support.py`

and moved storage-specific pieces into:

- `storage/artifact_inspection.py`
- `storage/artifact_persistence.py`
- `storage/artifact_queries.py`

On top of that, `operator_workflow.py` now provides a single orchestration layer for infer/list/compare/promote/explain/audit/proof flows, and `roundtrip_proof.py` gives schema-evidence workflows explicit typed reports instead of ad hoc aggregation.

### Converge the semantic stack

The late commits centralize semantic-surface contracts in `polylogue/rendering/semantic_surface_registry.py` and split unified provider-message handling into:

- `schemas/unified_adapters.py`
- `schemas/unified_fallbacks.py`
- `schemas/unified_models.py`
- `schemas/unified_provider_meta.py`

That matters because semantic proofing, archive semantics, and schema/operator tooling all need the same normalized message story. The contracts live in one place without being implied across several modules.

### Narrow the package-root API

Finally, `polylogue/__init__.py`, `polylogue/lib/__init__.py`, and `docs/library-api.md` make the public surface more explicit. Archive-core types stay at the root; higher-order semantic/reporting helpers remain public, but they are imported from their precise modules instead of being casually re-exported through package roots.

## Testing

The branch touches a lot of tests because it moves a lot of boundaries:

- `tests/unit/core/test_schema_generation.py`, `test_sampling.py`, and `test_schema_registry*.py` cover the runtime/tooling split and generation-path decomposition.
- `tests/integration/test_schema_operator_workflow.py` and `test_schema_evidence_roundtrip_lane.py` cover the new operator workflow and roundtrip proof surfaces.
- `tests/unit/core/test_synthetic_semantics.py` covers the split synthetic stack.
- `tests/unit/cli/test_check.py`, `test_qa.py`, `test_terminal_snapshots.py`, and related showcase tests cover the converged CLI/QA/schema flows.
- `tests/unit/core/test_semantic_facts.py` and `tests/unit/core/test_semantic_proof.py` cover the semantic-stack convergence.
- Pipeline/backend tests such as `test_parsing_service.py`, `test_resilience.py`, `test_async_index.py`, and `test_backend.py` keep the write-path narrowing honest.

## Notes

This branch is intentionally two-phase. The early commits do extraction work; the late commits do convergence work. That is why the final diff contains both “split this monolith apart” files and “pull the pieces back together through one operator surface” files. The package-root API narrowing is the only deliberately breaking end-user change, and it follows directly from the runtime/tooling boundary becoming explicit.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 24 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 10 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #118, opened 2026-04-07
feat: add action-aware query filters and persisted retrieval

Sinity:
## Summary

This branch turns tool-derived semantic actions into first-class query and retrieval primitives. The user-facing part is simple: CLI and MCP query surfaces can filter by action type and normalized tool name, inspect grouped stats by action or tool, require ordered action sequences, search action evidence text, and choose explicit retrieval lanes. Underneath that, I move from ad hoc semantic facts toward a persisted `ActionEvent` model with its own FTS index so action-aware retrieval does not have to hydrate entire conversations just to answer “show me the Bash runs that touched this file” or “find the session where `pytest -q` appeared in tool output.” The branch also restores the synchronous `_run()` seam, hardens Gemini/Codex fidelity, and moves stage/view models into typed modules because the retrieval layer depends on those contracts being explicit and stable. The final result is a query stack that can reason about work, not just dialogue text.

## Motivation

Polylogue already knew a lot about conversation semantics, but the query surface still treated most of that knowledge as derived view logic instead of queryable archive facts. That is fine if you only want text search or provider/date filters. It breaks down as soon as the operator intent becomes semantic: “find file-edit sessions,” “group results by tool,” “require the sequence file_read -> file_edit -> shell,” or “rank conversations by tool-call evidence rather than only by dialogue text.”

I could have solved some of that with more hydrated post-filtering, but that does not scale well and it does not give us a durable retrieval lane. I wanted action-aware search to work as a real search primitive, not just as a higher-level filter over fully materialized conversations. That required two things:

1. a normalized semantic event model with stable IDs and search text
2. persisted storage/index support so action retrieval can be ranked and fused with existing dialogue/vector lanes

The early compatibility and fidelity commits exist for that reason. If sync callers cannot still force coroutines predictably, if stage/view models move around implicitly, or if Gemini/Codex semantic extraction is lossy, the new retrieval surface will sit on shaky ground.

### Restore compatibility seams and typed state models

The branch opens by restoring `_run()` in `polylogue/sync.py` so downstream sync integrations can continue to execute terminal awaitables through a canonical helper. Move pipeline stage models into `polylogue/pipeline/stage_models.py` and non-durable state/view models into `polylogue/storage/state_views.py`.

At the same time, `prepare_records()` grows a typed return contract via `PersistedConversationResult`, with `prepare_bundle()` and `persist_prepared_bundle()` split out as explicit stages.

```python
result = await prepare_records(...)
conversation_id = result.conversation_id
counts = result.counts
changed = result.content_changed
```

That is partly cleanup, but it is also necessary foundation for the later retrieval and indexing work.

### Improve semantic fidelity and normalize action events

The next layer is semantic truth. I tighten Gemini/Codex archive fidelity and extend semantic extraction so the archive can consistently recover paths, commands, queries, URLs, branches, and tool-result output from real provider payloads and stored content blocks.

`polylogue/lib/action_events.py` becomes the central model. It takes tool calls and content-block data and produces stable `ActionEvent` instances with normalized search text. This replaces the idea that action search should be assembled ad hoc inside query execution.

### Add action and tool filters to CLI and MCP

Once that model exists, I surface it through `polylogue/lib/query_spec.py`, `polylogue/lib/filter_builder.py`, `polylogue/lib/query_execution.py`, `polylogue/cli/click_app.py`, and `polylogue/mcp/server_tools.py`.

The new surface includes:

- `--action` / `--exclude-action`
- `--tool` / `--exclude-tool`
- `--action-sequence`
- `--action-text`
- `--stats-by action`
- `--stats-by tool`
- `--retrieval-lane auto|dialogue|actions|hybrid`

`polylogue/cli/query_summary_output.py` knows how to emit grouped semantic stats by action or tool, and MCP tools see the same query vocabulary, not a thinner parallel surface.

### Converge on one action model instead of multiple experiments

The branch deliberately replaces an earlier, smaller action-filter experiment with the action-event layer. That is why there are commits that first add action filters, then remove a first formulation, then add `action_facts`, then replace that with `action_events`. I did not want two overlapping semantic abstractions for the same work. The converged model is:

- action-aware facts/events for semantics
- one query-spec vocabulary
- one execution layer that can decide between hydrated scans and persisted search

### Add explicit retrieval lanes and persisted action search

The last step is persistence. `polylogue/storage/backends/schema.py` adds `action_events_fts` and the trigger/rebuild story for keeping it in sync. `polylogue/storage/search.py` adds ranked action-search SQL. `polylogue/storage/repository_reads.py` and the query store add `search_actions()` and persisted action-record lookup. `polylogue/storage/fts_lifecycle.py` handles rebuild/update refresh behavior.

That gives us explicit retrieval lanes in `polylogue/lib/query_execution.py`:

- `dialogue` for existing dialogue-text retrieval
- `actions` for action-event retrieval
- `hybrid` for fused dialogue/action/vector ranking
- `auto` for the default policy

This is the real capability jump in the branch. Action-aware retrieval stops being a hydrated scan fallback and becomes a durable search surface.

## Testing

The test footprint is wide because the feature spans compatibility, semantics, query routing, and storage:

- `tests/unit/core/test_sync.py` covers restored `_run()`.
- `tests/unit/pipeline/test_prepare_records.py`, `test_branching.py`, `test_run_sources.py`, and `test_parse_tracking.py` cover the typed state/prepare changes.
- `tests/unit/sources/test_parsers_drive.py`, `test_parsers_codex.py`, `test_drive_ops.py`, and `tests/unit/core/test_semantic_facts.py` cover semantic fidelity improvements.
- `tests/unit/cli/test_click_app.py`, `test_query_exec.py`, `test_query_exec_laws.py`, and `test_query_fmt.py` cover new flags, grouped stats, and retrieval lanes.
- `tests/unit/mcp/test_tool_contracts.py` keeps MCP parity with the CLI surface.
- `tests/unit/storage/test_store_ops.py` and `tests/unit/storage/test_fts5.py` cover action-aware persistence and FTS refresh behavior.
- `tests/unit/devtools/test_validation_lanes.py` and `test_query_memory_budget.py` cover the new validation/tooling lanes.

## Notes

This branch is intentionally layered. The early fixes are not accidental prelude; they are prerequisites for the later retrieval work. The one clear dead end is the first generation of action filters. I kept the better parts—semantic intent and grouped stats—but converged the implementation on `ActionEvent` plus persisted search instead of trying to keep multiple semantic abstractions alive at once.


Sinity:
@codex review

PR #120, opened 2026-04-07
feat: add action-event read model and runtime introspection

Sinity:
## Summary

I replaced the last major chunk of action-aware query behavior that still depended on ad hoc `content_blocks` scans with a durable `action_events` read model, and then wired the rest of the runtime to treat that model as the canonical source for action/tool filtering, grouped stats, repair, and health. In the same pass I expanded embedding-status introspection so operators can see freshness, stale rows, missing provenance, model mix, and timestamp spread instead of a single coverage percentage. The branch also normalizes MCP query-spec construction and consolidates the session helper APIs under `polylogue.lib.session_products`, which makes the public runtime surface line up with the new read model rather than the older per-conversation derivation path. The net effect is that action-heavy queries stop paying the full semantic-hydration cost when the archive already has canonical rows, while still falling back safely when the read model is missing or out of date.

## Motivation

Entry 59 introduced semantic actions and tool-aware filters, but the implementation still had two structural problems:

1. action-aware query planning and grouped stats were repeatedly re-deriving semantics from `content_blocks` and hydrated conversations
2. embedding status only described coverage, not whether the vectors were fresh enough to power retrieval safely

That combination made the runtime hard to reason about. A surface could report “embeddings ready” while still containing stale or provenance-less rows, and action-aware stats could degrade to expensive semantic reconstruction even when the source archive had already stabilized.

I did not want to push more logic into `content_blocks`-based SQL, because that would keep duplicating partial action extraction rules in multiple places. Instead I introduced one canonical materialization step that turns tool-use blocks into durable action rows, then taught query planning and health to interrogate that single surface. Once that existed, it was worth tightening the public helper API as well: `build_session_profile()`, `build_session_threads()`, and auto-tag inference belong together, and the MCP tools needed one shared `ConversationQuerySpec` construction path rather than hand-written parameter normalization in each endpoint.

### Durable action-event rows and rebuild flow

The core of the change is the new materialized action-event layer:

- `polylogue/storage/action_event_rows.py`
- `polylogue/storage/action_event_lifecycle.py`
- `polylogue/storage/backends/queries/action_events.py`
- `polylogue/storage/backends/schema.py`

`action_event_rows.py` is the canonical bridge between message/content-block records and durable action rows. It attaches blocks to messages, rebuilds normalized `ActionEvent` values, and serializes those back to `ActionEventRecord` rows with sort keys, branch names, affected paths, command/query/url fields, and search text.

`action_event_lifecycle.py` then owns the read-model lifecycle:

```python
status = action_event_read_model_status_sync(conn)
rebuild_action_event_read_model_sync(conn, conversation_ids=candidate_ids)
```

That gives the runtime a stable answer to “is the action read model ready?” instead of inferring readiness from whatever the current query path happened to touch.

On the storage side I added the `action_events` table and indexes, kept the existing action FTS companion, and wired async query helpers for single-conversation and batch reads. Updated indexing so full rebuilds and targeted updates rebuild action rows before refreshing FTS state.

### Query planning, grouped stats, and safe fallback behavior

The read model only pays off if the query runtime can decide when to rely on it. I extended `ConversationQueryPlan` in `polylogue/lib/query_execution.py` with readiness checks and repository-aware candidate selection so action-specific filters only stay SQL-pushed when the canonical rows are actually available.

The important behavioral change is this split:

```python
if await self._action_read_model_ready(repository):
    return self.record_query, self.sql_pushed
return self.record_query.without_unstable_semantic_filters(), False
```

That lets action-heavy queries stay fast on a migrated archive while still returning correct results on a partially migrated one. `polylogue/cli/query_summary_output.py` picks up the same logic for action/tool grouped stats: it reads `repo.get_action_events_batch()` when the durable rows are ready, and only falls back to `build_conversation_semantic_facts()` when it has to.

Added `polylogue/mcp/query_support.py` and moved MCP query-spec normalization there, which removes a lot of repeated “tool/action/path/has_thinking” translation in prompts, tools, and resources.

### Embedding freshness and operator introspection

The other half of the branch is operator-facing introspection.

`polylogue/cli/commands/embed.py`, `polylogue/lib/stats.py`, `polylogue/health_archive.py`, and `polylogue/mcp/payloads.py` expose a richer status payload with fields such as:

```json
{
  "freshness_status": "stale",
  "stale_messages": 42,
  "messages_missing_provenance": 3,
  "embedding_models": {"voyage-3-large": 420},
  "embedding_dimensions": {1024: 420}
}
```

Changed `retrieval_ready` to mean “there is at least one fresh embedded message”, not merely “there are any embedding rows at all”. That is a much better contract for both CLI and MCP consumers.

Health now reports:
- orphaned content blocks
- action-event read-model readiness
- action FTS completeness
- embedding freshness as a separate concern from embedding coverage

### Public helper surface cleanup

The second feature-branch commit was partly cleanup, but it was purposeful cleanup around the new runtime boundary.

`docs/library-api.md` now documents `polylogue.lib.session_products` as the explicit import point for:

- `build_session_profile`
- `build_session_threads`
- `infer_auto_tags`

At the same time I removed `ConversationFilter` from the `polylogue` package root and narrowed some lazy exports so archive-core imports do not keep dragging higher-order helper modules along implicitly.

## Testing

I covered the read-model and introspection work from several angles:

- `tests/unit/cli/test_embed.py` exercises the expanded embedding status payload and stale/provenance reporting
- `tests/unit/cli/test_query_exec.py` covers action/tool stats, read-model readiness checks, and fallback behavior
- `tests/unit/storage/test_fts5.py` covers action-event status, FTS synchronization, and health-cache interactions
- `tests/integration/test_health.py` exercises the new health checks and repair behavior
- `tests/unit/pipeline/test_async_index.py` and `tests/unit/pipeline/test_indexing.py` cover rebuild/index wiring
- the benchmark updates in `devtools/benchmark_campaigns.py` keep the internal call sites aligned with the newer query-store APIs

## Notes

The main compatibility changes are import-path and export-surface changes. Code importing `ConversationFilter` from `polylogue` needs to move to `polylogue.lib.filters`, and code importing session helpers from `polylogue.lib.session_profile`, `polylogue.lib.threads`, or `polylogue.lib.tagging` should move to `polylogue.lib.session_products`.

I did not add any new third-party dependencies here. The repository-level deletions (`.coderabbit.yaml`, GitHub release metadata, CodeQL workflow) and the LICENSE update are real, but they are incidental to the runtime/storage work rather than the design center of the branch.

## Summary by CodeRabbit

* **New Features**
  * Added embedding freshness and staleness tracking to monitor which embedded messages are current vs. stale
  * Added action event read model for improved semantic search and action tracking
  * Extended health checks with orphaned data detection and repair capabilities

* **Improvements**
  * Enhanced embedding statistics with provenance gaps and time range bounds
  * Improved health reporting with action event and data integrity checks
  * Better query execution paths for semantic filtering

* **Refactor**
  * Consolidated session profile and tagging functionality into unified API


Sinity:
@coderabbitai full review

Sinity:

Sinity:
@codex review

PR #126, opened 2026-04-07
refactor: collapse product transport and retire obsolete proof layers

Sinity:
## Summary

This change finishes a second decomposition wave by splitting the remaining monolithic roots, introducing a registry-driven product transport layer, and removing proof/governance surfaces that were no longer worth carrying as first-class CLI, MCP, and health families. I split the last heavy compatibility roots (`check` rendering, `embed`, `paths`, artifact taxonomy, schema observation, storage record models, schema DDL fragments, and several rendering/schema helpers) into narrower modules, but I did not stop at extraction. Once the split boundaries were in place, I used a new generic product registry plus a consolidated `polylogue.health` module to delete a large amount of obsolete semantic-proof, governance, and maintenance-lineage infrastructure. The result is a codebase with fewer parallel transport trees and fewer low-value surfaces to maintain. This intentionally includes breaking surface changes for scripts that still depended on semantic-proof flags, governance product commands/tools, or some of the old health/archive-product imports.

## Motivation

After the earlier decomposition work, the codebase still had two related problems.

The first was that several “root” modules were still doing too much. `polylogue/cli/check_rendering.py`, `polylogue/cli/commands/embed.py`, `polylogue/paths.py`, `polylogue/lib/artifact_taxonomy.py`, `polylogue/schemas/observation.py`, `polylogue/storage/store_runtime_records.py`, and `polylogue/storage/backends/schema_ddl_products.py` were all acting as compatibility, model, runtime, and support modules at the same time. That made them look stable from the outside, but they were still hotspots internally.

The second problem was that some surfaces no longer justified their own dedicated stacks. The semantic-proof feature family had its own rendering modules, surface registries, check options, snapshot coverage, site/showcase hooks, and tests. The governance/readiness product family had the same shape across product commands, product rendering, product workflows, archive-product exports, and MCP tools. The product transport layer as a whole also repeated the same metadata in four places: command registration, item rendering, workflow dispatch, and MCP registration. Health reporting had similar duplication across `health_models.py`, `health_runtime.py`, `health_archive.py`, and `health_cache.py`.

The branch history shows a clear pivot. I first split the remaining roots and introduced a small set of reusable abstractions (`polylogue/products/registry.py`, `rendering/blocks.py`, `schemas/extraction.py`, `schemas/pinning.py`, and `showcase/context.py`). Once that groundwork existed, I chose deletion and consolidation rather than yet another round of one-file-to-three-file decomposition for the obsolete proof and governance layers.

### Finish the remaining root splits

I started by cleaning up the roots that were still mixing orchestration with concrete runtime logic.

`polylogue/cli/check_rendering.py` is just a wrapper over dedicated JSON and plain/rich rendering modules. `polylogue/cli/commands/embed.py` delegates execution to `embed_runtime.py` and statistics/reporting to `embed_stats.py`. `polylogue/paths.py` keeps the familiar import surface, but the actual responsibilities are split across roots, config accessors, data models, and sanitization helpers.

I applied the same pattern to artifact observation and storage contracts. `polylogue/lib/artifact_taxonomy.py` delegates to explicit model/runtime/support modules, and `polylogue/schemas/observation.py` delegates to identity/model/runtime modules. Storage record models and product DDL fragments got the same treatment through `store_runtime_action_records.py`, `store_runtime_archive_records.py`, `store_runtime_raw_records.py`, and the split `schema_ddl_product_*` files.

This is the general shape of the compatibility roots after the split:

```python
from polylogue.paths_config import get_drive_config, get_index_config, get_sources
from polylogue.paths_models import DriveConfig, IndexConfig, Source
from polylogue.paths_roots import archive_root, db_path, render_root
from polylogue.paths_sanitize import conversation_render_root, is_within_root, safe_path_component
```

That pattern matters because it keeps stable imports while making the real seams explicit.

### Build generic transport primitives instead of repeating product metadata

The bigger architectural move in this branch is the product registry. Before this change, every derived product family needed matching entries in product commands, product rendering, product workflow dispatch, and MCP tool registration. That duplication made the product surface hard to shrink because deleting a product family meant editing four different trees.

I introduced `polylogue/products/registry.py` as the shared descriptor layer for those surfaces:

```python
@dataclass(frozen=True)
class ProductType:
    name: str
    display_name: str
    json_key: str
    query_class_path: str = ""
    operations_method: str = ""
    cli_options: list[CliOption] = field(default_factory=list)
```

From there, the CLI and MCP can build their surfaces generically without carrying family-specific registrars forever. `polylogue/cli/commands/products.py` now constructs commands from the registry, `polylogue/cli/products_rendering.py` delegates to generic rendering by product descriptor, `polylogue/cli/products_workflow.py` delegates fetches through the registry, and `polylogue/mcp/server_product_tools.py` registers list-style tools from the same source of truth.

That removes a lot of mechanical duplication while keeping product semantics where they belong: the row builders, lifecycle code, and storage queries still live in regular code, but transport/presentation metadata is no longer forked across multiple surfaces.

### Retire semantic-proof infrastructure instead of decomposing it forever

Once the supporting abstractions existed, I made a deliberate choice not to keep decomposing the semantic-proof stack. The branch deletes the semantic-proof and semantic-surface families under `polylogue/rendering/semantic_proof*.py` and `polylogue/rendering/semantic_surface*.py`, removes the corresponding fields from `check_workflow.py`, trims the `check` option model and validation rules, and updates snapshots and tests to stop expecting those surfaces.

This is not a hidden internal refactor. It removes user-facing `check` surfaces and test contracts that were previously exposed. The help snapshot clearly loses the semantic-proof options, and the workflow/result model no longer carries `semantic_report` or `semantic_contracts`.

I preferred that outcome to keeping a large set of proof-specific modules alive simply because they had already been split. The branch history here is important: there was an intermediate point where the proof stack was decomposed more cleanly, but the final call is that the whole family is obsolete enough to remove.

### Consolidate health and remove governance-only product surfaces

Health reporting had a similar problem: type definitions, runtime checks, archive checks, and cache helpers were spread across multiple roots and then re-imported all over the CLI and tests. I replaced that with a single `polylogue.health` module that exports the canonical health/report types and the runtime/archive logic.

Call sites move to this kind of import:

```python
from polylogue.health import HealthCheck, HealthReport, VerifyStatus, run_runtime_health
```

At the same time, I removed the governance-only product surfaces. `products status`, `products maintenance`, and `products debt` are no longer registered as standalone command modules, and the matching MCP tools (`archive_debt` and `maintenance_runs`) are gone as dedicated product-tool registrations. The Python archive-product surface also shrinks: maintenance lineage query/model exports and archive-debt lineage payloads are removed from `polylogue.archive_products`.

This is the other major reason I wanted the registry in place first. Once the generic transport exists, it becomes practical to say “only keep the product families we still want to expose.”

## Testing

The test updates are concentrated around the removed and consolidated surfaces. `tests/unit/cli/test_check.py`, `tests/unit/cli/test_check_runtime.py`, `tests/unit/core/test_health_core.py`, `tests/unit/cli/test_products.py`, `tests/unit/mcp/test_tool_contracts.py`, `tests/unit/mcp/test_server_surfaces.py`, `tests/unit/storage/test_fts5.py`, and the CLI snapshot all move to the consolidated health/product behavior.

On the removal side, `tests/unit/core/test_semantic_proof.py` is deleted, and semantic-proof expectations are removed from `tests/unit/cli/test_click_app.py`, `tests/unit/site/test_builder.py`, `tests/unit/showcase/test_qa_runner.py`, and `tests/unit/showcase/test_report.py`. That is intentional coverage deletion because the underlying surface is being removed, not just reorganized.

One thing to watch in review is that the new registry dispatch is mostly validated through the updated CLI/MCP tests rather than through a dedicated new “registry unit test” file. The coverage is there, but it is indirect.

## Notes

This branch intentionally mixes two kinds of refactor: compatibility-preserving root splits and non-compatible surface removals. The split-only parts (`paths`, artifact taxonomy, observation, storage records, check/embed helpers) are meant to make internals easier to evolve while keeping familiar import surfaces. The semantic-proof, governance-product, and maintenance-lineage removals are different: they are deliberate contractions of the supported surface area.

There are no dependency changes here. The cost of the branch is migration for any remaining callers that still depended on removed proof/governance/health imports or command/tool names; the benefit is a much smaller amount of parallel transport infrastructure to keep in sync.


Sinity:
@codex review

PR #128, opened 2026-04-07
refactor: collapse query, storage, schema, and MCP surfaces

Sinity:
## Summary

I collapsed a large amount of leaf-module sprawl across the query CLI, storage backend, schema runtime, MCP payloads, and session-product helpers into a smaller set of modules that actually reflect where the ownership boundaries are. The immediate goal was not to invent new behavior; it was to stop forcing readers to bounce across dozens of one-function or one-family files before they can understand the flow of a query, a schema lookup, or a session-product rebuild. On the CLI side, the query-first surface lives primarily in `polylogue/cli/query.py`, `polylogue/cli/query_output.py`, and the new `polylogue/cli/query_stats.py`. On the storage and schema side, the async SQLite backend, runtime registry, MCP payload contracts, and session-product builders have been pulled back into cohesive modules with fewer re-export layers. Replaced some deleted proof-style scaffolding with direct property tests and finished the branch with a targeted fix to work-event classification so the refactor does not quietly degrade semantics.

## Motivation

The repository had accumulated a pattern where every concept eventually split into a root module, a support module, a runtime module, and several tiny specialization files. That makes sense while the concepts are still unstable, but once the interfaces settle it becomes counterproductive: reading a query path meant opening `query.py`, then `query_plan.py`, then `query_helpers.py`, then `query_summary_output.py`, then one of several grouped-stats leaves. The same thing had happened in the async SQLite backend, the runtime schema registry, the MCP payload layer, and the session-product row builders.

I wanted to get back to a codebase where the import graph communicates ownership instead of indirection. If a reviewer wants to understand how query routing works, they should be able to stay inside one query module. If they want to understand how runtime schema packages are loaded, they should not need to jump through path, resolution, support, and write mixins. The branch is large because I took that principle seriously across multiple layers at once rather than shaving off one file family at a time.

### Query CLI: three modules instead of a constellation

The most visible cleanup is in the query-first CLI. I folded routing, plan construction, filter-description helpers, and default-mode dispatch into `polylogue/cli/query.py`. That file owns the query execution primitives directly:

```python
class QueryAction(str, Enum): ...
class QueryRoute(str, Enum): ...
@dataclass(frozen=True)
class QueryExecutionPlan: ...
class QueryFirstGroupBase(click.Group): ...
def handle_query_mode(...): ...
```

The supporting output side got the same treatment. `polylogue/cli/query_output.py` owns formatted output, delivery, streaming, summary-list rendering, and the main dispatch entrypoint. Grouped stats that were previously scattered across `query_grouped_stats_*`, `query_sql_stats.py`, and `query_profile_stats.py` were consolidated into the new `polylogue/cli/query_stats.py`.

The effect is that `click_app.py` only has to import stable top-level surfaces without threading together a chain of helper leaves. It also means future behavior changes can be reviewed at the actual boundary where they happen.

### Schema runtime: stop distributing one registry across many files

`polylogue/schemas/runtime_registry.py` had become a thin root wrapped around multiple resolution/path/support modules. I inlined the actual authority back into the registry so provider listing, catalog loading, version resolution, schema element lookup, and package writes live together. The file is longer, but it is also much more legible because the workflow is visible in one place.

The same principle applies to the roundtrip and verification helpers: this branch deletes several auxiliary “proof” and “support” leaves where the only real effect was to make the schema path harder to trace.

### MCP payloads: one contract surface

The MCP JSON contracts were previously split across `payload_base.py`, `payload_archive.py`, `payload_conversations.py`, and `payload_health.py`. I collapsed those into `polylogue/mcp/payloads.py` so the payload layer reads as one coherent contract family. That module now defines the base models plus the concrete summary/detail/stats/mutation/health payloads together.

This is the kind of consolidation that makes maintenance easier later: reviewers can see the full outbound MCP shape without reconstructing it from imports.

### Storage and session products: collapse support strata

The async SQLite backend had also picked up too many layers: connection setup, transactions, derived actions, derived stats, and runtime helpers were spread across a family of `async_sqlite_*` leaves. I merged those back into `polylogue/storage/backends/async_sqlite.py` so connection configuration, writer locks, bulk connections, and public connection helpers sit together.

I applied the same cleanup to session products. Rather than maintaining separate row/text/signal/helper modules for profiles and phases, this branch consolidates that logic into modules named after the product concepts: `session_product_profiles.py` and `session_product_timeline_rows.py`. The rebuild/refresh path is easier to follow when it manipulates profiles, work events, phases, and threads directly instead of routing through generic support leaves.

### Tests and invariants: keep the behavior checks, drop the scaffolding

A large deletion-heavy refactor needs explicit invariant coverage. I removed older proof-style scaffolding, but I did not want to lose the guarantees it was trying to protect. Instead I added direct property tests:

- `tests/property/test_rendering_preservation.py`
- `tests/property/test_schema_roundtrip.py`

Those are much closer to the behavior we actually care about. They say “rendering preserves important structure” and “schema extraction roundtrips correctly” without requiring an extra framework vocabulary around them.

Finished the branch with a focused semantic correction in `polylogue/lib/work_event_extraction.py` so the session-product stack still classifies work events accurately after all the file-family merges.

## Testing

The branch updates and preserves coverage across the areas it touches:

- `tests/unit/cli/test_query_exec.py`, `test_query_exec_laws.py`, and `test_query_fmt.py` for the consolidated query surface
- `tests/property/test_rendering_preservation.py` and `test_schema_roundtrip.py` for the deleted proof infrastructure’s core invariants
- storage, facade, and CLI product tests touched by the session-product consolidation

## Notes

This is deliberately a structural refactor, so some import paths disappear even though the runtime behavior is largely preserved. That is intentional. I would rather have a few clear module boundaries than preserve dozens of internal paths that only existed because the codebase had been decomposed past the point of readability.


Sinity:
@codex review

PR #132, opened 2026-04-07
perf: SQLite tuning, FTS trigger suspension, and session product fixes

Sinity:
## Summary

I used this branch to keep two ideas aligned: bulk ingest should run like a bulk operation rather than a series of tiny SQLite maintenance steps, and session-product payloads should not pretend that weak heuristics are solid facts. The performance side of the change adds the SQLite pragmas we actually need for large archives and suspends FTS triggers while the pipeline is doing heavy batch work. The product side removes heuristic fields such as `primary_work_kind`, `decisions`, phase `kind`, and `refined_work_kind` from the public/session payloads and adjusts hydration so legacy rows can still be read until they are rebuilt. Cleaned up Claude Code title/work-event text generation so transport artifacts do not leak into user-facing summaries. The branch wandered through some broader acquisition experiments, but the durable changes that stayed together are the ones named in the squash message: SQLite tuning, FTS trigger suspension, and session-product contract cleanup.

## Motivation

Two separate problems were showing up in the code.

The first was performance. A large SQLite archive cannot be treated like a toy database; default cache sizing, default checkpoint cadence, and per-row FTS trigger updates are all disproportionately expensive once the corpus is large. The code already had bulk pipeline stages, but it was still paying work that belonged to steady-state single-row writes rather than big batch imports.

The second was semantic honesty. Session products had accumulated fields like `primary_work_kind`, `decision_signal_strength`, `decisions`, phase `kind`, and `refined_work_kind`. Reading through the extraction code made it clear that these were driven by dominant-tool counts, keyword matches, and proposal/acceptance phrase heuristics. That is fine for exploratory analytics, but once those values are persisted into public product payloads they start to look authoritative. I wanted to tighten the contract so we keep durable evidence and confidence but stop shipping fields whose names imply more certainty than the code can justify.

### SQLite connections and transactions

Both `polylogue/storage/backends/async_sqlite.py` and `polylogue/storage/backends/connection.py` now apply the same performance pragmas on open:

```python
PRAGMA cache_size = -524288
PRAGMA synchronous = NORMAL
PRAGMA mmap_size = 1073741824
PRAGMA temp_store = MEMORY
PRAGMA wal_autocheckpoint = 10000
```

These are not arbitrary micro-tweaks; they are the baseline settings that make a large WAL-backed archive behave predictably. Tightened the bulk-connection path so nested work uses savepoints against the existing bulk connection instead of fighting for additional writer connections.

### Suspend FTS trigger work during bulk pipeline stages

In `polylogue/storage/fts_lifecycle.py` I added explicit helpers to drop and recreate the `messages_fts_*` and `action_events_fts_*` triggers. `polylogue/pipeline/run_execution.py` uses those helpers to suspend triggers around the bulk ingest stages and restore them afterward.

The important point is not that triggers are “bad”; it is that they are the wrong mechanism during a large batch where we are going to rebuild/search-sync anyway. Paying per-row trigger overhead in that context is pure waste.

### Remove brittle session-product fields

The branch drops a cluster of heuristics from the session-product contract:

- `SessionInferencePayload.primary_work_kind`
- `SessionInferencePayload.decision_signal_strength`
- `SessionInferencePayload.decisions`
- `SessionPhaseInferencePayload.kind`
- `SessionEnrichmentPayload.refined_work_kind`
- `SessionProfile.decisions`
- the root-level `Decision` export

The implementation change spans `polylogue/archive_products.py`, `polylogue/lib/phase_extraction.py`, `polylogue/lib/session_profile_models.py`, `polylogue/lib/session_profile_runtime.py`, `polylogue/storage/session_product_profiles.py`, and the row mappers. The key structural change is that phases are time-gap segments with tool counts and confidence, not intent labels, and profiles expose durable evidence/support signals rather than a supposedly canonical work kind.

Relaxed archive-product model hydration to `extra="ignore"` so old materialized rows with the retired fields can still be read until the next refresh.

### Clean Claude Code protocol noise out of human-facing text

Claude Code sessions were still leaking XML-like protocol artifacts into work-event summaries and generated titles. I added `_clean_summary_text()` in `polylogue/lib/work_event_extraction.py` and `_clean_title_text()` in `polylogue/sources/parsers/claude_code_parser.py` so tags like `<system-reminder>` and `<command-name>` are stripped before we surface human-readable text.

This is a small change, but it matters because session-product cleanup is not only about dropping bad fields — it is also about making the remaining fields cleaner.

## Testing

I updated the tests that asserted the old product shape:

- `tests/unit/cli/test_products.py`
- `tests/unit/core/test_facade_api.py`
- `tests/unit/mcp/test_tool_contracts.py`
- `tests/unit/core/test_version.py`

Those changes make the suite assert confidence/support and the remaining payload structure rather than the removed heuristic fields.

## Notes

The feature branch includes some broader performance experiments around preview memory and acquisition flushing. I left those out of the final narrative because the changes that survived and hang together cleanly are narrower: SQLite tuning, trigger lifecycle, contract cleanup, and Claude Code text cleanup. I would rather keep the final commit honest about that scope than pretend every experiment in the branch belongs in the same story.


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 2 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 3 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #134, opened 2026-04-07
perf: session product refresh and acquisition optimization

Sinity:
## Summary

I moved session-product maintenance and acquisition skipping onto incremental paths so ingest work scales with the changed conversations rather than with the size of the whole archive. The two core changes are replacing the post-batch full session-product rebuild with a per-conversation refresh path and making the mtime skip logic cheap enough to use reliably, including for ZIP-backed sources. Around that core I added caching for project normalization, folded the refresh helpers into a single module, added profiling tools to the dev shell, and stabilized a few tests and empty-file edge cases that showed up while exercising the new paths. The result is that parse runs do less unnecessary work, especially when only a small subset of raw inputs changed. The branch also records some developer workflow documentation and hook guards so the performance work is easier to reproduce without relying on tribal knowledge.

## Motivation

The old refresh model had the right data but the wrong cost profile. After a parse batch finished, the code could still end up doing work that was closer to a whole-product rebuild than to a delta update. That is manageable on a small corpus, but it becomes the wrong asymptotic behavior once the archive is large and the common case is “a handful of conversations changed”.

The acquisition side had a similar problem in a different form. The system already tracked source mtimes, but the skip query could still be expensive and ZIP-backed sources were not using the optimization effectively because the stored keys were entry-scoped (`path.zip:entry.json`) while the incoming candidate was the parent ZIP path. That meant a source that should have been skipped could still cost time in avoidable I/O.

I wanted this branch to make the “incremental” story real in both places: delta refresh for session products, cheap mtime lookups for raw sources, and enough tooling/docs to profile the path when it regresses.

### Refresh only the conversations that changed

In `polylogue/pipeline/services/parsing_batch.py` I replaced the old post-batch rebuild path with `refresh_session_products_for_conversation_async()` on the set of conversations whose content actually changed. That moves the cost model from “rewrite the durable product surface after a batch” to “repair the specific profiles, phases, threads, and aggregates touched by this batch”.

The key shape is:

```python
for cid in changed_conversation_ids:
    await refresh_session_products_for_conversation_async(
        conn,
        cid,
        transaction_depth=1,
    )
```

That is the important behavioral shift of the branch. It also let me simplify `polylogue/storage/repository_write_conversations.py`, because the save path can assume the caller will handle product refresh at the batch boundary instead of trying to interleave it per conversation.

### Put the refresh contract in one module

Once the incremental path was working, the split between `session_product_refresh_deletes.py`, `session_product_refresh_updates.py`, and later the separate batch helpers stopped buying much. I merged that logic into `polylogue/storage/session_product_refresh.py` so delete-side and update-side maintenance live under one module with one vocabulary. The code reads more like “session product refresh” and less like an implementation trail of how the work was gradually decomposed.

This is especially useful for reviewers because the refresh surface is easy to audit in one file.

### Make mtime skips cheap and correct

I added a covering index on `(source_path, file_mtime)` in `polylogue/storage/backends/schema_ddl_archive.py` and `schema_upgrade.py`:

```sql
CREATE INDEX IF NOT EXISTS idx_raw_conv_source_mtime
ON raw_conversations(source_path, file_mtime)
WHERE file_mtime IS NOT NULL
```

The comment in the upgrade path is the real point: this avoids scanning a very large `raw_conversations` table just to answer a skip question.

Then in `polylogue/sources/cursor.py` I fixed the ZIP case so an incoming `path.zip` candidate can match already-indexed `path.zip:entry.json` rows by prefix. That closes the gap where ZIP sources were still doing unnecessary work even when their parent file mtime had not changed.

Added `lru_cache()` to `normalize_project_name()` and `normalize_repo_path()` in `polylogue/lib/project_normalization.py`, because the session-product path revisits the same project tokens often enough that the repeated normalization work showed up during profiling.

### Tooling and guardrails for performance work

The branch adds `py-spy` and `pyinstrument` to the Nix dev shell and checks in `.claude` workflow/architecture/testing docs plus the local hook settings that block direct pushes and force pushes. I do not consider those changes the headline, but they matter because this kind of performance work is much harder to maintain if the reproduction instructions live only in someone's shell history.

### Stability fixes discovered while exercising the path

While pushing on the incremental route, I picked up two small but worthwhile fixes:

- `polylogue/pipeline/services/acquisition_streams.py` skips empty files instead of emitting noisy warnings for them
- timestamp and MCP-related ordering flakes were cleaned up in the test suite so the branch does not trade speed for brittleness

## Testing

The branch touches both performance-oriented code and the supporting test/tooling surface. Relevant coverage updates include:

- `tests/benchmarks/conftest.py` for more realistic benchmark fixtures
- `tests/unit/core/test_timestamp_guards.py` for invalid-input stability
- `tests/unit/mcp/conftest.py` and `tests/infra/pty_cli.py` for ordering-sensitive test behavior
- `tests/unit/pipeline/test_resilience.py` for pipeline-path expectations touched by the refresh changes

## Notes

The feature branch contains some workflow and documentation changes because I wanted the profiling environment to be reproducible and guarded while I was pushing on the hot path. The core product/runtime story is still performance: incremental session-product refresh, cheaper acquisition skipping, and small supporting fixes that make those paths dependable.

## Release Notes

* **Documentation**
  * Added comprehensive project architecture documentation
  * Added Git workflow guidelines and branching conventions
  * Added testing procedures and benchmark guides

* **Improvements**
  * Optimized database performance with enhanced SQLite configuration and targeted indexing
  * Added caching for schema resolution and normalization operations
  * Implemented FTS trigger management for efficient bulk operations
  * Enhanced data cleaning to remove embedded protocol artifacts
  * Improved memory efficiency in pipeline operations

* **Bug Fixes**
  * Fixed PTY output handling to capture all buffered data
  * Corrected session phase ID generation

* **Tests**
  * Upgraded benchmark fixtures with realistic data distributions
  * Expanded test coverage for date parsing validation


Sinity:
@coderabbitai full review

Sinity:
@codex review

PR #136, opened 2026-04-07
feat: content-addressed blob store and pipeline instrumentation

Sinity:
## Summary

I replaced raw payload BLOB storage in the database with a content-addressed filesystem blob store and then rewired the raw-content read paths so they can operate on blob paths without assuming bytes are sitting in SQLite. The structural change is complemented by a second theme: the pipeline emits enough timing and slow-path information that we can see where ingest time is going once the storage contract changes. Concretely, `raw_conversations` stores metadata (`raw_id`, `blob_size`, provenance) while the bytes live under a SHA-256-addressed blob root, acquisition/Drive ingestion write into that store, JSONL decode and schema sampling can stream from `Path`, and the CLI gained `check --blob` plus `reset --blob`. Replaced the old cached health-summary stub with a live lightweight status line and added progress/timing instrumentation around ingest, parse, and save paths. One early idea in the branch — deferring artifact inspection during acquisition — was tried and then reverted once it became clear that the observation path still needed to stay inline.

## Motivation

The old storage model treated the database as the home of raw payload bytes. That is simple, but it scales badly once the corpus contains large JSON and JSONL payloads. It also pushes several unrelated concerns together: persistence, deduplication, integrity, and read-path memory behavior all get coupled to a single BLOB column.

I wanted a storage contract with better physical properties:

- identical raw payloads should deduplicate automatically
- large files should be written and hashed without loading the whole thing into Python memory
- JSONL parsing and schema sampling should be able to stream from disk
- the database should track metadata about raw inputs without being the byte store itself

Once I started pushing that through the pipeline, it was obvious I also needed better observability. Swapping storage contracts without per-stage timings, slow-item logs, and progress descriptions would leave us guessing about regressions. That is why the branch intentionally combines the blob store with pipeline instrumentation.

## Motivation and branch pivots

There is one important false start visible in the branch: I briefly tried to defer artifact inspection out of acquisition to reduce memory pressure, then reverted it. That was useful to prove the boundary, but the end state keeps artifact observation inline and solves the bigger memory problem by moving raw bytes into the blob store and streaming read paths. I am calling this out because it explains why the final implementation is not “just postpone work until later”.

### Add a content-addressed blob store

The new foundational module is `polylogue/storage/blob_store.py`. It introduces a simple but useful contract:

```python
class BlobStore:
    def write_from_path(self, source: Path) -> tuple[str, int]: ...
    def write_from_bytes(self, data: bytes) -> tuple[str, int]: ...
    def read_prefix(self, hash_hex: str, n: int = 65536) -> bytes: ...
    def read_all(self, hash_hex: str) -> bytes: ...
    def verify(self, hash_hex: str) -> bool: ...
```

Blobs are immutable files stored under `{root}/{hash[:2]}/{hash[2:]}` and addressed by the same SHA-256 digest we already use as `raw_id`. That means deduplication is essentially free and we do not need a second identifier system for the filesystem layer.

The write methods are atomic (`tempfile` + `os.replace`) and the file-path variant hashes in chunks, which means a large input can be ingested without holding the whole payload in memory.

### Move raw archive rows to metadata-only records

The database side of the change removes `raw_conversations.raw_content` and replaces it with `blob_size`. The related model and query changes show up in:

- `polylogue/storage/store_runtime_raw_records.py`
- `polylogue/storage/backends/schema_ddl_archive.py`
- `polylogue/storage/backends/queries/raw_writes.py`
- `polylogue/storage/backends/queries/mappers_archive.py`
- `polylogue/sources/parsers/base_models.py`

`RawConversationRecord` tracks `blob_size` instead of bytes, while `RawConversationData` can carry `blob_hash` and `blob_size` with empty `raw_bytes`. That is the crucial contract shift: bytes are optional in flight and absent in the DB row; the blob store is the source of truth for payload content.

### Rewire acquisition, parsing, and schema reads around blob paths

`polylogue/sources/source_acquisition.py` now streams non-ZIP files into the blob store with `write_from_path()` and only reads a small prefix back for provider detection. ZIP entries are still read into memory, but they are written to the blob store immediately and handed downstream by hash/size rather than by large byte payloads.

`polylogue/sources/drive.py` does the same for Drive-backed inputs and later learns to check a local drive-cache file before calling the API again. That gives us a nice two-layer behavior: cached file if present, otherwise download, write to the blob store, then cache for future runs.

On the read side, `polylogue/lib/raw_payload_decode.py` accepts `Path` inputs and streams JSONL directly from disk. That same path-oriented contract gets picked up by parsing, validation, schema sampling, and verification helpers.

### Add CLI management and live health for the new store

Once raw bytes move out of SQLite, operators need first-class ways to inspect and clean that state. I added:

- `reset --blob` in `polylogue/cli/commands/reset.py`
- `check --blob` in `polylogue/cli/check_workflow.py` / `commands/check.py`

The blob check compares DB `raw_id`s against on-disk hashes and reports missing/orphaned blobs. Replaced the `cached_health_summary()` stub in `polylogue/health.py` with a lightweight live read of schema version and conversation count so the non-verbose CLI can surface something real.

### Instrument the pipeline enough to debug the storage swap

The second half of the branch adds observability:

- per-stage ingest/parse timings in `parsing_workflow.py`
- `StageMetrics` sub-timings wired through `run_execution.py`
- slow-item / slow-batch / slow-refresh logging in `parsing_batch.py`
- per-step save timings in `repository_write_conversations.py`
- progress description updates after each parse batch
- live health summary instead of a dead cache stub

That gives us usable traces when the new storage path regresses.

## Testing

I had to touch a broad set of tests because the storage contract changed:

- fixtures in `tests/conftest.py` and `tests/infra/corpus_fixtures.py`
- schema/sampling/verification coverage in `tests/unit/core/test_sampling.py`, `test_schema_validation.py`, and `test_verification.py`
- ingestion/resilience coverage in `tests/unit/pipeline/test_prepare_records.py`, `test_resilience.py`, and `test_run_sources.py`
- source-law and encoding coverage in `tests/unit/sources/test_source_laws.py` and `test_acquisition_encoding.py`
- CLI snapshots for the new `--blob` flag

## Notes

The schema change deliberately keeps the archive on the project’s fresh-only path rather than inventing a migration chain just for raw blob storage. That keeps the implementation simpler, but it means the expectation is rebuild/reseed not in-place column migration. The branch also removes the old raw-size and attachment-size assumptions that were only defensible when bytes lived in the database.


Sinity:
@coderabbitai full review

Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 11 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550). 2 finding(s) confirmed fixed/stale.

Sinity:
@codex review

PR #138, opened 2026-04-07
perf: orjson hashing and FTS trigger ordering fixes

Sinity:
## Summary

I cleaned up the last few obvious hot-path inefficiencies that showed up after the blob-store ingest work: payload hashing was still paying the cost of stdlib JSON serialization, the save path was copying messages for no durable reason, and the FTS trigger suspension window was narrower than intended. The result is a small diff, but it lands on code that runs constantly. `hash_payload()` uses `orjson` with sorted keys, `save_via_backend()` no longer rebuilds the message list just to inject a provider name during stats upserts, and `process_raw_batch()` suspends FTS triggers in the right place and for the full duration of the batch. The second and third commits in the branch are both about the same trigger-ordering bug: the first fix moved suspension past schema setup, and the follow-up widened the scope so session refresh also runs with triggers disabled.

## Motivation

After the previous ingest work, the remaining regressions were no longer broad architectural problems; they were specific hot spots. Two of them were easy to explain from the code.

First, `polylogue/lib/hashing.py` was still doing this:

```python
serialized = json.dumps(payload, sort_keys=True, separators=(",", ":"))
return hashlib.sha256(serialized.encode("utf-8")).hexdigest()
```

That is deterministic, but it allocates an intermediate Python string on every call. We already depend on `orjson`, so keeping the same stable ordering with `OPT_SORT_KEYS` is a cheaper way to serialize.

Second, the FTS trigger suspension introduced earlier was correct in spirit but wrong in ordering. Opening a connection can run schema setup, and schema setup recreates the triggers with `CREATE TRIGGER IF NOT EXISTS`. Suspending them too early means they quietly come back before the batch is done. The first fix in the branch corrected the placement relative to `bulk_connection()`, and the second fix corrected the duration so session-product refresh does not reintroduce per-row trigger overhead after the save loop.

### Deterministic hashing without the stdlib JSON hop

In `polylogue/lib/hashing.py` I replaced the stdlib serializer with `orjson`:

```python
serialized = orjson.dumps(payload, option=orjson.OPT_SORT_KEYS)
return hashlib.sha256(serialized).hexdigest()
```

The important property here is not “use a faster library” in the abstract; it is that we preserve deterministic ordering while dropping the intermediate string encode step. The hash contract stays the same shape — stable SHA-256 over a sorted JSON representation — but the implementation is better suited to a hot path.

### Remove dead message copying in the save path

In `polylogue/storage/repository_write_conversations.py` there was still a branch that rebuilt the `messages` list with `model_copy()` just to ensure a provider name was present before stats aggregation. That copy was not buying us anything durable in the persisted message records. I removed the copy step and now pass `conversation.provider_name` directly into `upsert_conversation_stats()`.

Before:

```python
messages = [
    message.model_copy(update={"provider_name": pname})
    if not message.provider_name else message
    for message in messages
]
```

After, the save path leaves `messages` alone and only passes the conversation-level provider where the stats query needs it.

### Fix trigger suspension ordering and scope

The branch’s second and third commits are really one bug investigation. The first moved trigger suspension to the point after `bulk_connection()` had opened, because that opening step ensures schema and could recreate triggers. The follow-up then corrected the scope again so the suspension covers the entire batch — save plus session-product refresh — rather than only the inner write loop.

The durable shape is:

```python
async with backend.connection() as fts_conn:
    await suspend_fts_triggers_async(fts_conn)
    await fts_conn.commit()

async with backend.bulk_connection():
    ...
```

That ordering keeps the schema-init side effect from undoing the suspension, and it keeps the expensive per-row FTS work off for the whole batch, not just the first half of it.

## Testing

There are no new dedicated tests in this diff, so the validation burden is on the existing parse-batch and save-path suites. The key behaviors being protected are:

- deterministic `hash_payload()` output with sorted keys
- unchanged conversation/message persistence behavior in `save_via_backend()`
- successful parse batches and session refresh with FTS triggers restored after the batch

## Notes

This is intentionally a narrow follow-up branch. It does not introduce new interfaces or new telemetry; it just makes the existing interfaces cheaper and corrects the timing bug in the trigger lifecycle. The important reviewer question is whether the trigger suspension still composes correctly with schema setup and later FTS rebuilds, not whether the code is “clever”.

### DIFF
--- a/polylogue/lib/hashing.py
+++ b/polylogue/lib/hashing.py
@@ -7,9 +7,10 @@ implementations scattered across the codebase.
 from __future__ import annotations
 
 import hashlib
-import json
 import unicodedata
 from pathlib import Path
+
+import orjson
 
 
 def hash_text(text: str) -> str:
@@ -39,8 +40,8 @@ def hash_payload(payload: object) -> str:
     Callers should normalize strings before including in payload if
     normalization-invariant hashing is required.
     """
-    serialized = json.dumps(payload, sort_keys=True, separators=(",", ":"))
-    return hashlib.sha256(serialized.encode("utf-8")).hexdigest()
+    serialized = orjson.dumps(payload, option=orjson.OPT_SORT_KEYS)
+    return hashlib.sha256(serialized).hexdigest()
 
 
 def hash_file(path: Path) -> str:
--- a/polylogue/pipeline/services/parsing_batch.py
+++ b/polylogue/pipeline/services/parsing_batch.py
@@ -154,6 +154,15 @@ async def process_raw_batch(
     _blob_store = _get_blob_store()
 
     async with backend.bulk_connection():
+        # Suspend FTS triggers AFTER bulk_connection opens (which ensures
+        # schema, including CREATE TRIGGER IF NOT EXISTS). Previous suspension
+        # in run_execution.py was undone by schema extension triggers.
+        from polylogue.storage.fts_lifecycle import suspend_fts_triggers_async
+
+        async with backend.connection() as fts_conn:
+            await suspend_fts_triggers_async(fts_conn)
+            await fts_conn.commit()
+
         for convo_item, source_name_item, raw_id in work_items:
             t_item = time.perf_counter()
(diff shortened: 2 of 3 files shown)


Sinity:
@codex review

PR #140, opened 2026-04-07
perf: unify ingest around single-decode batch workers

Sinity:
## Summary

I rewired the hot path from "validate, then parse, then persist" into one ingest pipeline that decodes each raw blob once, performs validation and conversation extraction in the same subprocess, and writes normalized records through a single batch coordinator. The new path lives in `polylogue/pipeline/services/ingest_batch.py` and `polylogue/pipeline/services/ingest_worker.py`, with `parsing_workflow.py` reduced to scheduling and raw-state updates without reimplementing batch mechanics. The key practical change is that parse work is shaped like a producer-consumer system: workers do CPU-bound normalization, while the main process performs ordered SQLite writes and raw-state bookkeeping. That keeps the previous performance goal of a single decode pass, but avoids the earlier regression-prone attempt to collapse everything straight into synchronous write code. Cleaned up the public stage contract so `parse` is the validation+ingest stage, removed the leftover `_parse_raw_record()` compatibility hook, and fixed a couple of edge cases that surfaced while the worker path was being hardened.

## Motivation

The branch started with timing instrumentation in `validation_flow.py` / `validation_runtime.py` to confirm where the time was going. That showed two problems clearly: we were decoding the same raw payload twice across validation and parse, and the original thread-based validation path was not buying us anything for GIL-heavy JSON work. I tried a more direct "merge validate+parse and write everything synchronously" path, but that shape regressed enough behavior that I explicitly backed it out before landing the final version.

The version here keeps the architectural insight from that experiment — single decode, worker-side normalization, fewer async crossings — but rebuilds it in a safer form. without teaching every parsing surface about the merged flow, I put the new contract behind `process_ingest_batch()` and had the rest of the pipeline call into it. That made it much easier to preserve existing raw-state semantics, keep parent/child ordering constraints in one place, and add a test-friendly fallback when `ProcessPoolExecutor` hits unpicklable objects.

### 1. Unify decode, validation, parse, and transform

The center of the change is the new worker contract:

```python
# polylogue/pipeline/services/ingest_worker.py
def ingest_record(raw_record, archive_root_str, validation_mode):
    envelope = build_raw_payload_envelope(...)
    validator = SchemaValidator.for_payload(...)
    conversations = parse_payload(...)
    return IngestRecordResult(...)
```

Before this, validation and parse had separate entry points and both needed to decode the blob. Now the worker returns the validation outcome and DB-ready conversation bundles from the same decoded payload. The main process no longer needs to reconstruct parsing context from scratch after validation has already run.

Kept timestamp fallback logic and provider/payload-provider normalization inside the worker as well. That matters because it means the main-thread batch writer can stay narrowly focused on ordering, persistence, and counters.

### 2. Replace split batch orchestration with a producer-consumer coordinator

`polylogue/pipeline/services/ingest_batch.py` is new and intentionally central. It owns:

- ProcessPool submission
- collection of `IngestRecordResult` values as workers finish
- sync `sqlite3` writes in dependency-safe order
- raw-state and validation-state persistence
- batch observations / timing details

The structure is explicitly producer-consumer instead of "finish all compute, then do all writes":

```python
with ProcessPoolExecutor(...) as executor:
    futures = [...]
    for future in as_completed(futures):
        ir = future.result()
        _drain_ready_conversation_entries(...)
```

That lets CPU work continue while the main thread writes completed conversations. It also means we can keep parent/child conversation ordering localized to the batch writer instead of leaking it back into parsing surfaces.

### 3. Shrink the parsing surface back down

With the batch work centralized, `parsing.py` and `parsing_workflow.py` got simpler rather than more complicated:

- `_parse_raw_record()` is deleted
- `skip_validate` disappears from `ParsingService.ingest_sources()`
- `parse_from_raw()` delegates to `process_ingest_batch()` instead of stitching together parse helpers, validation state, and deferred refresh plumbing itself

That deletion mattered to me almost as much as the new performance path. The old compatibility wrapper would have forced every future change to preserve two parsing contracts. Removing it makes the new ingest path the only one the codebase has to reason about.

### 4. Make the stage contract match the new architecture

`run_support.py` and `run_stages.py` stop pretending there is a separate steady-state validate stage. `parse` is the stage that validates and ingests in one pass, and `execute_ingest_stage()` no longer needs a `skip_validate` toggle to punch holes through the worker contract.

That is a real semantic cleanup, not just help-text drift: the execution layer now lines up with how the pipeline actually works.

### 5. Harden edge cases exposed by the worker move

Two small fixes fell out of the transition:

- `validation_flow.py` falls back to sequential execution when ProcessPool cannot pickle a test object, which keeps the worker path usable in unit tests without weakening the production path.
- `polylogue/lib/hashing.py::hash_payload()` falls back to stdlib `json.dumps()` when `orjson` rejects a payload, so large integers still hash deterministically.

Neither is the headline feature, but both prevent the new path from becoming flaky in exactly the places where it is easiest to lose trust.

## Testing

I updated the tests that actually exercise the moved surfaces instead of just snapshotting help text:

- `tests/unit/pipeline/test_parsing_service.py` for parse orchestration and diagnostics
- `tests/unit/pipeline/test_resilience.py` for ingest failure paths and state updates
- `tests/unit/pipeline/test_ingestion_chaos.py` for more adversarial batch/worker behavior
- `tests/unit/pipeline/test_run_sources.py` for end-to-end parse stage expectations
- `tests/unit/cli/test_run_int.py` for run artifact behavior through the CLI path

The new sequential fallback in validation is specifically there to keep mock-heavy tests green while production still uses process workers.

## Notes

I deliberately did **not** keep the earlier direct-sync-write experiment. The final batch writer still uses synchronous `sqlite3`, but only behind the new coordinator that understands ordering and partial failure semantics. That is the compromise that preserved the performance goal without leaving the pipeline in a half-merged state.

The other intentional line I drew here is that session-product refresh stays a downstream concern. This PR unifies decode/validate/parse/transform/write; it does not try to solve every later-stage optimization at the same time.

## Release Notes

* **Chores**
  * Removed the standalone `validate` pipeline stage. Schema validation is now performed inline during the `parse` stage as part of a unified workflow.
  * Consolidated the data ingestion pipeline to streamline conversation record processing. The number of distinct pipeline stages has been reduced by merging validation directly into the parsing process.


Sinity:
@coderabbitai full review

Sinity:
@coderabbitai full review

Sinity:

Sinity:
@codex review

PR #142, opened 2026-04-07
perf: stream acquisition, batch persistence, and pipeline metrics

Sinity:
## Summary

I used this branch to finish the work around the unified ingest path rather than inside it. The code now streams more acquisition input directly into the blob store, writes more ingest-side state in batches, and persists enough run/probe metadata that performance work can be reproduced instead of eyeballed from logs. The centerpieces are a new `devtools/pipeline_probe.py`, richer `PipelineMetrics` / run payload persistence, and a heavily instrumented `ingest_batch.py` that can explain where a batch spent time, memory, and write effort. Cleaned up the public operator surface so the CLI/help/product queries reflect the inline-validation pipeline we already have, not keeping deprecated knobs alive. The end result is not one architectural rewrite; it is a set of boundary fixes that make the existing ingest path measurably cheaper and much easier to reason about.

## Motivation

After the previous ingest consolidation, the remaining bottlenecks were no longer "decode twice" style problems. They were boundary costs: reading whole zip entries into memory before hashing them, waiting for an entire batch of worker results before beginning writes, issuing too many tiny raw-state and session-product refresh updates, and lacking a reproducible way to measure any of it.

I intentionally approached this in that order. First I made the public surfaces honest: no separate `validate` stage, no stale `refined_work_kind` product filter, and a returned run object that points to the JSON artifact written under `archive_root/runs/`. Then I added a synthetic probe so the performance work had a bounded harness. Once that existed, it was simple to widen the probe to archive-subset replay and staged real-source inputs, which in turn made it much easier to justify the more surgical batching changes in acquisition and ingest.

### 1. Make the pipeline observable enough to tune

`devtools/pipeline_probe.py` starts as a synthetic smoke test, but the more important direction is that it becomes a reusable performance contract. It can:

- generate a bounded synthetic corpus
- replay a persisted archive subset through an isolated workspace
- stage real source paths into a disposable workspace
- fail against explicit runtime / peak RSS budgets
- persist manifests and provenance so results are replayable

The payload it emits is grounded in the same run artifact the main pipeline writes:

```python
return {
    "result": result.model_dump(),
    "run_payload": json.loads(Path(result.run_path).read_text(...)),
    "db_stats": _db_row_counts(db_path),
    "raw_fanout": _db_raw_fanout(db_path),
}
```

On the runtime side, `polylogue/lib/metrics.py` now captures stage timings plus current/peak RSS, and `polylogue/pipeline/run_finalization.py` writes the assembled run JSON to `archive_root/runs/run-*.json`. Carried `run_path` through `RunResult` so callers do not have to rediscover the artifact after the fact.

### 2. Clean up the public contract around inline validation

Commit-for-commit, this is smaller than the measurement work, but it matters because it removes ambiguity:

- `PlanStage.VALIDATE` is dropped
- CLI/help/docs no longer advertise `polylogue run --stage validate`
- `SessionEnrichmentProductQuery.refined_work_kind` and the related product filter are removed
- shell completion for source names is wired into `run` and `qa`

The effect is that the operator surface finally matches the pipeline contract: parse is the stage that consumes raw backlog and performs validation inline.

### 3. Stream more acquisition input directly into the blob store

A lot of acquisition work was still buffering raw bytes unnecessarily. This branch tightens that in a few places:

- grouped zip/plain capture now streams directly into blob storage
- non-grouped zip bundles are split during acquisition instead of preserved as larger composite raws
- stable provider hints and JSONL sniff results get reused not recomputed
- source paths are preserved on ingest errors so diagnostics stay tied to the original file

None of these are glamorous changes, but they address exactly the kind of hidden overhead that only shows up when archives get large.

### 4. Batch the persistence work that happens after workers return

`polylogue/pipeline/services/ingest_batch.py` got most of the mechanical performance work in this branch. The important shift is that it no longer waits for every worker before it begins writing. Instead, it records outcomes and drains ready conversations as worker results arrive.

That unlocks a few follow-on improvements:

- raw-state updates can be merged and written after the batch, not one-by-one mid-flight
- session-product refresh work can be deduped, chunked, and timed as a separate concern
- observations can report setup, wait, drain, write, flush, commit, and refresh breakdowns not one opaque batch time
- result payload sizes can be measured so the worker/main-process boundary remains visible

The batching theme continues in `polylogue/storage/session_product_refresh.py`, which now dedupes refresh inputs, batches thread-root lookups, and writes refresh results by chunk.

### 5. Preserve correctness around the new batching behavior

A few smaller fixes came along because the new probe/metrics surfaces made them obvious:

- action-event hydration preserves message hashes instead of quietly destabilizing downstream derivations
- parent conversation links are resolved during ingest
- Linux peak RSS reads from procfs when available, not relying solely on a coarser fallback
- parse backlog collection dedupes incrementally instead of rebuilding the same sets repeatedly
- phase extraction gets reused in session analysis rather than recomputed

These are good examples of why I wanted the probe work first: once timing and artifact output were easy to inspect, the correctness gaps stood out more clearly.

## Testing

The branch adds and expands tests around the exact surfaces that moved:

- `tests/unit/devtools/test_pipeline_probe.py` for probe modes, manifests, budgets, and persisted summaries
- `tests/unit/pipeline/test_ingest_batch.py` for write-as-completed batching and observation accounting
- `tests/unit/sources/test_source_laws.py` for split/stream acquisition behavior
- `tests/unit/storage/test_session_product_refresh.py` for dedupe/chunk refresh behavior
- `tests/unit/pipeline/test_run_sources.py` and `tests/unit/cli/test_run_int.py` for persisted run JSON and run-path expectations
- CLI tests for source completion helpers and the updated stage/help surfaces

## Notes

I did not try to turn the probe into a production control plane. It is still a development tool, but now one that speaks the same artifact language as the real pipeline. That keeps the measurement surface honest.

The other line I kept is that these optimizations stay mostly at the boundaries. The core ingest architecture from the previous work remains recognizable; this PR is about removing buffering and bookkeeping overhead around it, not introducing another central rewrite.

## Summary by CodeRabbit

* **New Features**
  * New pipeline probe CLI for fast, bounded pipeline testing (synthetic, archive-subset, source-subset) with JSON summaries and optional resource budgets.
  * CLI lanes for quick probes and showcase baselines.

* **Documentation**
  * Updated pipeline docs to reflect validation merged into parse.
  * Added guidance for fast pipeline probes and updated CLI examples/help.

* **Improvements**
  * Shell completion for source names.
  * Richer run reports (JSON artifacts) and improved memory/resource metrics.
  * Bulk/session product performance and reliability enhancements.


Sinity:
@coderabbitai full review

Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 16 finding(s) confirmed still present → tracked in [#546](https://github.com/Sinity/polylogue/issues/546), [#550](https://github.com/Sinity/polylogue/issues/550).

Sinity:
@codex review

PR #144, opened 2026-04-07
perf: reuse hot-path scans across ingest and enrichment

Sinity:
## Summary

This branch is a deliberately incremental performance pass. without changing the ingest architecture again, I looked for places where we were walking the same data twice, recomputing the same derived strings, or allocating large temporary buffers for work that could be streamed or cached locally. The changes land in five clusters: enrichment text-band reuse, timeline/aggregate row reuse, acquisition-to-blob streaming, message/render fast paths, and better timing/progress accounting around ingest/index work. Most of the files are small or medium changes on their own, but together they remove a lot of repeated hot-path work without changing the user-facing data model. Kept the tests close to each optimization so we can tell whether a future cleanup accidentally reintroduces the old repeated scans.

## Motivation

The previous branch gave us better instrumentation and better batching. Once that existed, it became much easier to see a different class of waste: work that was "correct but repeated." Session enrichment was walking the same `message_facts` list multiple times to produce related text bands. Timeline row builders were recomputing the same summaries and signals inside neighboring helper functions. Aggregate refresh paths would hydrate more profile rows than a given root/day actually needed. Acquisition still buffered zip entries before hashing them into blob storage. Rendering and runtime helpers did more regex/Markdown work than necessary for plain text or already-known predicates.

None of these justified a large rewrite, but all of them are exactly the kind of costs that show up once the bigger structural bottlenecks are gone.

### 1. Reuse enrichment text bands instead of rescanning message facts

`polylogue/storage/session_product_profiles.py` has a single `_collect_enrichment_text_bands()` helper that gathers:

- user turns
- assistant turns
- blocker candidates

from one pass over `analysis.facts.message_facts`. The existing public helpers then become thin readers over that collected structure, and `session_enrichment_payload()` / `enrichment_support_signals()` reuse the same bands instead of each constructing their own filtered list.

That is a small local change, but it matters because enrichment payload creation sits on the session-product path, where repeated scans add up quickly.

### 2. Build timeline and aggregate rows from local reusable summaries

The same pattern shows up in `polylogue/storage/session_product_timeline_rows.py`. Event and phase builders used to call helper functions that re-derived summary text, support signals, and fallback flags multiple times while assembling a single output row. I split those into helpers that accept precomputed values and then rewired row construction to compute them once per event/phase.

The aggregate refresh path also got more local:

- `session_product_threads.py` resolves each requested root from its own profile-record slice
- `session_product_aggregates.py` rebuilds provider-day rows from the profile records for that provider/day instead of hydrating a larger shared set and regrouping afterward

That reduces incidental work and makes the refresh code easier to review because each helper is explicitly local to a root or provider-day group.

### 3. Stream zip entries directly into the blob store

`BlobStore.write_from_fileobj()` is probably the most user-visible optimization in this branch because it changes the memory profile of acquisition. Before, zip-entry acquisition paths would do something like this:

```python
with zf.open(info.filename) as handle:
    raw_bytes = handle.read()
blob_hash, blob_size = blob_store.write_from_bytes(raw_bytes)
```

They now hash and write in one pass:

```python
with zf.open(info.filename) as handle:
    blob_hash, blob_size = blob_store.write_from_fileobj(handle)
```

I applied that to grouped zip capture and preserved zip-entry capture in `polylogue/sources/source_acquisition.py`, then tightened nearby JSONL sniffing so we do less repeated provider work when walking those archives.

### 4. Add local fast paths where the full general case is unnecessary

Two examples here:

- `polylogue/lib/message_model_runtime.py` now short-circuits more predicate work and guards rarer regex scans behind cheaper checks
- `polylogue/rendering/renderers/html_highlighting.py` adds an LRU-cached plain-text fast path for small messages that do not need full Markdown parsing

The HTML renderer still falls back to the MarkdownIt path whenever the text looks remotely non-plain. The fast path is intentionally conservative, and the tests assert equivalence for the cases it accepts.

### 5. Make timing and progress accounting more honest

Used the branch to close a couple of measurement gaps that the previous instrumentation exposed:

- `ingest_batch.py` separates setup and teardown timing from the rest of batch elapsed time and includes them in the unattributed residual calculation
- index rebuild/update can report real `processed/total` fractions through `IndexService`
- `RichProgressObserver` can parse those fractions and update totals instead of only advancing an indeterminate bar

This is not just cosmetic. Progress output is a debugging surface when indexing gets slow, and misattributed batch residual time makes the rest of the timing data harder to trust.

## Testing

The coverage is intentionally local to each optimization:

- `tests/unit/storage/test_session_product_profiles.py` for shared enrichment text bands
- `tests/unit/storage/test_session_product_timeline_rows.py` for reused timeline summaries/signals
- `tests/unit/storage/test_blob_store.py` and `tests/unit/sources/test_source_laws.py` for `write_from_fileobj()` and streamed zip acquisition
- `tests/unit/pipeline/test_ingest_batch.py` for setup/teardown timing accounting
- `tests/unit/rendering/test_html_highlighting.py` for the plain-text fast path and cache behavior
- `tests/unit/pipeline/test_indexing.py` and `tests/unit/cli/test_run_laws.py` for indexed progress fractions

## Notes

Intentionally kept this as a "many small cuts" branch. A few of the helpers have names that start with an underscore because the point is to localize reuse inside a module, not to create new broad abstractions prematurely.

The main thing to watch here is behavioral equivalence. Almost every change is motivated by "do the same work fewer times," so the most important review question is whether any shortcut accidentally changes payload shape, search text, or render output.

### DIFF
--- a/polylogue/cli/run_observers.py
+++ b/polylogue/cli/run_observers.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import re
 import time
 from collections.abc import Iterator
 from contextlib import contextmanager
@@ -10,6 +11,8 @@ from polylogue.cli.formatting import format_counts
 from polylogue.cli.types import AppEnv
 from polylogue.pipeline.observers import RunObserver
 from polylogue.storage.state_views import RunResult
+
+_PROGRESS_FRACTION_RE = re.compile(r"(?P<completed>\d[\d,]*)/(?P<total>\d[\d,]*)")
 
 
 def _format_elapsed(seconds: float) -> str:
@@ -87,10 +90,43 @@ class RichProgressObserver(RunObserver):
         self._progress = progress
         self._task_id = task_id
 
+    @staticmethod
+    def _progress_bounds(desc: str | None) -> tuple[int, int] | None:
+        if not desc:
+            return None
+        matches = list(_PROGRESS_FRACTION_RE.finditer(desc))
+        if not matches:
+            return None
+        match = matches[-1]
+        completed = int(match.group("completed").replace(",", ""))
+        total = int(match.group("total").replace(",", ""))
+        if total <= 0 or completed < 0 or completed > total:
+            return None
+        return completed, total
+
     def on_progress(self, amount: int, desc: str | None = None) -> None:
+        parsed_bounds = self._progress_bounds(desc)
+        if parsed_bounds is not None:
+            completed, total = parsed_bounds
+            if desc:
+                self._progress.update(
+                    self._task_id,
+                    description=desc,
+                    total=total,
+                    completed=completed,
+                )
+            else:
+                self._progress.update(
+                    self._task_id,
+                    total=total,
+                    completed=completed,
+                )
+            return
+
         if desc:
             self._progress.update(self._task_id, description=desc)
-        self._progress.update(self._task_id, advance=amount)
+        if amount:
+            self._progress.update(self._task_id, advance=amount)
 
 
 @contextmanager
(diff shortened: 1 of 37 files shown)


Sinity:
@codex review

PR #146, opened 2026-04-07
refactor: route run through chainable stage sequences

Sinity:
## Summary

I reworked `polylogue run` so the pipeline no longer thinks in terms of one opaque stage token. The code has an explicit leaf-stage sequence model (`acquire`, `parse`, `materialize`, `render`, `index`, `schema`, plus `reprocess` as a composite), and the CLI exposes that model through a chained Click group, not a flat `--stage` flag. This required changes in three places at once: planning/execution needed a real ordered sequence contract, the run CLI needed to collect and normalize stage requests, and the stage set itself needed to be cleaned up so `materialize` and `reprocess` are first-class rather than implicit side effects. Folded in the related operational work that depends on the same model: index-driven auto-embedding, explicit probe stage sequences, and provenance/fingerprint support for replaying those workflows. The result is a more explicit run surface with fewer hidden composites.

## Motivation

The old model had started to work against the code. `run_sources()` knew about composites that the CLI did not really expose. Planning could answer "how much parse work exists?" but not "what does a materialize-only or reprocess workflow mean?" Parse had to carry skip-acquire special cases without a clean representation of downstream stage order. And once I wanted the pipeline probe to replay exact workflows, a single enum-like stage value stopped being expressive enough.

I deliberately split the work into layers. First I made materialization explicit and fixed parse/reprocess semantics so the execution model could stand on its own. Then I formalized leaf-stage sequences and normalized them inside `run_support.py` / `run_execution.py`. Only after that did I flip the CLI to a chained Click group, because I wanted the runtime model to exist independently of the command surface. The index auto-embed and probe provenance pieces are here because they depend on the same explicit sequencing model.

### 1. Give the pipeline an ordered stage model

`polylogue/pipeline/run_support.py` now distinguishes between stage choices and the leaf stages they expand to. The important helpers are:

```python
RUN_STAGE_SEQUENCES = {
    "reprocess": ("parse", "materialize", "render", "index"),
    "all": ("acquire", "parse", "materialize", "render", "index"),
    ...
}

def normalize_stage_sequence(*, stage, stage_sequence=None) -> tuple[str, ...]:
    ...
```

`run_sources()` consumes that normalized sequence directly, so execution no longer has to guess at implied downstream steps. That also made it possible to add `materialize` cleanly as a stage in its own right.

### 2. Make materialize and reprocess real workflows instead of side effects

I added explicit materialization handling in `run_stages.py` and planning/runtime support in `planning_runtime.py`. That gives the codebase three useful guarantees:

- `materialize` can be planned and executed on its own
- `reprocess` has a defined ordered meaning (`parse -> materialize -> render -> index`)
- parse can stay "pure" and skip acquisition when that is genuinely what the caller requested

The stage vocabulary also got cleaned up by renaming the schema-generation stage from `generate-schemas` to `schema`, which makes both the CLI and the sequence tables easier to scan.

### 3. Expose the model through a chained run command

The CLI side now revolves around `RunStageRequest` objects and a chained `run` group:

```python
@click.group("run", chain=True, invoke_without_command=True)
class ...
```

Each subcommand contributes a `RunStageRequest`, and the result callback resolves:

- the canonical stage label
- the ordered leaf-stage sequence
- per-stage render options
- preview/watch display behavior

That makes composed invocations such as `polylogue run acquire parse` or `polylogue run render --format markdown` first-class rather than awkward special cases. When the user invokes `run` with no explicit stage in an interactive terminal, the prompt now selects from the stage workflows directly instead of inferring from the old global option.

### 4. Bring planning, probes, and indexing along with the new semantics

Once the ordered sequence model existed, a few related surfaces became much simpler:

- `run_sources()` can accept `stage_sequence=` directly for tests, probes, and composed workflows
- `devtools/pipeline_probe.py` can replay ordered stage sequences and attach provenance fingerprints to the staged inputs
- the validation lane catalog can point at a concrete live archive-subset parse probe
- `IndexService` can treat auto-embedding as part of the index workflow when config enables it

The auto-embed change is especially useful because it removes one more hidden "oh, and then also run embeddings" step from the operator model. If indexing owns search-surface preparation, it should own that step too.

### 5. Align docs and tests with the explicit stage model

The branch updates the help baselines, run snapshots, CLI tests, planning tests, probe tests, and docs because the user-visible surface genuinely changes here. I wanted those changes in the same branch because otherwise the new semantics would remain only half-described.

## Testing

The most important test coverage sits in:

- `tests/unit/pipeline/test_run_sources.py` for explicit sequences, materialize, reprocess, and stage matrices
- `tests/unit/pipeline/test_parsing_service.py` for planning and parse-stage semantics
- `tests/unit/pipeline/test_indexing.py` for index-stage auto-embed behavior
- `tests/unit/devtools/test_pipeline_probe.py` for probe sequence/provenance handling
- `tests/unit/cli/test_run.py` and `tests/unit/cli/test_terminal_snapshots.py` for the new chained CLI/help surface

## Notes

This is an intentional CLI break. I did not keep a long-lived compatibility layer for `--stage` because it would have forced the old mental model to remain visible everywhere the new one needs to be explicit.

The other conscious tradeoff is that not every new stage is a leaf stage exposed equally in every context. `reprocess` is still represented as a composite sequence because that is the useful behavior; the important change is that the sequence is explicit and inspectable.


Sinity:
@codex review

PR #148, opened 2026-04-07
refactor: align stage contracts, probes, and showcase coverage

Sinity:
## Summary

I tightened the integration layer around the pipeline rather than adding another large feature. The branch makes three supporting contracts line up with each other: the CLI coverage contract (derived from the real Click tree), the run/probe/planning contract (expressed in explicit stage-aware helpers), and the refresh/product support contract (localized enough that integration tests can reason about it without hydrating unrelated state). The most visible additions are `polylogue/cli/command_inventory.py`, recursive command-help generation, product-command JSON showcase coverage, and probe support for explicit stage sequences plus provenance. Kept a smaller set of runtime helpers in the branch — mostly indexing/reporting integration and localized aggregate/thread refresh helpers — because they are part of the same "make the contract verifiable" goal. The final shape is intentionally narrower than the branch's broadest intermediate experiments.

## Motivation

I wanted one place where the supporting infrastructure around the pipeline stopped disagreeing with itself. Before this work, showcase help coverage depended on curated command lists, product JSON coverage lived in more ad hoc assertions, and probe/planning/runtime surfaces each had a slightly different view of what a workflow stage meant. That is survivable while things are stable, but it becomes brittle the moment the CLI tree or stage semantics move.

The branch also had a clear false start: I briefly pulled in a much broader rollback/performance-baseline reset, but it was too wide for the problem I actually wanted to solve. The useful lesson from that detour was that the important parts were not the bulk revert itself; they were the places where we still lacked a single verifiable contract. The version here keeps those durable pieces and leaves the broader cleanup out.

### 1. Generate showcase coverage from the real command tree

`polylogue/cli/command_inventory.py` is the new anchor for this branch. It provides a small recursive inventory over the Click tree:

```python
@dataclass(frozen=True, slots=True)
class CommandPath:
    path: tuple[str, ...]
    command: click.Command
```

`iter_command_paths()` walks nested groups and yields those `CommandPath` values. `polylogue/showcase/generators.py` then uses that inventory to generate help exercises and to find product subcommands that actually expose `--json`. The effect is that showcase coverage now follows the live CLI topology, not a manually curated list that can drift.

This is especially valuable for product commands, where the registry already knows what surfaces exist. The new `generate_products_json_exercises()` keeps JSON coverage tied to that registry-backed shape.

### 2. Keep probes and run planning honest about stage semantics

The branch also keeps the stage-sequence work integrated instead of treating it as a CLI-only feature. `run_support.py`, `run_execution.py`, `run_stages.py`, and `planning_runtime.py` all keep the ordered stage model explicit, and `devtools/pipeline_probe.py` is updated so it can replay those explicit sequences rather than only a single stage label.

The probe also gains provenance and fingerprint output for staged sources / manifests. That matters because once the probe becomes part of the validation story, "what exact inputs produced this result?" has to be part of the payload, not tribal knowledge.

Conceptually, the probe does two jobs:

- it replays the same stage contract that `run_sources()` executes
- it records enough provenance to make a slow/fast result comparable later

### 3. Tie indexing/reporting integration to the same contract

There are a few pipeline-side adjustments in the branch that are not huge on their own, but are important for integration:

- stage/reporting state in `run_execution.py` / `state_views.py`
- index behavior and progress/count handling in `pipeline/services/indexing.py`
- run-int and run-source tests that assert those integrated results

The theme is consistency. If parse/materialize/index are distinct stage concepts, then the reported counts, index status, and help/probe surfaces all need to describe them the same way.

### 4. Localize session-product refresh helpers

Kept the `session_product_aggregates.py` and `session_product_threads.py` changes in this branch because they serve the same reviewability goal. Both helpers now rebuild only the requested group/root from its own profile-record slice rather than hydrating a larger shared set and regrouping afterward.

That is a small performance improvement, but more importantly it makes the integration tests easier to interpret: the code operating on a provider-day or thread root is visibly scoped to that provider-day or root.

### 5. Refresh tests and baselines around the integrated contract

The branch updates a lot of tests, but they cluster around a few themes:

- recursive help/showcase inventory
- product JSON showcase generation
- run/probe stage-sequence behavior
- localized refresh helpers
- validation-lane wiring to the updated probe contract

I wanted those pieces in one branch because the real value here is not one file; it is that the support surfaces now reinforce the same contract from several different entry points.

## Testing

Key coverage lives in:

- `tests/unit/showcase/test_exercise_catalog.py` for command inventory, nested help paths, and generated product JSON exercises
- `tests/unit/pipeline/test_run_sources.py`, `tests/unit/pipeline/test_parsing_service.py`, and `tests/unit/pipeline/test_indexing.py` for integrated stage/index behavior
- `tests/unit/devtools/test_pipeline_probe.py` and `tests/unit/devtools/test_validation_lanes.py` for probe sequence/provenance and lane wiring
- updated showcase baselines under `tests/baselines/showcase/` to pin the live help/product surfaces

## Notes

The branch is intentionally narrower than some of the intermediate commits. I am not trying to relitigate every performance/topology experiment here; I am trying to make the integrated contract easier to verify and harder to let drift.

The main thing to watch in review is silent drift. If command inventory, stage planning, probe output, and product coverage stay derived from live structures instead of curated lists, future refactors should become noticeably cheaper to review.


Sinity:

Sinity:
@codex review

PR #150, opened 2026-04-07
refactor: fold auxiliary commands into run and qa

Sinity:
## Summary

I collapsed a set of thin root commands into the broader workflows that already owned their behavior. `generate` lives under `qa`, `site` and `embed` live under `run`, and the standalone `sources` listing command is removed from the root CLI. The supporting code follows that topology change all the way through registration, help generation, stage modeling, and tests, so the new shape is reflected consistently rather than only in the visible command list. Used the move of `site`/`embed` into `run` to add the necessary stage plumbing (`site`, `embed`, and `publish`) to the run-stage model without keeping those commands as special one-offs. The result is a smaller root CLI and a clearer mapping between "workflow command" and "leaf operation."

## Motivation

These commands had become top-level mostly by inertia. `generate` already fed QA and seeded environment workflows. `site` and `embed` are downstream run operations more than independent product areas. And `sources` was a very thin listing surface that added another root command without really owning a workflow of its own.

I did not want to keep all of those as peers indefinitely because it makes the root CLI noisier while also forcing help baselines, showcase entries, and registration tests to care about topology that is mostly accidental. The better model is:

- QA owns generation/testing/demo data workflows
- run owns operational pipeline stages and downstream archive publication/search work
- the root command list should mostly show broad workflows, not every helper

### 1. Move `generate` under `qa`

`polylogue/cli/commands/qa.py` becomes a Click group and registers `generate_command` directly:

```python
from polylogue.cli.commands.generate import generate_command

qa_command.add_command(generate_command)
```

That lets us keep the generation logic almost untouched while moving the entry point to `polylogue qa generate`. Made the group behavior explicit by returning early only when a subcommand was invoked, so a plain `polylogue qa` call still runs the QA workflow exactly as before.

The examples and help text in `polylogue/cli/commands/generate.py` were rewritten to reflect the new nested path.

### 2. Fold `site` and `embed` into `run`

This is the more structural part of the branch. `site` and `embed` stop being root commands and become run stages / run subcommands instead.

On the run-model side:

- `RUN_STAGE_CHOICES` gains `site`, `embed`, and `publish`
- `RUN_STAGE_SEQUENCES` maps those into leaf-stage behavior
- `PlanStage` gains `SITE` and `PUBLISH`
- `run_stages.py` adds `execute_site_stage()` and `execute_embed_stage()`

On the CLI side, `run.py` grows stage-specific option carriers (`EmbedOptions`, site-option resolution, etc.) so `run` can dispatch these nested commands without overloading the normal pipeline flow.

`embed.py` is intentionally kept as a helper module, but the Click decorator is removed so the business logic can be reused by `run embed` instead of living behind a separate root command.

### 3. Remove the standalone `sources` root command

This is the smallest code change, but an important topology cleanup. The standalone sources-listing command disappears from root registration, from showcase baselines, and from deterministic JSON/help tests.

Chose removal instead of relocation because it did not fit as a run stage or QA subcommand cleanly. It was simply too small a surface to justify permanent root status.

### 4. Update the command inventory, showcase, and tests to match

Most of the supporting file churn is here:

- `click_command_registration.py` stops registering the removed root commands
- showcase/help baselines move from `help-generate`, `help-site`, `help-embed` to `help-qa-generate`, `help-run-site`, and `help-run-embed`
- site/embed tests now invoke the nested `run` entry points
- CLI metadata and error-boundary tests are updated around the smaller root command set

I wanted the branch to be explicit about the topology change. A command move is only "done" once the help baselines, snapshots, and tests describe the same tree the user sees.

## Testing

The branch updates the tests closest to the moved surfaces:

- `tests/unit/cli/test_click_app.py` and `tests/unit/cli/test_cli_error_boundaries.py` for registration / root-command changes
- `tests/unit/cli/test_run.py`, `tests/unit/cli/test_embed.py`, and `tests/unit/cli/test_site.py` for nested run-stage behavior
- `tests/integration/test_site.py` and `tests/integration/test_site_laws.py` for site generation through the new `run site` path
- updated showcase baselines and snapshots for the moved help surfaces
- QA generate path checks in the updated CLI tests

## Notes

I did not add compatibility aliases for the removed root commands. The point of the branch is to make the command tree smaller and more coherent, and aliases would leave both trees alive.

The one place where I preserved indirection is `embed.py`: the command decorator goes away, but the helper functions stay because the embedding runtime logic should remain testable independent of its CLI placement.


Sinity:

Sinity:
@codex review

PR #152, opened 2026-04-07
refactor: rename check/qa to doctor/audit and add governance

Sinity:
## Summary

I renamed the archive health and QA commands from `check`/`qa` to `doctor`/`audit`, then tightened the surrounding tooling so the renamed CLI surface is mechanically covered without being maintained by convention. The rename is not just a help-text pass: it reaches validation helpers, showcase fixtures, PTY snapshots, live validation lanes, and the synthetic generation examples under the audit group. Added a small governance registry that declares which command paths are expected to expose JSON or seeded showcase coverage, with tests that compare the registry against the actual Click tree and exercise catalog. Finally, I made provider filters reject unknown tokens at option-parse time and wired shell completion from the same provider list, which closes a usability hole that became more visible once the command tree got stricter.

## Motivation

`check` and `qa` both read like implementation placeholders rather than user-facing commands. In practice, `check` is the archive doctoring surface for health, repair, and proof operations, while `qa` is the orchestrator for schema audit, exercises, and invariants. Renaming them only in the Click decorators would still leave the rest of the repository talking in the old vocabulary: baseline fixtures, deterministic-output tests, live validation lanes, and docs examples would keep drifting.

The second issue is that the CLI had no mechanical statement of coverage expectations. We already rely heavily on command inventory and generated exercises, but there was nothing that said “this path must have `--json`” or “this product command must have a seeded exercise.” The new governance matrix is intentionally small, but it gives the tests something explicit to validate.

Folded in provider-token validation because the root query options were still permissive in an unhelpful way. A typo in `--provider` silently became a no-op filter. With stricter command naming and help surfaces, that kind of late failure stood out more.

### Command rename and error surfaces

The public CLI changes are simple:

```bash
polylogue doctor --json
polylogue doctor --repair --preview
polylogue audit --only exercises --tier 0
polylogue audit generate --seed
```

Under the hood, I updated the Click entrypoints and the validation helpers that format usage errors, so the error surface is consistent with the new command names instead of leaking `check` in exceptions:

```python
fail("doctor", "--schema-samples must be a positive integer or 'all'")
fail("doctor", "--vacuum requires --repair or --cleanup")
```

Updated the examples under `polylogue/cli/commands/generate.py` so the nested generator help reads as `polylogue audit generate...` rather than preserving the old group name.

### Governance matrix for command coverage

The governance layer lives in a new `polylogue/cli/governance.py` module:

```python
@dataclass(frozen=True, slots=True)
class GovernanceSpec:
    has_json: bool = False
    has_seeded_exercise: bool = False
```

and a `GOVERNANCE` map keyed by command-path display names such as `doctor`, `audit generate`, and `products analytics`.

The paired tests do two things:
- ensure every command path discovered from the Click tree has a governance entry and that there are no stale declarations, and
- ensure the declarations are truthful by checking actual Click parameters and showcase exercise presence.

This is intentionally mechanical rather than clever. The repository already has a lot of CLI surface area, and I want new commands to force an explicit coverage decision instead of quietly bypassing JSON or seeded-exercise expectations.

### Provider filter validation and completion

In `polylogue/cli/click_option_groups.py` I derive CLI-valid providers from `CORE_SCHEMA_PROVIDERS` and use that list both for validation and shell completion:

```python
click.option(
    "--provider",
    "-p",
    callback=_validate_provider_tokens,
    shell_complete=_complete_providers,
)
```

That means a typo now fails at parse time with a concrete list of valid providers instead of producing an empty filter later in execution. The same logic is applied to `--exclude-provider`.

### Showcase, baselines, and live lanes

The rename required coordinated fixture work:
- `polylogue/showcase/exercise_catalog.json` renames doctor/audit help exercises and the proof-related showcase entries.
- new help baselines land for `help-doctor.txt`, `help-audit.txt`, and `help-audit-generate.txt`.
- old `help-check*` and `help-qa*` baselines are removed.
- `devtools/validation_lane_catalog_live.py` and its tests switch live lanes to `doctor` and `audit`.

The rename therefore stays honest all the way from `--help` output to the showcase recorder and validation-lane construction.

## Testing

I updated or added coverage in:
- `tests/unit/cli/test_check.py` for renamed `doctor` invocations
- `tests/unit/cli/test_click_app.py` and `test_click_app_main.py` for command routing and error wrapping
- `tests/unit/cli/test_deterministic_output.py` and `test_json_envelope_contract.py` for renamed JSON surfaces
- `tests/unit/cli/test_terminal_snapshots.py` and the `.ambr` snapshots for rendered help output
- `tests/unit/devtools/test_validation_lanes.py` for renamed live commands
- `tests/unit/showcase/test_exercise_catalog.py` for renamed showcase exercises
- `tests/unit/cli/test_governance.py` for governance completeness and truthfulness
- provider filter validation via the `click_option_groups.py` callback path

## Notes

There is no compatibility shim for `check` or `qa`; the repository rules explicitly prefer delete-and-replace over long-lived aliases. Scripts and docs need to move to `doctor` and `audit` immediately.

The governance map is deliberately declarative rather than generated. I want reviewers to notice new command paths and decide whether JSON and seeded coverage are expected, not inheriting behavior accidentally.


Sinity:
@codex review

PR #154, opened 2026-04-07
feat: add provider assembly layer for sidecar title resolution

Sinity:
## Summary

I extracted sidecar discovery and post-parse enrichment into a provider-level assembly protocol so title correction is no longer a Claude Code special case embedded in the source walk. The new assembly layer lets each provider discover its own sidecars and enrich parsed conversations in one place, which allowed me to add Codex thread-name resolution without teaching `source_walk.py` or `_ConversationEmitter` about Codex-specific files. Tightened title correctness for Claude Code by refusing to treat obvious git branch names as human titles, and I now record `title_source` in provider metadata so downstream consumers can understand where a title came from. The parse pipeline remains content-driven; the new abstraction only runs on parse paths, not on raw-data acquisition.

## Motivation

Before this change, the source-walk and emitter path knew a very specific fact about Claude Code: some directories contain a `sessions-index.json` sidecar, and parsed conversations should be enriched from it. That coupling was already awkward on its own, but it became an obvious blocker once Codex needed similar treatment. Codex stores thread names in `session_index.jsonl`, which is the same architectural shape—a provider-owned sidecar that should improve parsed conversations after raw payload parsing.

I wanted the abstraction boundary to follow that shape instead of branching inside the generic pipeline. Rather than adding a second bespoke sidecar path for Codex, I moved the concept into a provider protocol with two responsibilities: discover sidecars from source paths, and enrich a `ParsedConversation` from the discovered data.

Title correctness was part of the same problem. Claude Code summaries from the side index can contain branch names like `feature/auth-fix`, and those should not outrank a real prompt-derived title. Once title enrichment became provider-owned, it made sense to fix that heuristic at the same time and expose the decision as metadata.

### Provider assembly protocol

The new core surface lives in `polylogue/sources/assembly.py`:

```python
class ProviderAssemblySpec(Protocol):
    def discover_sidecars(self, source_paths: list[Path]) -> dict[str, Any]: ...
    def enrich_conversation(
        self,
        conv: ParsedConversation,
        sidecar_data: dict[str, Any],
    ) -> ParsedConversation: ...
```

and a simple registry:

```python
def get_assembly_spec(provider: Provider) -> ProviderAssemblySpec | None:
    if provider is Provider.CLAUDE_CODE:
        return ClaudeCodeAssemblySpec()
    if provider is Provider.CODEX:
        return CodexAssemblySpec()
    return None
```

That gives the generic pipeline exactly one place to ask, “does this provider have sidecars and enrichment rules?”

### Claude Code assembly and title-source metadata

`ClaudeCodeAssemblySpec` mostly wraps existing behavior, but it moves it behind the new boundary. `discover_sidecars()` walks the relevant parents, parses `sessions-index.json` once per directory, and flattens the entries by session id so lookup stays cheap during enrichment.

Improved `polylogue/sources/parsers/claude_index.py` so branch-shaped summaries are rejected:

```python
if index_entry.summary and not _looks_like_git_branch(index_entry.summary):
    title = index_entry.summary
    title_source = "session-index:summary"
elif index_entry.first_prompt and index_entry.first_prompt != "No prompt":
    title = index_entry.first_prompt[:80]
    title_source = "session-index:first-prompt"
```

That keeps the existing summary-first behavior for real summaries but avoids titles like `main` or `feature/refactor/cleanup`.

### Codex sidecar support and fallback titles

The new `polylogue/sources/assembly_codex.py` module reads `~/.codex/session_index.jsonl` relative to any discovered `sessions/` root. I intentionally made the parser tolerant of append-only history and slightly different field names (`id`/`thread_id`, `thread_name`/`name`). The last name for a thread wins, which matches the append-only file semantics.

When a parsed Codex conversation still has the raw session id as its title, enrichment now follows a two-step strategy:
1. prefer the thread name from `session_index.jsonl`, and
2. otherwise, use the first non-empty user message preview.

That fallback matters because Codex session payloads often carry usable conversational openings even when no side index is present.

### Parse-path plumbing

To support the new abstraction, I generalized the parse context:
- `_ParseContext.session_index` became `_ParseContext.sidecar_data`
- `_setup_source_walk(..., build_session_indices=...)` became `_setup_source_walk(..., discover_sidecars=...)`
- `iter_source_conversations_with_raw()` now asks the provider assembly to discover sidecars, while `iter_source_raw_data()` explicitly does not

`_ConversationEmitter._apply_provider_enrichment()` is the only place that calls the registry:

```python
spec = get_assembly_spec(p)
if spec is not None:
    return spec.enrich_conversation(conv, self._ctx.sidecar_data)
```

That change removes Claude-specific imports from the emitter and makes unsupported providers a clean no-op.

## Testing

I added `tests/unit/sources/test_assembly.py` to cover:
- provider registry resolution
- Claude Code sidecar discovery and summary/first-prompt enrichment
- Codex `session_index.jsonl` parsing, including malformed lines and newest-entry-wins behavior
- first-user-message fallback for Codex
- git-branch rejection for Claude Code summaries

Updated `tests/unit/sources/test_source_laws.py` so the helper context passes `sidecar_data` instead of `session_index`, and so the enrichment contract asserts the new `title_source` metadata.

## Notes

This change is intentionally additive for providers without sidecars. `get_assembly_spec()` returns `None` for everyone else, so ChatGPT, Claude web, Gemini, and unknown providers stay on the old parse path.

I did not try to generalize sidecars beyond title enrichment in this pass. The new protocol is broad enough for richer provider-owned enrichment later, but right now I kept it scoped to sidecar discovery and post-parse conversation mutation.

### DIFF
--- a/polylogue/sources/assembly.py
+++ b/polylogue/sources/assembly.py
@@ -1,0 +1,62 @@
+"""Provider assembly layer — generic sidecar discovery and conversation enrichment.
+
+Replaces Claude-specific session index logic with a protocol that any provider
+can implement for sidecar discovery and post-parse enrichment.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Protocol
+
+from polylogue.types import Provider
+
+from .parsers.base import ParsedConversation
+
+
+@dataclass(frozen=True, slots=True)
+class TitleResolution:
+    """Result of provider-specific title resolution."""
+
+    title: str
+    source: str  # e.g. "session-index:summary", "first-user-message", "session-id"
+
+
+class ProviderAssemblySpec(Protocol):
+    """Provider-specific sidecar discovery and conversation enrichment."""
+
+    def discover_sidecars(self, source_paths: list[Path]) -> dict[str, Any]:
+        """Discover provider-specific sidecars from source paths.
+
+        Returns opaque provider data keyed by parent directory or similar.
+        """
+        ...
+
+    def enrich_conversation(
+        self,
+        conv: ParsedConversation,
+        sidecar_data: dict[str, Any],
+    ) -> ParsedConversation:
+        """Enrich a parsed conversation using discovered sidecar data."""
+        ...
+
+
+def get_assembly_spec(provider: Provider) -> ProviderAssemblySpec | None:
+    """Return the assembly spec for a provider, or None if no enrichment needed."""
+    if provider is Provider.CLAUDE_CODE:
+        from .assembly_claude_code import ClaudeCodeAssemblySpec
+
+        return ClaudeCodeAssemblySpec()
+    if provider is Provider.CODEX:
+        from .assembly_codex import CodexAssemblySpec
+
+        return CodexAssemblySpec()
+    return None
+
+
+__all__ = [
+    "ProviderAssemblySpec",
+    "TitleResolution",
+    "get_assembly_spec",
+]
--- a/polylogue/sources/assembly_claude_code.py
+++ b/polylogue/sources/assembly_claude_code.py
@@ -1,0 +1,50 @@
+"""Claude Code provider assembly — sessions-index.json sidecar discovery and enrichment."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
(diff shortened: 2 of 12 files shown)


Sinity:
@codex review

PR #156, opened 2026-04-07
feat: add full-corpus schema inference and proof output

Sinity:
## Summary

I pushed the schema pipeline toward corpus-first behavior by adding an explicit full-corpus mode, making semantic-role eligibility depend on artifact kind, and exposing a proof surface for role-assignment review. Before this change, schema generation always flowed through sample caps and exemplar caps somewhere in the stack, which meant “give me the provider schema” still really meant “give me a bounded sample of the provider schema.” I also stopped record-stream artifacts from attempting impossible roles like `conversation_title`, and I renamed the annotation key from `x-polylogue-confidence` to `x-polylogue-score` to reflect that the value is a heuristic ranking rather than a calibrated confidence measure. The operator-facing `schema explain` path can emit the full proof surface when needed.

## Motivation

There were three related problems in the existing schema workflow.

First, full-corpus generation was not actually expressible end to end. Some call sites allowed `max_samples=None`, but the sampling and generation stack still applied caps through record extraction, provider sample limits, or structure exemplar limits. That makes corpus-scale alignment work difficult because it is easy to believe you are reviewing a provider’s whole shape while still looking at a bounded approximation.

Second, semantic-role inference treated all artifacts as if they were equally eligible for all roles. Record-stream artifacts are the clearest counterexample: an individual record stream can have message bodies, message roles, and timestamps, but it does not meaningfully carry a conversation-level title. The previous inference path would still score title candidates and could annotate them.

Third, `schema explain` only exposed the selected annotations. It did not tell an operator which roles were ineligible, which ones abstained for lack of viable candidates, or which competing paths lost. That made review harder, especially once role eligibility started depending on artifact kind.

### Full-corpus generation path

The CLI makes the intent explicit:

```bash
polylogue schema generate --provider codex --full-corpus
```

I threaded that flag through the operator request models and down into the sampling/generation stack:
- `SchemaInferRequest.full_corpus`
- `generate_provider_schema(..., full_corpus=...)`
- provider bundle construction
- cluster collection
- schema-unit iteration
- record extraction and exemplar selection

The key behavioral changes are:
- `extract_record_samples_from_raw_content(..., max_samples=None)` returns every valid record in the stream without applying caps
- `_iter_schema_units_from_db()` treats `full_corpus=True` as “do not backfill from the provider sample cap”
- `_generate_cluster_schema()` disables the per-fingerprint exemplar cap when full-corpus mode is enabled

That means full-corpus mode is not just a higher number; it is a bypass of the places that were previously forcing approximation.

### Artifact-aware semantic role eligibility

`polylogue/schemas/semantic_inference_runtime.py` now defines explicit record-stream kinds and the roles they are allowed to infer:

```python
RECORD_STREAM_KINDS = frozenset({
    "conversation_record_stream",
    "subagent_conversation_stream",
})

RECORD_STREAM_ELIGIBLE_ROLES = frozenset({
    "message_container",
    "message_role",
    "message_body",
    "message_timestamp",
})
```

`infer_semantic_roles()` accepts `artifact_kind`, and `annotate_semantic_and_relational()` passes it through. The effect is that record-stream artifacts simply abstain from `conversation_title` instead of pretending title inference is meaningful there.

To make that possible, I also threaded `artifact_kind` through cluster schema generation and provider-bundle packaging so the annotation step knows what it is annotating.

### Score rename and proof surface

I renamed the annotation key from `x-polylogue-confidence` to `x-polylogue-score` across emitters, readers, explain output, and audit walkers. The old term implied more calibration than we really have; the new term better matches the underlying rank/scoring behavior.

The more interesting operator-facing addition is `--proof` on `schema explain`:

```bash
polylogue schema explain --provider claude-code --proof
```

Internally, `build_review_proof()` walks the annotated schema and constructs:
- eligible roles
- ineligible roles
- per-role chosen paths and scores
- competing candidates
- abstention reasons when no role is selected

The terminal renderer then prints a dedicated proof surface instead of the normal explain view when `review_proof` is present.

### Updated models and explain results

`SchemaExplainRequest` carries `proof: bool`, and `SchemaExplainResult` can carry `review_proof`. The dataclasses added for that surface are:
- `SchemaRoleProofEntry`
- `SchemaReviewProof`

Updated `SchemaRoleAssignment.to_dict()` to emit `score` instead of `confidence`, which keeps the operator-model JSON consistent with the schema annotation rename.

## Testing

`tests/unit/core/test_schema_corpus_alignment.py` is the main new coverage file. It checks:
- full-corpus extraction returning all records
- capped extraction still staying bounded
- record-stream artifacts abstaining from title inference
- document artifacts continuing to infer titles
- proof surfaces containing all roles, abstentions, and eligibility sets
- `x-polylogue-confidence` being fully removed from source and generated schemas

Updated `tests/unit/core/test_schema_annotation_contracts.py` so existing annotation-contract coverage now asserts `x-polylogue-score`.

## Notes

The `confidence` -> `score` rename is intentionally breaking for downstream consumers who parse these annotations directly. I did not add a duplicate compatibility key because that would make it easier for the codebase to drift back into supporting both terms indefinitely.

Full-corpus mode is potentially expensive by design. The point of the flag is to trade bounded generation for fidelity when an operator explicitly asks for it, not to change the default schema-generation cost profile.


Sinity:
@coderabbitai full review

Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 9 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550).

Sinity:
@codex review

PR #158, opened 2026-04-07
refactor: replace query flags with positional verbs

Sinity:
## Summary

I turned the action-like query flags into positional verbs so the query-first CLI reads like “filters, then action” instead of accumulating more root options that secretly behave like subcommands. The new surface introduces real Click commands for `list`, `count`, `stats`, `open`, and `delete`, moves their dedicated options onto those commands, and keeps the actual query execution path unchanged by translating verb invocations back into the existing parameter flags. Updated the argv splitter so verbs can appear after filters or search terms without blocking regular subcommand routing. While doing that cleanup, I generalized JSON contract exercise generation from product-only coverage to any command that actually exposes `--json`, which made the hand-maintained governance matrix unnecessary and let me remove it.

## Motivation

The root CLI had accumulated a cluster of options that were not really options:
- `--list`
- `--count`
- `--stats`
- `--stats-by`
- `--open`
- `--delete`
- `--dry-run`
- `--force`

Those flags each switched the command into a different execution mode, which is exactly what subcommands are for. Keeping them at the root made the help output harder to teach, made query parsing more special-cased, and forced the mental model “type filters, then remember which magic flag changes the action.”

Once I moved those modes into verbs, it became clearer that our coverage machinery was also still overly flag-shaped. The governance matrix introduced earlier was useful, but it duplicated information that the command inventory can discover directly for JSON-capable commands. Since this commit already touches the query surface and showcase generation, it was a good place to replace the manual governance map with generated JSON contract exercises.

### New verb commands

The new CLI surface lives in `polylogue/cli/query_verbs.py`:

```python
VERB_NAMES = frozenset({"list", "count", "stats", "open", "delete"})
```

Each verb is a real Click command. For example:

```python
@click.command("stats")
@click.option("--by", "stats_by", type=click.Choice([...]))
@click.option("--format", "-f", "output_format", type=click.Choice([...]))
@click.pass_context
def stats_verb(ctx, stats_by, output_format):
    params = _parent_params(ctx)
    params["stats_only"] = True
    ...
    _execute_query_verb(ctx, params)
```

The verbs do not introduce new query-engine code. They just collect the root filter state from `ctx.parent.params`, set the corresponding execution switches, and call the existing `execute_query()` path. That keeps the behavioral risk low while giving Click a much better command tree to expose.

The root help now teaches the intended patterns directly:

```bash
polylogue "error" -p claude-ai --since 2025-01 list
polylogue -p chatgpt count
polylogue --provider codex stats --by provider
polylogue --latest open
polylogue "urgent" delete --dry-run
```

### Root option cleanup

With the verbs in place, `polylogue/cli/click_option_groups.py` no longer needs to define root-level `--list`, `--count`, `--stats`, `--stats-by`, `--open`, or delete modifiers. That simplifies the root signature and also clarifies what still counts as a “root-mode request” in `polylogue/cli/root_request.py`: filters, limits, streaming, dialogue-only views, and metadata mutation remain root concerns, while execution-mode switches move into verbs.

This is also why the delete safety check still lives in the query planner rather than the verb itself. The verb only sets `delete_matched=True`; the planner still enforces “you must have at least one filter before delete is legal.”

### Query parsing changes

The interesting routing work is in `_split_query_mode_args()` inside `polylogue/cli/query.py`. Non-verb subcommands are still only recognized when they appear before any bare query terms, but verbs are recognized anywhere in the argv stream and terminate the query-term collection:

```python
if arg in VERB_NAMES:
    verb_args = option_args + [arg] + list(args[index + 1 :])
    return verb_args, tuple(query_terms), True
```

That keeps these cases working:
- `polylogue "search terms" list`
- `polylogue --provider codex count`
- `polylogue --path foo --action file_read list`

while still routing real subcommands such as `run`, `doctor`, or `schema` normally when they appear at the front.

Updated the error text to point users toward `list` rather than `--list`, since that old hint would otherwise become actively misleading.

### Showcase and JSON contract generation

The showcase catalog and generator layer had to move with the CLI:
- static exercises switched from `--list`/`--count`/`--stats-by` to `list`/`count`/`stats --by`
- generated filter exercises now append `["list", "-n", "3"]`
- generated format exercises treat `list` and `count` as modes rather than flag bundles

More importantly, I generalized JSON contract exercise generation:

```python
def generate_json_contract_exercises() -> list[Exercise]:
    for cp in inventory_command_paths():
        if _has_json_flag(cp.command):
            ...
```

That means JSON contract coverage is no longer limited to product subcommands. Any command with `--json` gets a generated contract exercise. With that in place, the manual governance map and its tests were redundant and I removed them.

## Testing

I updated:
- `tests/integration/test_cli_query_mode.py` for `count`, `list -f json`, and `stats --by provider`
- `tests/unit/cli/test_click_app.py` for command inventory, help output, and verb registration
- `tests/unit/cli/test_query_exec.py` and `test_query_exec_laws.py` for delete-plan errors, no-result hints, and flag-to-verb routing
- `tests/unit/cli/test_cli_error_boundaries.py` so the new verbs are treated as real commands
- terminal/help snapshots in `tests/unit/cli/__snapshots__/test_terminal_snapshots.ambr`
- showcase generation and catalog tests so generated and static exercises speak the new verb surface

I removed `polylogue/cli/governance.py` and `tests/unit/cli/test_governance.py` because JSON contract exercise generation enforces that surface mechanically from the command inventory.

## Notes

There is no flag compatibility layer. Existing invocations need to migrate, for example:
- `polylogue "error" --list -n 5` -> `polylogue "error" list -n 5`
- `polylogue --stats-by provider` -> `polylogue stats --by provider`
- `polylogue -p old --delete --dry-run` -> `polylogue -p old delete --dry-run`

The query engine still uses the old internal booleans (`list_mode`, `count_only`, `stats_only`, `open_result`, `delete_matched`) as the execution plan inputs. I left that intact intentionally so this refactor stays about CLI shape and coverage generation rather than reopening the execution planner.


Sinity:

Sinity:
@codex review

PR #160, opened 2026-04-07
feat: add provider events for compaction records

Sinity:
## Summary

I introduced a first-class provider-event surface for compaction and turn-context records so parser outputs can represent non-message provider artifacts without smuggling them through normal message structures. The main target is compaction: both Claude Code and Codex emit records that materially change session interpretation, but those records were either buried in provider-specific message streams or reduced to ad hoc metadata. This change normalizes those records, adds `ParsedProviderEvent` to the parser model, and wires compaction counts into session profiles and evidence products. I kept `provider_meta["context_compactions"]` populated for now so downstream profile code stays stable while the new event surface lands.

## Motivation

Compaction is semantically important. It tells us that a long-running session crossed a context boundary and gives us information about what was preserved or summarized. That is useful for analytics, evidence products, and session interpretation, but the parser layer did not have a proper place to represent it. Claude Code had legacy `summary` records and a newer `system` / `compact_boundary` shape. Codex has `compacted` records and `turn_context` envelopes. None of those are normal messages, and forcing them through message parsing is the wrong model.

At the same time, session products already benefit from knowing whether compaction happened. Rather than waiting for every downstream consumer to migrate to the new provider-event surface before shipping anything, I wanted a transitional change that:
1. normalizes the provider-specific record shapes,
2. exposes them as first-class parser output, and
3. keeps the existing metadata path alive long enough for session products to consume the new information immediately.

### Normalized compaction detection

`polylogue/pipeline/semantic_capture.py` now normalizes two Claude Code shapes inside `detect_context_compaction()`:
- legacy `{"type": "summary",...}`
- modern `{"type": "system", "subtype": "compact_boundary",...}`

The returned structure is consistent across both shapes:

```python
{
    "summary": ...,
    "timestamp": ...,
    "trigger": ...,
    "pre_tokens": ...,
    "preserved_segment_id": ...,
    "is_modern": ...,
}
```

I made the extractor tolerant of string or block-based message content and of both `preTokens` and `pre_tokens` because the point of the helper is to hide provider/version differences from the parser.

### First-class provider events

The core parser-model addition is:

```python
class ParsedProviderEvent(BaseModel):
    event_type: str
    timestamp: str | None = None
    payload: dict[str, object] = Field(default_factory=dict)
```

and `ParsedConversation` carries:

```python
provider_events: list[ParsedProviderEvent] = Field(default_factory=list)
```

This gives the parser layer a clean place to store compaction and turn-context records without pretending they are assistant/user/system messages.

### Provider-specific parser changes

For Claude Code:
- `ClaudeCodeRecord` now treats both `summary` and `system` / `compact_boundary` as context-compaction records.
- `parse_code()` appends a `ParsedProviderEvent(event_type="compaction",...)` whenever compaction is detected and skips those records in normal message handling.

For Codex:
- `CodexRecord` gained `is_compaction`, `is_turn_context`, `compacted_message`, and `has_replacement_history`.
- `parse()` handles `compacted` envelopes before message parsing, emits `compaction` events, emits `turn_context` events, and continues to parse adjacent response items as normal messages.

The key behavior here is that provider events are additive; they do not reduce message extraction fidelity.

### Backward-compatible metadata and session products

Kept `provider_meta["context_compactions"]` populated for both parsers so existing downstream code continues to work. Session profile generation still reads compaction information from conversation metadata for now:

```python
compactions = conversation.provider_meta.get("context_compactions")
if isinstance(compactions, list):
    compaction_count = len(compactions)
```

That let me wire the new product-facing fields immediately:
- `SessionProfile.compaction_count`
- `SessionEvidencePayload.compaction_count`
- `SessionEvidencePayload.has_compaction`

and the storage serializer writes both `compaction_count` and `has_compaction` into the evidence payload.

### Exported parser surface

`polylogue/sources/parsers/base.py` now re-exports `ParsedProviderEvent`, so code importing parser model types from the base module can access the new event class without reaching into `base_models.py` directly.

## Testing

The new `tests/unit/sources/test_compaction.py` file covers:
- legacy and modern Claude Code compaction detection
- `ClaudeCodeRecord.is_context_compaction`
- provider-event emission from the Claude Code parser
- `CodexRecord` helpers for compaction and turn-context envelopes
- provider-event emission from the Codex parser
- preservation of `provider_meta["context_compactions"]`
- session-profile and evidence-payload compaction fields

The tests also explicitly check that compaction records do not get counted as normal messages.

## Notes

This is intentionally a transitional design. `provider_events` is the long-term clean surface, but session products still populate `compaction_count` from backward-compatible metadata in this commit. That minimizes migration risk while making the new event model available immediately.

I scoped the change to compaction and turn-context events because those are the clearest non-message provider artifacts in the current providers. The `ParsedProviderEvent` surface is broader than that, but I did not want to invent consumers for hypothetical future event types in the same patch.

### DIFF
--- a/polylogue/archive_products.py
+++ b/polylogue/archive_products.py
@@ -64,6 +64,8 @@ class SessionEvidencePayload(ArchiveProductModel):
     total_duration_ms: int = 0
     wall_duration_ms: int = 0
     cost_is_estimated: bool = False
+    compaction_count: int = 0
+    has_compaction: bool = False
     tool_categories: dict[str, int] = Field(default_factory=dict)
     repo_paths: tuple[str, ...] = ()
     cwd_paths: tuple[str, ...] = ()
--- a/polylogue/lib/session_profile_models.py
+++ b/polylogue/lib/session_profile_models.py
@@ -46,6 +46,7 @@ class SessionProfile:
     cost_is_estimated: bool = False
     thread_id: str | None = None
     continuation_depth: int = 0
+    compaction_count: int = 0
     tags: tuple[str, ...] = ()
     auto_tags: tuple[str, ...] = ()
     is_continuation: bool = False
@@ -103,6 +104,7 @@ class SessionProfile:
             "engaged_minutes": round(self.engaged_duration_ms / 60_000.0, 4),
             "wall_duration_ms": self.wall_duration_ms,
             "cost_is_estimated": self.cost_is_estimated,
+            "compaction_count": self.compaction_count,
             "thread_id": self.thread_id,
             "continuation_depth": self.continuation_depth,
             "tags": list(self.tags),
@@ -184,6 +186,7 @@ class SessionProfile:
             engaged_duration_ms=int(payload.get("engaged_duration_ms", 0) or 0),
             wall_duration_ms=int(payload.get("wall_duration_ms", 0) or 0),
             cost_is_estimated=bool(payload.get("cost_is_estimated", False)),
+            compaction_count=int(payload.get("compaction_count", 0) or 0),
             thread_id=str(payload["thread_id"]) if payload.get("thread_id") is not None else None,
             continuation_depth=int(payload.get("continuation_depth", 0) or 0),
             tags=tuple(str(item) for item in payload.get("tags", []) or []),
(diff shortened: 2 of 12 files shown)


Sinity:
@codex review

PR #162, opened 2026-04-07
refactor: inherit root query filters in product commands

Sinity:
## Summary

I wired the root query filters into product commands so callers can write `polylogue --provider codex products profiles` instead of having every product surface redeclare `--provider`, `--since`, and `--until` for itself. The core of the change is in the product callback layer: product commands walk up to the root Click context, read inherited filter params, and inject only the subset their query model actually accepts. To make that work reliably, I also adjusted query-mode routing so leading filter options do not prevent `products` from being recognized as a real subcommand. The registry surface and help baselines are correspondingly smaller because duplicated provider/time options disappear from many product commands.

## Motivation

The CLI already has a strong root-level filtering story. Query mode accepts provider and date filters at the top level, and the positional-verb refactor made that model even more explicit. Product commands were the remaining major holdout: many of them still redeclared `--provider`, `--since`, and `--until` locally, which forced users to switch mental models mid-command.

That duplication has a few concrete costs:
- help output is longer than it needs to be
- live validation and examples have to place common filters after the product subcommand even though those filters are conceptually global
- the registry has to carry repeated option declarations that are already represented at the root
- parsing gets more awkward around cases like `--provider x products...`

The goal here is not to make product commands magical; it is to make them consistent with the rest of the query-first CLI by letting them inherit the filters that already belong to the root.

### Product callback inheritance

The core change is in `polylogue/cli/commands/products.py`.

I added a helper to walk to the root context:

```python
def _find_root_params(ctx: click.Context) -> dict[str, Any]:
    cur = ctx
    while cur.parent is not None:
        cur = cur.parent
    return cur.params
```

and changed product callbacks from `@click.pass_obj` to `@click.pass_context`, which lets them read both the `AppEnv` and the root params from the Click context chain.

I did not blindly inject every root param. Instead, I resolve the query model class for the product type and only copy the inherited keys that the model actually supports:

```python
query_cls = _resolve_query_class(pt.query_class_path)
accepted_root_keys = frozenset(
    k for k in _ROOT_FILTER_KEYS if k in query_cls.model_fields
)
```

That keeps inheritance explicit and avoids smuggling unrelated root params into product query objects.

### Query routing for filtered subcommands

Once product filters move to the root, commands like this need to route cleanly:

```bash
polylogue --provider claude-code products analytics --json
polylogue --provider codex --since 2026-03-01 products day-summaries --json
```

The query-mode splitter in `polylogue/cli/query.py` previously tied non-verb subcommand routing too closely to “did anything precede the command name?”. I relaxed that logic so leading filter options do not block subcommand detection as long as we have not collected bare query terms yet.

That distinction matters:
- `--provider claude-ai my_search` should still be treated as a query
- `--provider claude-ai products --help` should route to the `products` subcommand

The updated tests in `tests/unit/cli/test_click_app.py` cover exactly that boundary.

### Product registry cleanup

Once root inheritance exists, many local product options become duplication. I removed repeated `provider`/`since`/`until` declarations from the registry entries that can consume those values from the root. Product-specific filters remain where they belong:
- `--session-date-since`
- `--conversation-id`
- `--kind`
- `--query`
- and similar leaf-specific options

The effect is that product help becomes narrower and more obviously about product-specific shaping rather than repeating global filters on every command.

### Validation lanes and baselines

The live lane for day summaries now leads with the root filters before the subcommand:

```python
"--provider", "claude-code",
"--since", "2026-03-01",
"products", "day-summaries",
```

Refreshed the help baselines for the new verb-aware main help and the lightweight verb subcommands (`help-list.txt`, `help-count.txt`, `help-stats.txt`, `help-open.txt`, `help-delete.txt`). Several product help baselines shrink because inherited provider/time filters are no longer printed there.

## Testing

I updated:
- `tests/unit/cli/test_products.py` so product JSON tests use root-level `--provider` before `products`
- `tests/unit/cli/test_products.py` callback tests to build a minimal Click context for the new `@click.pass_context` callback path
- `tests/unit/cli/test_click_app.py` to assert that filter options before a real subcommand still route to that subcommand, while filter options before a bare word remain query mode
- live lane command construction via `devtools/validation_lane_catalog_live.py`
- showcase help baselines for the new help surface

## Notes

This is intentionally a CLI refactor, not a product query-model refactor. The product query models stay authoritative about which filters they accept; the CLI only inherits root params that those models already declare.

The migration is simple but visible: users need to move global product filters before `products`. For example:
- `polylogue products analytics --provider claude-code --json`
 becomes
- `polylogue --provider claude-code products analytics --json`


Sinity:
@codex review

PR #164, opened 2026-04-07
test: add schema-driven semantic parser properties

Sinity:
## Summary

I expanded the schema-driven parser test story from “does not crash” into “does not crash and still returns semantically valid structures.” The new property suite uses schema-conformant payload generation to assert valid roles, valid content block types, title stability, and `ParsedProviderEvent` round-tripping, while the existing crashlessness harness now draws provider-specific payloads without feeding every parser a ChatGPT-shaped sample. Propagated the same matrix mindset into showcase generation by adding provider × feature exercises, and into mutation testing by defining a parser-focused mutmut campaign. The result is a test surface that is both more semantically meaningful and better aligned with the actual providers under test.

## Motivation

The existing crashlessness tests were useful, but they left two gaps.

First, they only proved that parsers did not explode on generated input. That is necessary, but not sufficient. A parser can survive a payload and still emit semantically invalid output: an unknown role, an impossible content block type, or a title that changes between parses of the same payload.

Second, the crashlessness harness itself was not as provider-aware as it should have been. It used `schema_conformant_payload("chatgpt")` directly in places that were supposed to exercise multiple providers, which means some providers were not actually getting inputs shaped by their own schemas.

The rest of the branch is about applying the same idea consistently. If provider capabilities differ—some have `thinking`, some only `tool-use`—showcase exercises should reflect that. If parser semantics are especially sensitive in a small set of modules, mutation testing should target that cluster directly.

### Semantic property tests

The new `tests/property/test_semantic_properties.py` file adds four kinds of checks.

1. Valid roles:
```python
assert msg.role in VALID_ROLES
```

2. Valid content block types:
```python
assert block.type in set(ContentBlockType)
```

3. Stable titles for repeated parses of the same payload

4. `ParsedProviderEvent` serialization round-trips

The parser properties run over schema-conformant payloads for the parseable providers currently in scope:
- ChatGPT
- Claude Code
- Codex

I marked the generated parser properties `xfail(strict=False)` intentionally. The point of this suite is to surface real schema-conformant edge cases without immediately making the whole build permanently red while the parser layer catches up.

### Provider-specific crashlessness inputs

In `tests/unit/sources/test_parser_crashlessness.py` I replaced the direct generator usage with `st.data()` and then draw the payload for the provider under test:

```python
payload = data.draw(schema_conformant_payload(provider))
```

That fixes a structural weakness in the old test: it was parameterized by provider, but the generated payload source was still hard-coded in a way that skewed coverage toward ChatGPT-like shapes.

The no-crash contract is unchanged:
- `looks_like()` should not throw `IndexError` or `RecursionError`
- `parse()` should not throw those crash exceptions either

What changed is that the generated inputs now actually match the provider we claim to be testing.

### Provider capability exercise matrix

`polylogue/showcase/generators.py` now includes an explicit `_PROVIDER_FEATURES` map and `generate_provider_feature_exercises()`:

```python
_PROVIDER_FEATURES = {
    "chatgpt": {"tool-use", "thinking"},
    "claude-ai": {"tool-use", "thinking"},
    "claude-code": {"tool-use", "thinking"},
    "codex": {"tool-use"},
    "gemini": {"tool-use"},
}
```

The generator emits `count` exercises for each provider × feature combination, such as:
- `--provider codex --has-tool-use count`
- `--provider claude-code --has-thinking count`

That keeps showcase coverage honest about provider differences instead of treating `--has-thinking` as if it were universally meaningful.

### Focused mutation campaign

I added a `provider-parsers` campaign to `devtools/mutmut_campaign.py` targeting the parser/compaction cluster:
- `polylogue/sources/parsers/chatgpt.py`
- `polylogue/sources/parsers/claude_code_parser.py`
- `polylogue/sources/parsers/codex.py`
- `polylogue/sources/parsers/claude_index.py`
- `polylogue/pipeline/semantic_capture.py`

with the corresponding parser/property/compaction/assembly tests as its execution set.

That gives mutation testing a campaign that lines up with the semantic correctness work in this branch instead of treating parser regressions as just another part of a broader bucket.

## Testing

New or updated coverage in this branch:
- `tests/property/test_semantic_properties.py` for semantic parser invariants and provider-event round-trips
- `tests/unit/sources/test_parser_crashlessness.py` for provider-specific schema-conformant input drawing
- showcase exercise generation through `generate_provider_feature_exercises()`
- mutmut campaign targeting for parser-semantic modules

## Notes

The new property suite is intentionally partially non-blocking because its purpose is discovery as much as enforcement. `xfail(strict=False)` keeps the failures visible without pretending every newly surfaced parser edge case can be resolved in the same patch.

Kept the semantic property set small and concrete. It checks invariants that should hold across providers regardless of payload details, rather than baking in provider-specific content expectations that would make the generated tests fragile.

### DIFF
--- a/devtools/mutmut_campaign.py
+++ b/devtools/mutmut_campaign.py
@@ -286,6 +286,26 @@ CAMPAIGNS: dict[str, Campaign] = {
             "tests/unit/sources/test_parsers_drive.py",
         ),
     ),
+    "provider-parsers": Campaign(
+        name="provider-parsers",
+        description="Provider parser semantic correctness — where message extraction and compaction detection live",
+        paths_to_mutate=(
+            "polylogue/sources/parsers/chatgpt.py",
+            "polylogue/sources/parsers/claude_code_parser.py",
+            "polylogue/sources/parsers/codex.py",
+            "polylogue/sources/parsers/claude_index.py",
+            "polylogue/pipeline/semantic_capture.py",
+        ),
+        tests=(
+            "tests/unit/sources/test_parsers_chatgpt.py",
+            "tests/unit/sources/test_parsers_codex.py",
+            "tests/unit/sources/test_parsers_props.py",
+            "tests/unit/sources/test_parser_crashlessness.py",
+            "tests/unit/sources/test_compaction.py",
+            "tests/unit/sources/test_assembly.py",
+        ),
+        notes=("Focused on the parser modules where semantic correctness is most critical.",),
+    ),
     "providers-semantics": Campaign(
         name="providers-semantics",
         description="Provider semantic extraction, harmonization, and viewport contracts",
--- a/polylogue/showcase/generators.py
+++ b/polylogue/showcase/generators.py
@@ -289,6 +289,33 @@ def generate_schema_exercises() -> list[Exercise]:
     return exercises
 
 
+_PROVIDER_FEATURES: dict[str, set[str]] = {
+    "chatgpt": {"tool-use", "thinking"},
+    "claude-ai": {"tool-use", "thinking"},
+    "claude-code": {"tool-use", "thinking"},
+    "codex": {"tool-use"},
+    "gemini": {"tool-use"},
+}
+
+
+def generate_provider_feature_exercises() -> list[Exercise]:
+    """Generate provider × content-type cross-product exercises."""
+    exercises: list[Exercise] = []
+    for provider, features in _PROVIDER_FEATURES.items():
+        for feature in sorted(features):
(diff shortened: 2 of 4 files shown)


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 5 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550).

Sinity:
@codex review

PR #205, opened 2026-04-16
chore: restore mypy --strict CI gate and clean the mechanical tier

Sinity:
## Summary

Re-introduce the `mypy --strict` type-checking gate that was removed in #174, and clean the mechanical-tier debt in the same PR so the exclude list starts at a realistic size.

## Problem

After #174 the codebase had 1,764 strict errors across 232 files and no CI gate. Issue #188 set the direction but the work wasn't done. Two kinds of errors existed:

- **Mechanical tier** (~500 errors): missing annotations, missing generic parameters, `Any`-typed returns. These are safe to fix in bulk.
- **Design tier** (~1,200 errors): `attr-defined` (741, mostly the mixin-attribute pattern), `arg-type` (224, often real bugs), `union-attr`, `call-overload`. Each needs per-module thought and belongs in focused PRs.

### `typecheck` CI job

Runs `uv run mypy polylogue/` on every PR. Regressions fail CI:
- new error in a covered file
- new file that doesn't pass strict

### mypy version pinned

`mypy>=1.17,<1.18` in the dev extras, matching the Nix devshell (1.17.1). Local and CI align. `uv.lock` regenerated (1.19.1 → 1.17.1). No more silent drift when a new mypy minor surfaces new errors.

### `devtools verify` runs mypy

Between `ruff check` and `render-all`. Cold ~25s, warm ~1s. Pre-push hook now catches type regressions before CI.

### Mechanical cleanup in this PR

Cleaned 53 files that only had mechanical errors. The exclude list dropped from 232 to 179 entries; gate coverage went from 332 → 385 files. Six files that *looked* mechanical but surfaced real issues on annotation stayed in exclude (details in the second commit).

### `CONTRIBUTING.md`

Adds a "Widening Type-Check Coverage" section: pick a file, remove its exclude entry, fix errors, verify, commit.

### CI cleanup

Removed the "typecheck is tracked debt" comment in `.github/workflows/ci.yml`.

## Follow-up

The 179 files still in exclude fall into clusters that deserve dedicated PRs:

- **Async/sync backend** (most `no-untyped-call` cascades): blocked by #190.
- **Mixin-attribute Protocol declarations** (`attr-defined`): per-mixin PRs, each adding a `Protocol` for the consumer interface.
- **`arg-type` / `union-attr`**: often real bugs — investigate per-case.

## Verification

```
nix develop -c mypy polylogue/
# Success: no issues found in 385 source files

nix develop -c devtools verify
[...]

PR #206, opened 2026-04-17
chore: clean 4 mypy-excluded files with shadowing/untyped-self patterns

Sinity:
## Summary

- Four files held out of the #205 mechanical tier surfaced non-trivial annotation issues. This PR resolves each and removes them from the mypy `exclude`.
- Gate grows from 385 → 389 files checked under `mypy --strict`.

## Problem

During the #205 mechanical sweep I skipped four files where straight `dict[str, Any]` / `self: Any` sprays either failed or exposed design cruft:

1. `polylogue/lib/filters.py` — `ConversationFilter.list()` shadows the builtin in its own return annotation. `async def list(self) -> list[Conversation]:` fails because `list` in that scope resolves to the method, not the builtin.
2. `polylogue/schemas/operator_annotations.py` — JSON-schema walkers with untyped `dict` parameters.
3. `polylogue/schemas/synthetic/build_records.py` and `build_wire_formats.py` — module-level helpers bound to `SyntheticCorpus` via class-body assignment (`_generate_tree_json = _generate_tree_json`), with untyped `self` and `dict` parameters.

## Solution

- `filters.py`: reference `builtins.list[Conversation]` / `builtins.list[ConversationSummary]` for the three methods that return lists and share the name. Add typed property returns for `_since_date`, `_until_date`, `_continuation`, `_sidechain`, `_has_branches`. Localized `# type: ignore[no-any-return]` on three call-sites that cross into `polylogue.lib.query_plan_execution` (still in the exclude list — these markers drop naturally when that module is cleaned).
- `operator_annotations.py`: adopt `dict[str, Any]` consistent with sibling `polylogue/schemas/audit_walkers.py` and `audit_checks.py`.
- `build_records.py`, `build_wire_formats.py`: adopt `self: Any` pattern already present in `polylogue/schemas/synthetic/runtime.py`, plus `dict[str, Any]` for JSON payloads.

## Verification

Commands run:
- `nix develop -c mypy --config-file pyproject.toml polylogue/` — `Success: no issues found in 389 source files`
- `nix develop -c devtools verify --quick` — all checks passed (format, lint, mypy, render-all --check)
- `git push` — pre-push hook (`devtools verify --quick`) passed

Full `devtools verify` including pytest deferred to CI (local btrfs metadata pressure, unrelated to this change).


Sinity:
@codex review

PR #207, opened 2026-04-17
chore: clean 3 single-error mypy files, add types-PyYAML stubs

Sinity:
## Summary

Three excluded files each carrying a single mypy error — cleaned and promoted into the gate. Gate grows from 389 → 392 files under `mypy --strict`.

## Problem

From running mypy against the full exclude list, 29 files had exactly one reported error. These three resolved without cascade work:

1. `polylogue/cli/commands/check.py:96` — `_make_schema_progress_callback()` wrapper missing return annotation, but the underlying impl in `check_support.py` explicitly returns `Any`.
2. `polylogue/mcp/server.py:63` — `extract_fenced_code=lambda text, language="": _extract_fenced_code(text, language)`. Mypy could not infer the lambda against `Callable[[str, str], list[dict[str, str]]]` when a default argument was involved.
3. `polylogue/rendering/formatting.py:107` — `import yaml` raised `import-untyped` because PyYAML stubs were never declared.

## Solution

- `check.py`: annotate the wrapper as `-> Any` to propagate the impl's declared return.
- `server.py`: drop the forwarding lambda and pass `_extract_fenced_code` directly — signatures match exactly.
- `pyproject.toml`: add `types-PyYAML>=6.0` to the `dev` extra. Re-lock `uv.lock`. The source file needed no change.

## Verification

- `nix develop -c mypy --config-file pyproject.toml polylogue/` — `Success: no issues found in 392 source files`
- `nix develop -c devtools verify --quick` — all checks passed (format, lint, mypy, render-all --check)

### DIFF
--- a/polylogue/cli/commands/check.py
+++ b/polylogue/cli/commands/check.py
@@ -93,7 +93,7 @@ def check_command(
     render_plain_output(env, result, options)
 
 
-def _make_schema_progress_callback():
+def _make_schema_progress_callback() -> Any:
     return _make_schema_progress_callback_impl()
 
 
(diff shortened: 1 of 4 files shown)


Sinity:
@codex review

PR #208, opened 2026-04-17
fix: tighten None/Optional narrowing in 3 mypy-excluded files

Sinity:
## Summary

Three excluded files, each carrying a single mypy error that exposed a real narrowing defect (not a missing annotation). All three are now clean and promoted into the gate. Gate grows from 392 → 395 files under \`mypy --strict\`.

## Problem

- **\`polylogue/lib/phase_extraction.py:66\`** — \`(start_time or end_time).date() if (start_time or end_time) else None\`. Mypy evaluates the method call first and cannot tie the \`if\`-guard to the \`or\`-chain; \`.date()\` resolves on \`datetime | None\`.
- **\`polylogue/lib/action_event_parsing.py:106\`** — \`category = classified if semantic_category in (None, ToolCategory.OTHER) else semantic_category\`. In the else branch mypy still sees \`ToolCategory | None\` because \`x in (...)\` is not a narrowing form; but \`ToolCall.category\` is non-Optional.
- **\`polylogue/ui/facade_prompts.py:50\`** — \`ui_error_cls: type[Exception]\` with \`raise ui_error_cls(..., prompt_topic=prompt_topic)\`. \`Exception\` does not accept \`prompt_topic=\`; the real constructor signature is \`UIError\`'s.

## Solution

- **phase_extraction**: bind \`ref_time = start_time or end_time\` before the dataclass construction; use \`ref_time.date() if ref_time else None\`. Same runtime behavior, mypy-legible narrowing.
- **action_event_parsing**: rewrite as \`if semantic_category is None or semantic_category is ToolCategory.OTHER\` so mypy narrows \`semantic_category\` to \`ToolCategory\` in the else branch.
- **facade_prompts**: narrow every \`ui_error_cls\` in the module to \`type[UIError]\`. UIError is imported under \`TYPE_CHECKING\` to avoid the runtime cycle with \`polylogue.ui.facade\`. Every caller already passes UIError.

## Verification

- \`nix develop -c mypy --config-file pyproject.toml polylogue/ui/facade_prompts.py polylogue/lib/phase_extraction.py polylogue/lib/action_event_parsing.py\` — \`Success: no issues found in 3 source files\`
- \`nix develop -c devtools verify --quick\` — format, lint, mypy, render-all --check all passed
- Pre-push hook (\`devtools verify --quick\`) passed

Full \`devtools verify\` (with pytest) deferred to CI.

### DIFF
--- a/polylogue/lib/action_event_parsing.py
+++ b/polylogue/lib/action_event_parsing.py
@@ -82,7 +82,10 @@ def build_tool_calls_from_content_blocks(
         normalized_input = _normalized_mapping(block.get("tool_input"))
         semantic_category = _tool_category_from_semantic(block.get("semantic_type"))
         classified_category = classify_tool(name, normalized_input)
-        category = classified_category if semantic_category in (None, ToolCategory.OTHER) else semantic_category
+        if semantic_category is None or semantic_category is ToolCategory.OTHER:
+            category = classified_category
+        else:
+            category = semantic_category
         raw = {
             "block_id": block.get("block_id"),
             "block_index": block.get("block_index"),
--- a/polylogue/lib/phase_extraction.py
+++ b/polylogue/lib/phase_extraction.py
@@ -60,10 +60,11 @@ def _build_phase(
     if start_time and end_time:
         duration_ms = max(int((end_time - start_time).total_seconds() * 1000), 0)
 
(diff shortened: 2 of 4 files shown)


Sinity:
@codex review

PR #210, opened 2026-04-17
fix: enum coercion at call sites in 3 mypy-excluded parsers/validators

Sinity:
## Summary

Three excluded files each passing a bare \`str\` where the declared type is a Literal/Enum. Gate grows from 398 → 401 under \`mypy --strict\`.

## Problem

- **\`polylogue/sources/parsers/claude_common.py:98\`** and **\`polylogue/sources/parsers/claude_code_parser.py:71\`** — \`ParsedContentBlock(type="text", ...)\`. The field is declared \`type: ContentBlockType\`. Pydantic has a \`coerce_type\` validator that accepts str at runtime, but mypy sees the annotated field type.
- **\`polylogue/schemas/validator.py:81\`** — \`extract_payload_samples\` takes \`sample_granularity: Literal["document", "record"]\`. The caller only checked \`isinstance(granularity, str)\`; a schema with an arbitrary string value would have propagated downstream.

## Solution

- **Parsers**: pass \`ContentBlockType.TEXT\` directly instead of \`"text"\`. Add \`ContentBlockType\` to the existing \`polylogue.types\` import in \`claude_code_parser.py\`; \`claude_common.py\` didn't import from \`polylogue.types\` previously — add that import too.
- **validator**: constrain to the two known Literal values. Accept \`raw_granularity == "record"\` or \`"document"\`; anything else (including non-str) falls through to the provider-based default. Small behaviour tightening: a schema carrying a typo like \`"records"\` now takes the provider-based default instead of passing through to \`extract_payload_samples\` where it would have been accepted as a str and broken downstream.

## Verification

- \`nix develop -c mypy --config-file pyproject.toml polylogue/sources/parsers/claude_code_parser.py polylogue/sources/parsers/claude_common.py polylogue/schemas/validator.py\` — \`Success: no issues found in 3 source files\`
- \`nix develop -c devtools verify --quick\` — format, lint, mypy, render-all --check all passed
- Pre-push hook (\`devtools verify --quick\`) passed

Full \`devtools verify\` (with pytest) deferred to CI.


Sinity:
@codex review

PR #211, opened 2026-04-17
fix: NoReturn, Provider coercion, dead union in 3 mypy-excluded files

Sinity:
## Summary

Three excluded files each carrying one mypy error that surfaced a real defect. Gate grows from 401 → 404 under `mypy --strict`.

## Problem

- `polylogue/schemas/unified_models.py` / `unified_adapters.py:20` — `_missing_role()` unconditionally `raise ValueError(...)` but was annotated `-> str`. Call sites like `meta.role or _missing_role()` ended up typed `Role | str`, which failed the assignment to `HarmonizedMessage.role: Role`.
- `polylogue/pipeline/services/validation_runtime.py:86` — `payload_provider` was initialized from `stored_payload_provider` (`str | None`) and later reassigned to `envelope.provider` (`Provider`). In the except branch it stayed as `str | None`, but `_ValidationOutcome.payload_provider: Provider | None`.
- `polylogue/rendering/core_formatter.py:69` — the list comp `[b.model_dump(mode="json") if hasattr(b, "model_dump") else b ...]` produced `list[dict[str, Any] | ContentBlockRecord]`. But `MessageRecord.content_blocks: list[ContentBlockRecord]` — every element is a pydantic model and `hasattr(b, "model_dump")` is always True; the else was dead code at both type and runtime level.

## Solution

- **unified_models**: re-annotate `_missing_role() -> NoReturn`. No call-site change needed; `Role | NoReturn` narrows to `Role`.
- **validation_runtime**: coerce `stored_payload_provider` via `Provider.from_string` at init so the type is `Provider | None` stably across try/except.
- **core_formatter**: drop the hasattr guard and the dead else branch.

## Verification

- `nix develop -c mypy --config-file pyproject.toml polylogue/schemas/unified_adapters.py polylogue/pipeline/services/validation_runtime.py polylogue/rendering/core_formatter.py polylogue/schemas/unified_models.py` — `Success: no issues found in 4 source files`
- `nix develop -c devtools verify --quick` — format, lint, mypy, render-all --check all passed


Sinity:
@codex review

PR #212, opened 2026-04-17
fix: narrow object-valued Row/sentinel access in 3 storage files

Sinity:
## Summary

Three more single-error excluded files in the storage layer. Each was passing an `object` or custom-sentinel value to `int()` / `float()`; mypy rejects, the try/except at runtime absorbed it silently. Gate grows from 404 → 407 under `mypy --strict`.

## Problem

- `polylogue/storage/search_query_support.py:18` — `float(sort_key)` where `sort_key: object`. The try/except catches `TypeError` at runtime but mypy requires a pre-narrow.
- `polylogue/storage/session_product_status.py:254` — `int(row[0] or 0)` where `row[0]` is `object | Any` from a sqlite3 Row.
- `polylogue/storage/backends/queries/raw_state.py:68` — `int(state.validation_drift_count)` inside `if state.validation_drift_count is not UNSET`. `UNSET` is an instance of `_RawStateUnset`, not a Sentinel/Literal, so `is not UNSET` doesn't narrow the `_RawStateUnset` arm out of the union.

## Solution

- **search_query_support**: `isinstance(sort_key, (int, float, str))` guard before `float()`; same fallback path as the try/except, but explicit at the type level.
- **session_product_status**: bind `value = row[0]`, guard `isinstance(value, (int, float, str))` before `int()`, return 0 otherwise.
- **raw_state**: explicit `isinstance(drift_count, _RawStateUnset)` narrowing and `or 0` for None; import `_RawStateUnset` alongside `UNSET`.

## Verification

- `nix develop -c mypy --config-file pyproject.toml polylogue/storage/search_query_support.py polylogue/storage/session_product_status.py polylogue/storage/backends/queries/raw_state.py` — `Success: no issues found in 3 source files`
- `nix develop -c devtools verify --quick` — format, lint, mypy, render-all --check all passed


Sinity:
@codex review

Sinity:
@codex review

PR #307, opened 2026-04-21
test: stabilize ingestion interrupt timing

Sinity:
## Summary
- Replace fixed sleep timing in the interrupted-ingestion integration test with a wait for persisted message rows.
- Reuse a small helper to finish or kill interrupted subprocesses consistently.

## Problem
- Python 3.12 CI exposed a timing race in tests/integration/test_ingestion_interrupts.py: the test sent SIGINT after a fixed delay and then assumed the SQLite database had already been created.
- Slow startup or scheduler contention can make that assertion fail before the pipeline has reached persistence, even though the production code is unrelated.

## Solution
- Poll the isolated SQLite database until the messages table contains persisted rows before sending SIGINT.
- Preserve the corruption/resume contract while removing the startup-speed assumption.

## Verification
- ruff check tests/integration/test_ingestion_interrupts.py
- mypy tests/integration/test_ingestion_interrupts.py
- pytest -q tests/integration/test_ingestion_interrupts.py
- git push hook: devtools verify --quick

Ref #300.

### DIFF
--- a/tests/integration/test_ingestion_interrupts.py
+++ b/tests/integration/test_ingestion_interrupts.py
@@ -20,6 +20,46 @@ from tests.infra.cli_subprocess import setup_isolated_workspace
 pytestmark = [pytest.mark.integration, pytest.mark.chaos]
 
 
+def _message_count_if_ready(db_path: Path) -> int | None:
+    if not db_path.exists():
+        return None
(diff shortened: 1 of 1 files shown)


Sinity:

PR #347, opened 2026-04-22
test: capture raw quarantine regressions

### LINKED ISSUE #328: test: capture raw quarantine regression cases
## Context

Live archive repair split stale validation-state cases from two legitimate quarantine failures:

- malformed JSONL that should preserve the decode failure as a real quarantine case
- zero-length Codex/session export that should remain an explicit decode failure

The behavior is important enough to deserve deterministic local regressions rather than staying only as operator memory.

## Scope

Add synthetic/raw regression fixtures for explicit quarantine behavior around malformed and empty raw exports.

## Acceptance criteria

- Malformed JSONL input produces a preserved, inspectable decode failure rather than stale validation-state ambiguity.
- Zero-length raw input produces an explicit decode/quarantine failure.
- These cases do not regress parser tolerance for recoverable malformed-but-provider-shaped payloads.
- Tests verify both persisted state and user-facing diagnostic/reporting behavior where practical.
- `devtools verify` passes before merge.

Ref #196

Sinity:
## Summary

Closes #328.

Adds focused regression coverage for legitimate raw quarantine failures seen during archive repair: zero-length raw blobs and malformed JSONL records.

## Problem

The existing quarantine coverage mostly exercised `ingest_record` result objects and schema-verifier internals. The deployment-readiness gap was proving that these failure shapes persist through the raw validation path and remain visible in the operator-facing `doctor --schemas --schema-quarantine-malformed` output.

## Solution

- Added validation-flow coverage in `tests/unit/pipeline/test_quarantine_fixtures.py` proving zero-length and malformed JSONL records persist as failed validation rows with `parse_error`, stay unparsed, and classify as quarantined.
- Added a real CLI regression in `tests/unit/cli/test_check.py` proving `doctor --schemas --schema-quarantine-malformed` reports `decode_errors=1 quarantined=1` and persists the quarantine state for malformed JSONL.
- Kept the existing recoverable malformed JSONL validation-off behavior unchanged.

## Verification

```bash
pytest tests/unit/pipeline/test_quarantine_fixtures.py -q
pytest tests/unit/cli/test_check.py -q -k quarantine
ruff format --check tests/unit/pipeline/test_quarantine_fixtures.py tests/unit/cli/test_check.py
ruff check tests/unit/pipeline/test_quarantine_fixtures.py tests/unit/cli/test_check.py
devtools verify

git push -u origin feature/test/raw-quarantine-regressions
# pre-push: devtools verify --quick
```

### DIFF
--- a/tests/unit/cli/test_check.py
+++ b/tests/unit/cli/test_check.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+from datetime import datetime, timezone
 from pathlib import Path
 from unittest.mock import ANY, patch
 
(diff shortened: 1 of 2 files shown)


Sinity:
@codex review

Issue #448, opened 2026-04-26
feat(proof): commit high-value witness snapshots


Sinity:
## Problem

`tests/witnesses/` currently contains exactly **one** committed witness (`golden-tool-use-json.witness.json`). The `WitnessMetadata` lifecycle (origin, minimization, privacy, validation) is built out — but there's only one example.

The audit identified five high-value places where committing witness snapshots would catch real regressions early:

1. **MCP tool schema** — the wire schema each MCP tool publishes. Drift in `polylogue/mcp/server_tools.py` would invalidate clients silently today.

2. **TUI mutation under archive change** — the dashboard's behavior when the archive mutates beneath it. Hard to test conventionally; a snapshot of pre/post states is concrete.

3. **Site publication output** — the generated static site for a representative archive. CSS/template drift detected immediately. (Aligned with #417.)

4. **BlobStore durability** — the on-disk shape of the content-addressed blob store. Schema changes to the blob layout would surface here.

5. **Browser-extension message sequencing** — the JSON-RPC transcript between extension and receiver. Flapping behavior in popup/content scripts becomes visible.

## Target

For each of the five surfaces:

1. Generate a representative input (synthetic, not real archive data).
2. Capture the produced output as a witness snapshot under `tests/witnesses/`.
3. Author the witness JSON with full lifecycle metadata (origin, minimization, privacy, preserved facts).
4. Add a regression test that re-runs the input and diffs against the witness.
5. Wire each witness into `polylogue/proof/witnesses.py`'s loader so `committed_witness_errors()` validates the lifecycle.

## Acceptance criteria

- 5 new committed witnesses under `tests/witnesses/`.
- Each witness validates per `WitnessMetadata.validation_errors() == ()`.
- Each surface has a regression test that loads the witness and asserts equality (or asserts a structural invariant if exact equality is too brittle).
- `committed_witness_errors()` returns empty over the realized witness set.
- `docs/verification-catalog.md` references the new witness coverage.
- Coordinates with #421 (product-surface obligation families) — each new witness can become a concrete `EvidenceEnvelope` for that surface's proof obligation.

## Cross-reference

Several of these surfaces also lack proof subjects per the proof-coverage audit. The plan is:
- #421 adds the subject builders.
- This issue commits the witnesses.
- The two together close the surface-coverage gap that the audit flagged.

## Related

- #421 — product-surface obligation families.
- #401 — post-renewal consolidation umbrella.
- #429 — pattern source (evidence-driven specification).

[...]

PR #459, opened 2026-04-26
feat(proof): commit MCP tool schema and BlobStore witnesses

Sinity:
## Summary

Ref #448. Lands the first two committed witnesses identified by the
proof-coverage-audit (swarm 2026-04-26). Remaining three (TUI mutation,
site publication, browser-extension messaging) are deferred as separate
work — they each need a substantial harness, which the issue notes.

## Problem

`tests/witnesses/` had exactly one committed witness. The
`WitnessMetadata` lifecycle infrastructure was built out but underused.
Two surfaces with high regression value were missing snapshots:

- The MCP tool wire schema — drift here silently invalidates client
  integrations that consume the published JSON Schema.
- The BlobStore on-disk layout — any change to the addressing scheme
  or directory split silently invalidates every blob ever written.

## Solution

- ``mcp.tool-schemas`` — snapshot of ``build_server()._tool_manager._tools``
  reduced to ``{name, parameters}`` per tool. Regression test diffs the
  current registration against the committed JSON.
- ``storage.blob-store-layout`` — three synthetic fixtures with their
  base64-encoded content bytes, expected SHA-256, expected size, and
  expected on-disk relative path. Regression test re-writes them into a
  fresh ``BlobStore`` and asserts every field.

Both witnesses validate per ``WitnessMetadata.validation_errors() == ()``
and link the test file in their provenance.

## Verification

- ``devtools verify`` (full, with pytest): all checks pass.
- 6 new tests across both surfaces.

### DIFF
--- a/tests/data/witnesses/blob-store-layout.json
+++ b/tests/data/witnesses/blob-store-layout.json
@@ -1,0 +1,27 @@
+{
+  "scheme": "sha256",
+  "directory_split": [2, 62],
+  "fixtures": [
+    {
+      "name": "alpha",
+      "content_bytes_b64": "YWxwaGEK",
+      "expected_sha256": "b6a98d9ce9a2d9149288fa3df42d377c3e42737afdcdaf714e33c0a100b51060",
+      "expected_size": 6,
+      "expected_relative_path": "b6/a98d9ce9a2d9149288fa3df42d377c3e42737afdcdaf714e33c0a100b51060"
+    },
+    {
+      "name": "beta",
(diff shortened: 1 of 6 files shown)


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 1 finding(s) confirmed still present → tracked in [#545](https://github.com/Sinity/polylogue/issues/545).

PR #460, opened 2026-04-26
fix(hooks): pre-push distinguishes env failure from verify failure

### LINKED ISSUE #423: [bug] pre-push hook fails with misleading message when invoked outside the polylogue devshell
## Actual behavior

\`.githooks/pre-push\` runs \`devtools verify --quick\`. When invoked from a shell that hasn't entered polylogue's devshell (e.g. an agent or terminal whose direnv is loaded for a sibling repo, leaving the shell on a different Python interpreter), \`devtools verify --quick\` aborts with a Python ABI traceback ending in:

\`\`\`
ModuleNotFoundError: No module named 'pydantic_core._pydantic_core'

verify: FAILED — fix before pushing
\`\`\`

The single-line summary "verify: FAILED — fix before pushing" reads as if the repo's verification surface (lint, mypy, render-all) actually failed against the diff being pushed. The reality is that \`devtools\`/pydantic just couldn't import in the active environment — nothing about the push or the diff is wrong.

The push is correctly blocked, but the user/agent then chases a non-existent code defect.

## Expected behavior

The hook should distinguish "verify rejected your changes" from "verify could not even start in this environment" and print different guidance. For the latter case, suggest \`nix develop --command git push …\` (or equivalent direnv re-eval) so the user knows the repo isn't broken — their shell just isn't pointing at polylogue's Python env.

## Reproduction or trigger

\`\`\`bash
# In a terminal whose direnv is loaded for a sibling repo with a different Python env
# (in my case: lynchpin's Python 3.13 env, polylogue wants Python 3.12).
cd /realm/project/polylogue
git push origin --delete <some-stale-branch>
# → traceback + "verify: FAILED — fix before pushing"

# Same command via polylogue's own devshell works fine:
nix develop --command git push origin --delete <some-stale-branch>
# → verify passes, push succeeds
\`\`\`

Triggered most reliably for agents that cd between repos in one long-lived shell, where direnv only loaded once for the first repo. Also reproduces for humans who land in polylogue from a non-direnv terminal.

## Suspected scope

\`.githooks/pre-push\` — short script that just calls \`devtools verify --quick\`. The mitigation is in the hook itself, not in \`devtools\`:

- Detect that \`devtools verify --quick\` aborted on import/setup (non-zero exit before any verification step ran), vs failed on a verification step.
- For the import/setup-aborted case, print an actionable hint: "polylogue's devshell isn't active — try \`nix develop --command git push …\` or run \`direnv allow\`."
- Optionally: have the hook re-exec itself inside \`nix develop --command\` if it detects it isn't in one (would silently fix the case for agents/humans who don't know about the convention).

The auto-reexec option is more aggressive but eliminates the footgun entirely. Either is better than the current uniform "FAILED — fix" message that points at code rather than env.

## Severity

Low-moderate. Doesn't corrupt anything; the hook still does its job (no bad pushes). The cost is wasted investigation time when the failure is misread.

## Environment

- polylogue 0.1.0+c97b6099
- Hook: \`.githooks/pre-push\` (5 lines, runs \`devtools verify --quick\`)
- Repro env: long-lived zsh with direnv loaded for sibling repo on Python 3.13; polylogue wants Python 3.12 via its own devshell.

Sinity:
## Summary

Closes #423. Pre-push now detects when ``devtools`` cannot import in the
active shell (devshell not loaded) and prints an actionable hint instead
of a Python traceback that reads like a verify failure.

## Problem

When the calling shell is loaded with a sibling repo's Python env, the
hook fires ``devtools verify --quick`` against an ABI-mismatched
pydantic_core, dies with a traceback, and prints
``verify: FAILED — fix before pushing``. The user/agent then chases a
non-existent code defect; the actual fix is to enter the devshell.

## Solution

A pre-flight ``devtools --help`` check runs first. If it exits non-zero,
the hook emits a hint suggesting ``nix develop --command git push`` or
``direnv allow``, and exits without invoking the full verify pipeline.

## Verification

- ``tests/unit/devtools/test_pre_push_hook.py`` runs the hook with a
  doctored ``PATH`` that omits devtools, asserts the hint appears, and
  asserts no Python traceback surfaces.
- ``devtools verify --quick`` (full local baseline minus pytest): all
  checks pass.

## Release Notes

* **Tests**
  * Added unit test coverage for pre-push hook validation scenarios.

* **Chores**
  * Improved pre-push hook with early environment validation to better detect missing development dependencies.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -2,8 +2,30 @@
 # Pre-push gate: full verification baseline (quick mode).
 # Runs format + lint + render-all --check. Tests run in CI.
 # Installed automatically by the devshell (shellHook in flake.nix).
+#
+# Distinguishes "verify rejected your changes" from "devtools could not even
+# import in this shell" — the latter usually means polylogue's devshell isn't
+# loaded in the current terminal (see #423).
 
 set -euo pipefail
+
+# Pre-flight: confirm devtools imports cleanly. If it doesn't, the active
+# shell isn't pointing at polylogue's Python env, and a verify failure here
+# is environmental, not a defect in the diff being pushed.
+if ! devtools --help >/dev/null 2>&1; then
+    cat >&2 <<'MSG'
+pre-push: devtools is not importable in this shell.
+
+This usually means polylogue's devshell isn't loaded — the push isn't being
+rejected for any defect in your diff. To run the gate against the diff:
+
+    nix develop --command git push <args...>
+
+or re-enter the project directory under direnv (run `direnv allow` once if
+you haven't), and try the push again.
+MSG
+    exit 1
+fi
 
 echo "pre-push: running quick verification baseline" >&2
 devtools verify --quick
--- a/tests/unit/devtools/test_pre_push_hook.py
+++ b/tests/unit/devtools/test_pre_push_hook.py
@@ -1,0 +1,50 @@
+"""Pre-push hook env detection (#423).
+
+Simulates the hook's pre-flight by running it with a doctored ``PATH`` that
+either provides or omits ``devtools``. The successful path delegates to
+``devtools verify --quick``; the failed path emits an actionable hint about
+the devshell, instead of a Python traceback.
+"""
+
+from __future__ import annotations
+
+import os
+import shutil
+import subprocess
+from pathlib import Path
+
+HOOK_PATH = Path(__file__).resolve().parents[3] / ".githooks" / "pre-push"
+
+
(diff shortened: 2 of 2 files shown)


Sinity:
🤖 Post-hoc audit note: this PR received bot review feedback verified against current HEAD as part of [#544](https://github.com/Sinity/polylogue/issues/544). 2 finding(s) confirmed still present → tracked in [#550](https://github.com/Sinity/polylogue/issues/550).

PR #461, opened 2026-04-26
fix(pipeline): invert embed dependency, pipeline imports substrate not cli

Sinity:
## Summary

Closes #430. Moves embedding execution and stats payload helpers out of
``polylogue/cli/`` and into a new substrate module under
``polylogue/lib/embeddings/``. The CLI modules become thin click-formatting
wrappers; pipeline calls substrate directly without the ``_PlainUI`` /
``_EmbedEnv`` shims that proxied a fake CLI surface back into pipeline.

## Problem

``polylogue/pipeline/run_stages.py`` imported ``embed_batch`` /
``embed_single`` / ``show_embedding_stats`` from ``polylogue/cli/``,
inverting the documented placement rule and creating a true substrate↔
surface cycle that ``devtools verify-cluster-cohesion`` flags. The
imports were lazy (inside function bodies) so they didn't crash at module
load, but they bound pipeline to a CLI shape and required pipeline to
fabricate a fake UI object to satisfy the CLI's protocol.

## Solution

- ``polylogue/lib/embeddings/runtime.py`` — click-free
  ``embed_conversation_sync`` + ``iter_pending_conversations`` +
  ``EmbedConversationOutcome``.
- ``polylogue/lib/embeddings/stats.py`` — ``embedding_status_payload``
  and payload type definitions (was already click-free; just moved).
- ``polylogue/cli/embed_runtime.py`` / ``embed_stats.py`` — thin click
  wrappers composing substrate outcomes into the existing CLI surface;
  no behavior change to ``polylogue run embed`` output.
- ``polylogue/pipeline/run_stages.py`` — imports substrate; emits its
  own ``click.echo`` for the pipeline-stage user output. ``_PlainUI``
  and ``_EmbedEnv`` shims removed.

## Verification

- ``tests/unit/pipeline/test_no_cli_imports.py`` (new) — regression
  guard asserting no ``polylogue/pipeline/**/*.py`` imports from
  ``polylogue/cli/**``.
- ``tests/unit/pipeline/test_run_stages_runtime.py`` — updated to patch
  substrate locations rather than CLI re-exports.
- ``devtools verify`` (full, including pytest): all checks pass.

## Acceptance criteria checklist

- [x] No ``polylogue/pipeline/**/*.py`` imports from ``polylogue/cli/**``.
- [x] Embedding behavior is callable from pipeline without going through ``cli``.
- [x] Existing CLI behavior unchanged.
[...]

PR #462, opened 2026-04-26
fix(export): apply --message-role / --dialogue-only to single export

### LINKED ISSUE #406: fix(export): apply message-role filters to exports
## Repro

\`\`\`
polylogue --message-role user export <conversation-id> --format json
\`\`\`

returns the full message list including assistant + system + unknown. The filter is silently ignored on \`export\`.

## Expected

\`--message-role\` should filter the \`messages\` array in the export output, same as it does on \`list\`. Currently the only workaround is to dump full JSON and post-filter in Python:

\`\`\`python
import json, subprocess
d = json.loads(subprocess.run(['polylogue', 'export', sid, '--format', 'json'], capture_output=True).stdout)
users = [m for m in d['messages'] if m['role'] == 'user']
\`\`\`

## Why it matters

When extracting user prose across many conversations (e.g. for retrospective intent analysis), having to round-trip every message through Python before discarding 90 % is wasteful. The flag exists; just needs to be wired through to the export path.

## Related

Came up while extracting user inputs for thoughtspace-related sessions in #405 territory.

Sinity:
## Summary

Closes #406. Stacked on #461.

## Problem

The single-conversation ``export`` command silently ignored root-level
``--message-role`` and ``--dialogue-only`` flags. Users had to dump full
JSON and post-filter in Python to extract user-only prose:

    polylogue --message-role user export <id> --format json
    # returns full message list, ignoring --message-role

## Solution

Wires the flags through the same helper the verbs use:
``conversation.with_roles(roles)``. Honors ``--message-role`` (one or
more roles); when ``--dialogue-only`` is set without an explicit role
filter, defaults to user+assistant.

## Verification

``tests/unit/cli/test_export.py`` covers both paths against a synthetic
conversation with mixed roles (user / assistant / system).
``devtools verify`` (full): all checks pass.

### DIFF
--- a/docs/verification-catalog.md
+++ b/docs/verification-catalog.md
@@ -58,7 +58,7 @@ This catalog is generated from the proof-obligation kernel. It records subjects,
 | `polylogue dashboard` | `polylogue/cli/commands/dashboard.py:10` `polylogue.cli.commands.dashboard.dashboard_command` |
 | `polylogue delete` | `polylogue/cli/query_verbs.py:89` `polylogue.cli.query_verbs.delete_verb` |
 | `polylogue doctor` | `polylogue/cli/commands/check.py:31` `polylogue.cli.commands.check.check_command` |
-| `polylogue export` | `polylogue/cli/commands/export.py:13` `polylogue.cli.commands.export.export_command` |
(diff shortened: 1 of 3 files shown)


Sinity:
@codex review

Sinity:
@codex review

Issue #405, opened 2026-04-25
feat(messages): detect pasted content in user messages


Sinity:
## Problem

User messages in the archive frequently contain large paste content (prior chatlogs being handed forward, AI tool output, code blocks, command output). When trying to surface what the user *actually wrote* — e.g. extracting their original framing of a project or their captured intent across many sessions — these pastes dominate by byte volume and bury the typed prose.

Concrete example: dumping all user messages across 35 sessions in one project yielded 906 KB total. The top 10 messages accounted for 69 % of that and were chatlog pastes (`> previous chatlog attached below`, `> showing you last chatlog`, etc.). The remaining ~280 KB was real user prose.

A coarse heuristic filter (length > 4 KB + content-shape markers) caught the worst offenders, but doesn't solve the general case: small embedded code pastes, tool-output pastes, structured documents pasted into otherwise-typed messages.

## What success looks like

Per-message metadata distinguishing typed-by-human content from pasted/inserted content. Enables:

- A `polylogue --typed-only` filter that strips paste segments
- A "what did the user actually write" query mode for analysis / retrospective intent extraction
- Stats grouped by paste-ratio for understanding session shapes

Sinity:
## Detection method (portable)

Content-shape classifier as default — works on any provider's archive, no external data needed:

- **Length cutoff**: > ~4000 chars near-certainly contains paste; < 200 chars near-certainly typed; in between, content-shape decides.
- **Provider-specific markers**: `<task-notification>`, `● Bash(`, claude-code tool-output prefixes, shell-prompt patterns (`~ ❯`, `/realm/...`), bracketed system reminders, fenced code with shell language hint.
- **Resumption pastes**: messages starting with `> previous chatlog`, `showing you last chatlog`, `tail end of`, `here's tail end of recent session`.
- **Structural uniformity**: long contiguous paragraphs with no typo / consistent capitalization / consistent markdown formatting are paste-shaped.
- **AI-output telltales**: \`✓\`, \`⎿\`, ` … +N lines (ctrl+o to expand)`, \"### Key insight\" headings.

Produce per-message: `{ paste_likelihood: 0..1, paste_segments: [{start, end, source: 'shape', confidence}], shape_signals: {...} }`.

Some providers (claude.ai web, codex) may also persist paste-event metadata on submit (timestamped paste actions, attachment vs typed). Worth checking what's in the raw conversation JSON before computing — if the provider already marks pastes, use that as ground truth.

[...]

PR #464, opened 2026-04-26
fix(parser): reclassify claude tool_result-only user envelopes to Role.TOOL

### LINKED ISSUE #428: fix(parser): claude-code tool_result blocks classified as user-role pollute --message-role user
## Problem

In claude-code transcripts, the \`user\` role is structurally overloaded. The Anthropic API spec requires \`tool_result\` content blocks to be carried by \`role: user\` messages — the assistant emits \`tool_use\` blocks in turn N, the runtime replies with the corresponding \`tool_result\` blocks in turn N+1, and that turn N+1 is *protocol-mandated* to be \`role: user\`. There is no way for a claude-code conversation to use any other role for tool outputs.

Polylogue's role normalization (\`polylogue/lib/roles.py:42\`) maps the *outer envelope role* — so claude-code messages whose content is exclusively \`tool_result\` blocks land under \`Role.USER\` and are indistinguishable from real typed-by-human user prose at every downstream surface (\`--message-role user\`, role-filtered export, dialogue-only retrieval, FTS attribution, stats).

This is **distinct from #405** (which is about humans pasting prior chatlogs / code blocks into typed prose — needs heuristics). The \`tool_result\` case has *deterministic, typed* markers (every block has \`type: tool_result\` and a \`tool_use_id\` linking back to the specific tool_use call). No heuristic needed; the source data already separates them.

## Concrete reproduction

Mining user intent across 3 claude-code sessions (\`5b02bfa8\`, \`bd8c0201\`, \`838282b5\`, 2026-04-25→26):

\`\`\`bash
for id in claude-code:5b02bfa8-... claude-code:bd8c0201-... claude-code:838282b5-...; do
  polylogue export \"\$id\" -f markdown
done | extract '## user' sections
\`\`\`

- **Total user-role messages:** 735, ~570 KB
- **Substantive typed user prose** (after manual filter for \`→\` tool-result prefixes, line-numbered cat outputs, \`<local-command-*>\`, \`<system-reminder>\`, fenced-only blocks): **31 messages, ~19 KB**

Roughly **96% of \`role: user\` content is tool results**, not user prose. The CLI's \`--message-role user\` filter is functionally broken for claude-code analysis workflows — every consumer has to re-implement the same ad-hoc shell post-filter.

## Where it breaks (file:line)

- \`polylogue/lib/roles.py:42\` — \`Role.normalize\` operates on the message envelope role string only; the outer role for tool_result-bearing turns is \`user\`, so they're classified \`Role.USER\` regardless of content.
- \`polylogue/sources/parsers/claude_common.py:29-30\` — when iterating segments, \`tool_use\`/\`tool_result\` blocks *are* recognized as a distinct \`seg_type\`, but get flattened into the message \`text\` field as \`json.dumps(segment)\`. The block-type metadata survives in \`content_blocks\` but the text field is what role-filtered exports surface.
- \`polylogue/sources/parsers/claude_code_parser.py:201\` — title extraction uses \`message.role == \"user\"\` and \`message.text\`; titles can therefore be derived from a tool result rather than the user's actual opening message. (Existence of \`_clean_title_text\` suggests this has been hit before; that helper is a partial workaround for the same root issue.)

## Why this is upstream of #405 and #415

- **#405** (detect pasted content) — human pasted chatlogs into typed messages. Fundamentally heuristic. Needed eventually but layered on top.
- **#415** (umbrella for authored-content extraction) — lists #405 as the message-content child. With the current roles, even a \"perfect\" #405 implementation would still attribute claude-code tool results to the user.
- **This issue** — protocol-level misclassification with deterministic markers. Should be solved first; #405's heuristics then run only over what's actually user-typed.

## Proposed shape

**Two-axis classification** for messages, replacing the single normalized role:

| | \`outer_role: user\` | \`outer_role: assistant\` |
|---|---|---|
| **content kinds: text only** | \`role=user, kind=typed\` | \`role=assistant, kind=typed\` |
| **content kinds: tool_result blocks** | \`role=tool, kind=tool_result\` | n/a (protocol-invalid) |
| **content kinds: tool_use blocks** | n/a | \`role=assistant, kind=tool_use\` (or split out as \`role=tool, kind=tool_use\`) |
| **content kinds: mixed (text + tool_result)** | split into N messages OR retain composite with per-block kinds | … |

Concrete options:

1. **Preferred — split-on-parse.** A claude-code envelope with \`role: user\` and only \`tool_result\` blocks becomes a stored message with \`role=Role.TOOL\`. A mixed envelope (rare but legal) splits into one \`Role.TOOL\` message (the tool_results) + one \`Role.USER\` message (any inline typed text the user added in the same turn). Preserves \`tool_use_id\` linkage in \`content_blocks\` metadata.
2. **Alternative — keep envelope role, add \`message_kind\` field.** \`role=Role.USER, message_kind=tool_result\` vs \`role=Role.USER, message_kind=typed\`. Less invasive; \`--message-role user\` still needs to learn about \`message_kind\` to be useful.

Option 1 aligns claude-code with the existing \`Role.TOOL\` enum (which currently never actually appears for claude-code despite being defined for \`tool_use\`/\`tool_result\` strings at \`roles.py:54\` — that branch is dead code for this provider).

## Acceptance criteria

- [ ] \`polylogue --provider claude-code --message-role user list\` returns only conversations with substantive typed user prose, not tool-result-only turns.
- [ ] \`polylogue export <claude-code-id> --message-role user -f markdown\` emits the user's typed prose, with tool results either omitted or attributed to \`role: tool\` (not user).
- [ ] Tool-result content blocks retain their \`tool_use_id\` linkage in \`content_blocks\` metadata so action-event analysis (\`area:parser\` action_event_parsing) is unaffected.
- [ ] Title extraction (\`claude_code_parser.py:201\`) selects from typed user prose, not tool results — \`_clean_title_text\` becomes redundant or simpler.
- [ ] Tests cover: (a) tool_result-only user envelope → \`role: tool\`, (b) mixed envelope → split, (c) typed-prose envelope → \`role: user\`, (d) title selection across all three.
- [ ] Backfill / re-materialization path documented (existing 1980 conversations need re-classification, not just new ones).
- [ ] Same fix considered for claude-ai (sidechain/SDK contexts where the API protocol is identical), even if claude-ai exports may flatten tool calls differently.

## Out of scope

- Heuristic detection of pasted chatlogs / code blocks in *typed* user prose. That's #405. After this lands, #405's heuristics can run over a much cleaner corpus.
- Repository filter (\#407), bulk export (\#408), export role-filter propagation (\#406) — orthogonal improvements all sharing umbrella \#415.

## Related

- Umbrella: #415
- Adjacent: #405 (paste detection), #406 (export role propagation), #407 (repo filter), #408 (bulk export), #310 (closed — content-kind decomposition predecessor).

Sinity:
## Summary

Closes #428 (conservative slice).

## Problem

Anthropic's API protocol requires ``tool_result`` blocks to be carried
by ``role: user`` envelopes. Polylogue's outer-envelope role
normalization put those under ``Role.USER``, polluting
``--message-role user`` with non-typed content. The issue's mining of
three claude-code sessions found ~96% of user-role messages were tool
results, not typed user prose.

## Solution

``polylogue/sources/parsers/claude_common.py`` gains
``reclassify_tool_result_envelope(role, content_blocks)``. The function
flips ``USER → TOOL`` when content is exclusively ``tool_result`` blocks;
mixed envelopes (text + tool_result) keep ``USER`` (the
split-on-parse alternative is left for follow-up).

Applied in both parsers:
- ``claude_common.extract_messages_from_chat_messages`` (claude web)
- ``claude_code_parser._parse_code_records`` (claude-code JSONL)

Side benefit: title extraction in ``claude_code_parser`` (which iterates
"first user message" and uses ``_clean_title_text`` as a workaround)
naturally skips tool_result envelopes now.

## Verification

- 8 new tests in
  ``tests/unit/sources/test_tool_result_role_reclassification.py``
  cover the helper directly and end-to-end through the chat-messages
  parser.
- 166 claude-related tests pass; ``devtools verify --quick`` passes.

## Follow-ups (left for #428 reopens or successor)

- Mixed envelope splitting (Option 1 from the issue).
- Schema-level ``message_kind`` field (Option 2).
- Companion improvements once #405 lands.

### DIFF
--- a/polylogue/sources/parsers/claude_code_parser.py
+++ b/polylogue/sources/parsers/claude_code_parser.py
@@ -22,7 +22,7 @@ from .base import (
     ParsedProviderEvent,
     content_blocks_from_segments,
 )
-from .claude_common import extract_message_text, normalize_timestamp
+from .claude_common import extract_message_text, normalize_timestamp, reclassify_tool_result_envelope
 
 _TAG_RE = re.compile(r"<[^>]+>")
 _WHITESPACE_RE = re.compile(r"\s+")
@@ -139,13 +139,15 @@ def _parse_code_records(records: Iterable[object], fallback_id: str) -> ParsedCo
         text = record.text_content or extract_message_text(
             record.message.get("content") if isinstance(record.message, dict) else None
         )
+        envelope_role = Role.normalize(record.role) if record.role else Role.UNKNOWN
+        content_blocks = _content_blocks_from_record(record, text)
         messages.append(
             ParsedMessage(
                 provider_message_id=str(record.uuid or f"msg-{index}"),
-                role=Role.normalize(record.role) if record.role else Role.UNKNOWN,
+                role=reclassify_tool_result_envelope(envelope_role, content_blocks),
                 text=text or "",
                 timestamp=timestamp,
-                content_blocks=_content_blocks_from_record(record, text),
+                content_blocks=content_blocks,
                 parent_message_provider_id=record.parentUuid,
             )
         )
--- a/polylogue/sources/parsers/claude_common.py
+++ b/polylogue/sources/parsers/claude_common.py
@@ -16,6 +16,27 @@ from .base import (
 )
 
 
+def reclassify_tool_result_envelope(role: Role, content_blocks: list[ParsedContentBlock]) -> Role:
+    """Reclassify a ``role: user`` envelope whose content is all ``tool_result`` to ``Role.TOOL``.
+
+    The Anthropic API protocol requires ``tool_result`` blocks to be carried by
+    ``role: user`` messages — the assistant emits ``tool_use`` blocks and the
+    runtime replies with corresponding ``tool_result`` blocks under the
+    protocol-mandated ``user`` role. Polylogue's outer-envelope role
+    normalization classifies these as ``Role.USER``, polluting
+    ``--message-role user`` filters with non-typed content.
+
+    See `#428 <https://github.com/Sinity/polylogue/issues/428>`_.
+    """
+    if role is not Role.USER:
+        return role
+    if not content_blocks:
+        return role
+    if all(block.type == ContentBlockType.TOOL_RESULT for block in content_blocks):
+        return Role.TOOL
+    return role
+
+
 def extract_text_from_segments(segments: list[object]) -> str | None:
     lines: list[str] = []
     for segment in segments:
@@ -97,6 +118,8 @@ def extract_messages_from_chat_messages(
         content_blocks = content_blocks_from_segments(raw_content) if isinstance(raw_content, list) else []
         if not content_blocks and text:
             content_blocks = [ParsedContentBlock(type=ContentBlockType.TEXT, text=text)]
+
+        role = reclassify_tool_result_envelope(role, content_blocks)
 
         if text:
             messages.append(
--- a/tests/unit/sources/test_tool_result_role_reclassification.py
+++ b/tests/unit/sources/test_tool_result_role_reclassification.py
@@ -1,0 +1,73 @@
+"""Claude tool_result envelopes are reclassified from USER to TOOL (#428).
+
+Anthropic's protocol requires ``tool_result`` blocks to be carried by
+``role: user`` envelopes. Polylogue's outer-envelope role normalization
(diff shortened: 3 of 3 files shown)


Sinity:
@codex review

PR #489, opened 2026-04-26
fix: aggregate small fixes from adversarial audit (waves 1+2)

Sinity:
## Summary

35 fixes from a comprehensive 27-agent adversarial audit. Every fix is small-scope (1-15 lines) and independently verifiable.

## Problem

Two waves of adversarial audit (15 + 12 specialized agents reading 500+ source files) surfaced ~50 findings. The critical and high-severity ones were filed as individual issues (#466-#487). This PR bundles all the small-scope, low-risk fixes that remain.

## Solution

35 fixes across these categories:

**Parsers** — `str()` guard on `Role.normalize()` inputs prevents `AttributeError` on non-string role values (fixes #469)

**Hashing** — NFC normalization in `_normalize_for_hash()`, sentinel tokens in `attachment_seed()` prevent field-shift collisions, attachment SHA-256 digest included in content hash

**Storage** — FTS trigger DDL fixed to match 6-column schema (fixes #467), triggers use `rowid` instead of UNINDEXED columns (fixes #484), `COALESCE(excluded.metadata, conversations.metadata)` preserves user tags on content change (fixes #473), `quick_readiness_summary` uses correct `db_path()`, redundant index removed, misleading comment fixed

**CLI** — `open_verb` uses candidate-concatenation pattern matching `show_verb`, `show_verb` added (was absent on master per #422), `doctor`→`check` with hidden alias, `fail()` unified to `click.echo("Error: ...")`, empty archive shows onboarding hint, export help documents `provider:id`

**MCP** — `_clamp_limit` upper bound (1000), `get_stats_by` uses `Literal` type, `resolve_id` strict flag for destructive ops, `bulk_tag_conversations` item-count limits + batch TX

**Rendering** — heading escape covers all levels (`# ` through `###### `), CSV injection prevention with `_csv_safe()`, plaintext renderer added

**Config/security** — dead `fts_enabled` removed, `max_length` falsy fix, XDG path dedup, `POLYLOGUE_FORCE_PLAIN` respected in logging, db/blob `chmod 0o600`, manifest path leak fixed

**Build** — `h2` added to Nix deps for `httpx[http2]`, `hatch_build.py` sdist fix, `typing_extensions`→stdlib `typing` (18 files)

**Products** — `profile_bucket_day` timezone-aware, `records_provenance` ISO parse before compare, `pricing.py` model prefix sorted by length, `name` added to privacy content-field denylist

**Dead code** — `cli/commands/embed.py` deleted (own docstring admitted it was orphaned)

[...]

PR #504, opened 2026-04-27
fix: remaining audit findings — content hash, XSS, TOCTOU, detection, site

### LINKED ISSUE #466: fix(ids): include content blocks in content hash
## Problem

`message_content_hash()` and `conversation_content_hash()` in `pipeline/ids.py:157-227` only hash `id`, `role`, `text`, and `timestamp`. Content blocks — `tool_use`, `tool_result`, `thinking` blocks with their types, tool names, tool inputs — are completely excluded.

Two conversations whose messages have identical text but **different tool calls or tool inputs** produce the **same content hash**. During re-import, the UPSERT at `ingest_batch.py:175` checks `WHERE content_hash != excluded.content_hash` — if the hash matches, all message/attachment/block writes are skipped.

**Impact:** Structural changes in content blocks are silently lost during re-parsing and re-ingestion. This is a data-loss bug for any archive containing tool-use conversations (Claude Code, ChatGPT with tools).

## Affected code

- `pipeline/ids.py:157-175` — `message_content_hash()` builds 4-field payload
- `pipeline/ids.py:178-227` — `conversation_content_hash()` same 4-field per message
- `pipeline/services/ingest_batch.py:433` — `_check_content_unchanged()` gates on hash match
- `pipeline/prepare_enrichment.py:33` — `existing.content_hash != content_hash` comparison

## Fix direction

Include content block data in the hash: for each block, hash at minimum `type`, `tool_name` (for tool_use), `tool_id` (for tool_result), and `text`. The `_normalize_for_hash()` helper should handle blocks the same way it handles message fields.

Discovered during adversarial audit 2026-04-26.

### LINKED ISSUE #468: fix(site): XSS vectors in generated static site
## Problem

Three injection vectors in the generated static site:

1. **`| safe` in conversation page template** (`site/templates_conversation.py:215`): `{{ msg.html_content | safe }}` strips all defense-in-depth from Jinja2 autoescaping. Same pattern in `rendering/templates/conversation.html:376`.

2. **`innerHTML` in search JS** (`site/search.py:134-143`): Search results rendered via JavaScript template literals interpolating `doc.title`, `doc.provider`, and `doc.preview` from the database into `innerHTML` without sanitization. A conversation title containing `</h2><script>alert(1)</script>` executes when anyone loads the index page and searches.

3. **Search index JSON** (`site/models.py:80`): The search index file contains full conversation previews for ALL conversations, placed in the static site output and served to any visitor without authentication.

While the user owns their own archive (self-XSS in local context), if the generated static site is published to a web server, these become stored XSS.

## Affected code

- `site/templates_conversation.py:215` — `{{ msg.html_content | safe }}`
- `rendering/templates/conversation.html:376` — `{{ msg.html_content | safe }}`
- `site/search.py:134-143` — `innerHTML` with unsanitized `${doc.title}`, `${doc.preview}`
- `site/models.py:80` — search index contains full previews for all conversations

## Fix direction

- Replace `| safe` with proper HTML sanitization (e.g., `bleach` or markupsafe escape + selective unescape)
- Escape conversation titles/previews in the search JS using `textContent` or a JS-side escape function
- Consider truncating or omitting previews from the public search index
- Add a CSP header to the generated site as defense-in-depth

Discovered during adversarial audit 2026-04-26.

### LINKED ISSUE #470: fix(storage): BlobStore TOCTOU — source file read twice, content can change
## Problem

`BlobStore.write_from_path()` (`storage/blob_store.py:64-103`) opens the source file **twice**:

1. First open (line 64-73): reads full content to compute SHA-256 hash, records `blob_size`
2. Second open (line 85): reads content again to write into blob store

If the source file is modified by another process between the two reads (e.g., a JSONL log being appended to), the hash stored in the filename corresponds to the **old content** but the bytes stored on disk are the **new content**. The blob is now corrupted — `verify()` would detect the mismatch, but only if called explicitly.

Additional issue in `write_from_fileobj()` (lines 105-153): deduplication check happens **after** the full temp-file write (line 135), wasting I/O on content that already exists. Also a TOCTOU between `dest.parent.mkdir()` and `os.replace()` — a concurrent process could write the same blob and be silently overwritten.

## Affected code

- `storage/blob_store.py:64-103` — `write_from_path()` double-read window
- `storage/blob_store.py:105-153` — `write_from_fileobj()` TOCTOU on dedup

## Fix direction

- `write_from_path`: compute hash during the first (and only) read pass, buffer or tee the content into the blob store in a single pass
- `write_from_fileobj`: check deduplication before writing to temp file; use `O_EXCL` or similar to detect concurrent writes

Discovered during adversarial audit 2026-04-26.

---

**Severity note (2026-04-26):** This requires the source file to be actively modified by another process during the import window (between the two reads). Most export files are static — JSON dumps, zip archives, or completed log files. The practical risk is lower than initially assessed. Treat as HIGH, not CRITICAL.

### LINKED ISSUE #473: fix(storage): connection leaks, readiness DB path, unbounded queries
### Connection cache leaks

`storage/backends/connection.py:70-96` — thread-local `_connection_cache` accumulates connections forever. Threads that terminate leave SQLite connections open, preventing WAL checkpointing. 20-thread pool = 20 leaked file descriptors.

### quick_readiness_summary uses wrong DB path

`readiness.py:588-597` hardcodes `archive_root / "polylogue.db"` but the real DB path is `data_home() / "polylogue.db"`. When `POLYLOGUE_ARCHIVE_ROOT` differs from `XDG_DATA_HOME/polylogue`, the summary looks for the DB in the wrong directory.

### ConversationRecordQuery defaults to unbounded limit

`storage/query_models.py:67` — `limit: int | None = None`. Any code path constructing a `ConversationRecordQuery` without setting `limit` fetches ALL matching rows.

## Affected code

- `storage/backends/queries/conversations_writes.py:19-80`
- `storage/backends/connection.py:70-96`
- `readiness.py:588-597`
- `storage/query_models.py:67`

Discovered during adversarial audit 2026-04-26.

---

**Additional finding (wave 2, 2026-04-26):** Tags destroyed on content change. When `content_hash` DOES change during re-import, the UPSERT fires with `metadata = excluded.metadata` — and the pipeline always provides `"{}"` for metadata (`ingest_worker.py:708` hardcodes `"{}"`). This **overwrites any user-set tags to an empty object**. Fix: change `_CONVERSATION_UPSERT_SQL` to use `COALESCE(excluded.metadata, conversations.metadata)` for the metadata column, or remove `metadata` from the `DO UPDATE SET` clause entirely.

This is distinct from the earlier-corrected finding: when `content_hash` matches, metadata correctly survives because the UPDATE WHERE clause evaluates to FALSE. The bug only triggers when content actually changes.

### LINKED ISSUE #475: fix(site): drop content_block messages, drop branches, CSV injection, heading escape
### Messages with content_blocks but empty text silently dropped

`site/conversation_pages.py:44` — `if not msg.text: continue` drops messages that have `tool_use`, `tool_result`, or `thinking` blocks but no text field. The standalone HTML renderer correctly checks `not text and not content_blocks`.

### Branch/alternative messages silently dropped

`site/conversation_pages.py:43-45` — no `attach_rendered_message_branches()` call. Branching conversations (common in Claude Code, ChatGPT) render only the selected branch; alternatives are invisible on site pages.

### CSV export drops content blocks + injection risk

`rendering/formatting.py:142-159` — only writes `msg.text` to CSV. Messages whose content lives in content_blocks produce empty cells. Text beginning with `=`, `+`, `-`, `@` creates CSV injection when opened in Excel/LibreOffice.

### Markdown heading escape only handles `## `

`rendering/core_markdown.py:55-57` — `_escape_message_section_markers` only escapes `## ` (H2). Lines starting with `# ` (H1) or `### ` (H3+) become document headings, breaking structure.

### create_renderer() supports 2 formats; CLI advertises 8

`rendering/renderers/__init__.py:31-34` — `OutputRenderer` protocol supports only markdown and HTML. JSON, YAML, CSV, plaintext, obsidian, org handled by parallel `format_conversation()` code path.

## Affected code

- `site/conversation_pages.py:43-45` — drops content_block msgs and branches
- `rendering/formatting.py:142-159` — CSV gaps
- `rendering/core_markdown.py:55-57` — heading escape
- `rendering/renderers/__init__.py:31-34` — renderer gap

Discovered during adversarial audit 2026-04-26.

### LINKED ISSUE #477: fix(parsers): provider detection ordering + broad exception catches
### Provider detection trivially spoofable

`sources/dispatch.py:80-95` — `chatgpt.looks_like` (checks `"mapping": {}`) and `claude.looks_like_ai` (checks `"chat_messages": []`) run BEFORE the more specific `codex.looks_like` and `claude.looks_like_code` checks. A Claude Code or Codex record with a stray `"mapping": {}` key is misidentified as ChatGPT and dispatched to the wrong parser.

### Bare `except Exception` masks MemoryError in Gemini parser

`sources/parsers/drive.py:197` catches bare `Exception` (including `MemoryError`, `KeyboardInterrupt`) for defense-in-depth on Pydantic validation. Should catch only `(ValidationError, TypeError, ValueError)`.

### Non-deterministic parsing path from transient JSONL probe failure

`pipeline/services/ingest_worker.py:409-418` — `_build_stream_parse_plan()` catches broad `Exception` during JSONL probing and returns `None`, causing fallback to a different parsing path. Transient I/O error or memory pressure can cause the **same file** to produce different parse results across runs.

### No file size limit during blob acquisition

`sources/source_acquisition_components.py:210-224` — files are streamed into the blob store without any size check. Multi-GB files are fully persisted before any size gate engages.

## Affected code

- `sources/dispatch.py:80-95` — detection ordering
- `sources/parsers/drive.py:197` — bare except Exception
- `pipeline/services/ingest_worker.py:409-418` — non-deterministic probe
- `sources/source_acquisition_components.py:210-224` — no size limits

Discovered during adversarial audit 2026-04-26.

Sinity:
## Summary

Six fixes from the adversarial audit that remained after #489, plus a CodeQL-flagged refinement of the new HTML sanitizer.

## Problem

The audit pass after #489 surfaced six concrete bugs across hashing, storage durability, lifecycle hygiene, parser ordering, site rendering, and template safety:

- **Content-block coverage in content hash (#466)**: tool-use / tool-result / thinking blocks were excluded from `message_content_hash()` and `conversation_content_hash()`, producing silent identity collisions on re-import for any conversation that has structured content blocks.
- **BlobStore TOCTOU race (#470)**: `write_from_path()` read the source file twice (once to hash, once to copy). Source mutations between the two reads landed inconsistent (hash, content) pairs in the blob store.
- **Connection cache leak on exit (#473)**: cached SQLite connections were not closed on process exit; ResourceWarnings and occasional WAL leftovers in tests.
- **Provider detection ordering (#477)**: `codex.looks_like` and `claude.looks_like_code` ran *after* the weaker dict-key heuristics, so codex/claude-code records that also matched a generic shape were misclassified.
- **Template XSS surface (#468)**: conversation templates used `{{ msg.html_content | safe }}`; any pre-rendered HTML reached the page unfiltered. Defense-in-depth in the template layer was missing.
- **Site dropping content_block-only messages (#475)**: the per-message render guard was `if not msg.text`, so messages whose payload lives entirely in content_blocks (e.g. tool_use without summary text) silently disappeared from generated sites.

After the original PR opened, CodeQL further flagged the new sanitizer's `</script>` literal as not matching `</script >` (whitespace before `>`), which browsers accept — a regex bypass.

## Solution

One commit per concern, plus the sanitizer regression follow-up:

- `pipeline/ids.py` — added `_content_block_payload()`; `message_content_hash()` and `conversation_content_hash()` now include normalized content blocks. Closes #466.
- `storage/blob_store.py` — rewrote `write_from_path()` to hash and write to a temp file in a single pass, then rename atomically. Closes #470.
- `storage/backends/connection.py` — registered `atexit` handler closing every cached connection. Closes #473.
- `sources/dispatch.py` — moved `codex.looks_like` and `claude.looks_like_code` ahead of the dict-key shape checks. Closes #477.
- `rendering/renderers/html_sanitizer.py` (new) + `html_template.py` + `templates/conversation.html` + `site/templates_conversation.py` + `site/template_environment.py` + `site/search.py` — Jinja2 `sanitize_html` filter strips `<script>` tags, `on*` event handlers, and `javascript:` URLs. All conversation templates use `| sanitize_html` instead of `| safe`. Client-side search escapes user data before `innerHTML`. Closes #468.
- `site/conversation_pages.py` — render guard now respects `msg.content_blocks` (renders when text *or* blocks are present). Closes #475.
- `rendering/renderers/html_sanitizer.py` (sanitizer fix-up commit) — closing-tag regex now accepts `</script\s*>` rather than only `</script>`. New `tests/unit/rendering/test_html_sanitizer.py` covers whitespace-before-`>`, mixed case, attributed opening tags, multiline bodies, event handlers, javascript URLs, safe markup, and empty input.

## Verification

```
devtools verify
  ruff format / check ... ok
  mypy ... ok (642 source files)
  render-all ... ok
  verify-topology / -file-budgets / -test-ownership / -migrations / -cross-cuts ... ok
  pytest -q --ignore=tests/integration ... 5751 passed (incl. new sanitizer tests)
verify: all checks passed
```

CodeQL annotation on the sanitizer regex is addressed by the second commit (closing-tag whitespace tolerance + regression suite).

## Release Notes

* **Security Enhancements**
  * HTML content is now sanitized to prevent script injection attacks
  * Search results are escaped to protect against malicious content injection

* **New Features**
  * Enhanced message rendering with improved content support

* **Performance & Stability**
  * Optimized blob storage operations for faster writes
  * Improved database connection cleanup on exit
  * Provider detection refinements


Sinity:

Issue #479, opened 2026-04-26
fix(error): bare excepts in security, search degradation, diagnostics, benchmarks


Sinity:
### Bare `except Exception:` in path sanitization (CRITICAL context)

`lib/security.py:38` — `except Exception:` in `sanitize_path()` silently sets `has_symlink = False` on any unexpected exception (e.g., `PermissionError` on unreadable directories). This is a security-function guard that could mask a real traversal attack.

### Search silently degrades without warning

`lib/query_retrieval_search.py:124,146` — `except Exception:` in `search_action_results()` and `search_hybrid_results()` silently falls back to slower/less-accurate fallback search. Users get degraded results with zero indication.

### Query-miss diagnostics blinded to failures

`lib/query_miss_diagnostics.py:89,146,152,189` — six `except Exception: return None` patterns. The module designed to *diagnose* failures is itself blinded to them.

### Benchmark corruption masked

`devtools/synthetic_benchmark_runtime.py:50,55,60,102,118,291` — six bare `except Exception:` that silently record zero counts for failed queries. DELETE failure on line 291 also silently skipped.

### Archive stats/analytics failures silently downgraded

`cli/helper_summary.py:68,177` — `except Exception: logger.debug(...)` swallows archive statistics and analytics failures. Users see missing data with no error, operators never see the `debug`-level log.

### Session product refresh failure only logged as warning

`pipeline/services/ingest_batch.py:1241-1247` — bare `except Exception: logger.warning(...)`. Callers never check for failure. Products silently go stale.

### CompositeObserver swallows individual observer failures

`pipeline/observers.py:53-58` — each observer runs in `try/except Exception` that logs but continues. Broken webhook is invisible.

### assert in production code (8 locations)

`storage/raw_state_models.py:59,72,83`, `cli/query_stats.py:283,330`, `pipeline/services/ingest_worker.py:275,524` — `assert` statements vanish with `python -O`, silently proceeding to `AttributeError` on `None`.

### RuntimeError instead of typed exceptions

`operations/archive.py:942,950`, `pipeline/services/parsing.py:54`, `pipeline/services/parsing_models.py:87` — `raise RuntimeError(...)` cannot be caught by `except PolylogueError`.

## Fix direction

- Narrow exception catches to specific types; log warnings with `exc_info=True` before falling back
- Replace `assert` with explicit `if value is None: raise TypeError(...)` 
- Define typed exception subclasses for subsystem init failures
[...]

PR #523, opened 2026-04-27
fix(error): narrow excepts, surface degradations, replace asserts/RuntimeError

Sinity:
## Summary

Partial fix for the audit listed in #479. Tackles security-critical, visibility-critical, and typed-exception items; defers diagnostics-of-diagnostics, benchmark runtime, observers, and remaining asserts to a follow-up so this PR stays focused.

Ref #479.

## Problem

The 2026-04-26 adversarial audit catalogued 9 distinct error-handling failure modes. Three categories most directly affect users today:

1. **Path sanitization can be bypassed.** `lib/security.py:38` caught `Exception` and silently set `has_symlink = False`. A `PermissionError` on an unreadable directory would defeat the symlink probe — exactly the case where a real traversal attack would benefit from masked detection.
2. **Search degradations are silent.** `lib/query_retrieval_search.py` fell back to slower paths and emptied vector results on any exception with zero log output.
3. **Top-level boundaries can't catch internal failures.** Four `raise RuntimeError(...)` and three `assert isinstance(...)` calls bypassed the `PolylogueError` hierarchy and would either escape `except PolylogueError` or vanish under `python -O`.

### Security path (CRITICAL) — `polylogue/lib/security.py`

- `sanitize_path` symlink probe: narrow to `OSError`; on uncertainty set `has_symlink = True` so the path is hashed/blocked. Defense-in-depth: an unreadable parent directory is treated as suspicious rather than waved through.
- Removed the dead `try/except` around the path-cleaning loop (split/strip on a known string can't raise).
- New test `test_symlink_check_oserror_treats_path_as_suspicious` monkey-patches `Path.is_symlink` to raise `PermissionError` and asserts the result starts with `_blocked_`.

### Visible degradation — `polylogue/lib/query_retrieval_search.py`

- `search_action_results` and `search_hybrid_results` log at `WARNING` with `error`/`error_type` before falling back. The fallback behavior is unchanged; only operator visibility improves.

### Visible degradation — `polylogue/cli/helper_summary.py`

- Archive stats and analytics failures escalated from `logger.debug` → `logger.warning`. Operators searching logs for "missing data" now find an attribution.

### Typed exceptions (replace RuntimeError)

| File | Was | Now |
| --- | --- | --- |
| `operations/archive.py` × 2 | `RuntimeError("ArchiveOperations requires …")` | `ConfigError(…)` |
| `pipeline/services/parsing.py` | `RuntimeError("repository backend is not initialized")` | `DatabaseError(…)` |
| `pipeline/services/parsing_models.py` | `RuntimeError("Cannot {action}: expected phase {x}, got {y}")` | `PolylogueError(…)` |

All four are now catchable by `except PolylogueError` at CLI/MCP boundaries.

### Replace `assert` with explicit type checks — `polylogue/storage/raw_state_models.py`

Three `assert isinstance(value, str)` instances after exhaustive type narrowing replaced with `if not isinstance(value, str): raise TypeError(...)`. Asserts vanish under `python -O`, so the original code would have silently fed the wrong type to `Provider.from_string` (etc.) and produced a downstream `AttributeError`.

### Test fixtures updated

- `tests/unit/pipeline/test_parsing_service.py`: expect `DatabaseError`.
- `tests/unit/pipeline/test_prepare.py`: expect `PolylogueError`.
- `tests/unit/security/test_path_sanitization.py`: new uncertainty-as-block test.

### Deferred (will track on #479 if not split into siblings)

- `lib/query_miss_diagnostics.py` (4 bare excepts in diagnostic-of-diagnostic paths)
- `devtools/synthetic_benchmark_runtime.py` (6 bare excepts; devtools-only)
- `pipeline/observers.py` (already logs warning; lower priority)
- `pipeline/services/ingest_batch.py:1242` (already returns `failed: True` marker; caller-side change)
- `cli/query_stats.py` and `pipeline/services/ingest_worker.py` asserts (mypy narrowing only; next line would `AttributeError` immediately if the assert vanished)

## Verification

```
$ nix develop -c devtools verify --quick
verify: all checks passed

$ nix develop -c pytest -q tests/ --ignore=tests/integration
5595 passed, 10 xfailed in 181.43s
```

### DIFF
--- a/polylogue/cli/helper_summary.py
+++ b/polylogue/cli/helper_summary.py
@@ -66,7 +66,7 @@ def print_summary_impl(
     try:
         archive_stats = run_coroutine_sync(env.repository.get_archive_stats())
     except Exception:
-        logger.debug("Archive stats computation failed", exc_info=True)
+        logger.warning("Archive stats computation failed; summary will omit stats", exc_info=True)
 
     lines = [
         f"Archive: {config.archive_root}",
@@ -184,7 +184,7 @@ def print_summary_impl(
                     ui.console.print()
 
     except Exception:
-        logger.debug("Analytics computation failed", exc_info=True)
+        logger.warning("Analytics computation failed; summary will omit analytics", exc_info=True)
 
 
 __all__ = ["print_summary_impl"]
--- a/polylogue/lib/query_retrieval_search.py
+++ b/polylogue/lib/query_retrieval_search.py
@@ -7,7 +7,10 @@ from heapq import heappush, heappushpop
 from typing import TYPE_CHECKING
 
 from polylogue.lib.query_support import provider_values
+from polylogue.logging import get_logger
 from polylogue.storage.search_providers.hybrid import reciprocal_rank_fusion
+
+logger = get_logger(__name__)
 
 if TYPE_CHECKING:
     from polylogue.lib.models import Conversation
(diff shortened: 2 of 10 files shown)


Sinity:
@codex review

PR #554, opened 2026-04-28
fix: add FTS trigger awareness to readiness checks and self-healing repair

Sinity:
## Summary
FTS triggers could be silently missing after suspend/resume cycles, causing index staleness that went undetected by readiness checks and doctor diagnostics.

## Problem
When `suspend_fts_triggers` ran and the process terminated before `restore_fts_triggers` (crash, interrupt, or exception between the two calls), triggers stayed dropped permanently. Readiness checks only verified row counts — they never checked whether the FTS triggers actually existed. `ensure_fts_index` also didn't recreate triggers, so any code path that created FTS tables via that function left them triggerless.

## Solution
Three-layer fix:
1. **`fts_lifecycle.py`**: Split `_FTS_TRIGGER_NAMES` into `_MESSAGE_FTS_TRIGGER_NAMES` and `_ACTION_EVENT_FTS_TRIGGER_NAMES`. Added `_triggers_present_sync/async` helpers that query `sqlite_master` for trigger existence. Updated `message_fts_readiness_sync/async` to check triggers and return `triggers_present` in the result dict. `ensure_fts_index_sync/async` now call `restore_fts_triggers` after creating FTS tables.
2. **`repair.py`**: `repair_dangling_fts` now calls `restore_fts_triggers_sync` after fixing orphaned rows, using `CREATE TRIGGER IF NOT EXISTS` for idempotency.
3. **`readiness/__init__.py`**: Doctor reads `triggers_present` from the readiness result and reports a specific WARNING when triggers are missing.

## Verification
- `devtools verify --quick` (format, lint, mypy, render-all, topology checks)
- Unit tests: `pytest -q --ignore=tests/integration` — 3963 passed
- `test_fts5.py` stub updated to return proper trigger count

## Summary by CodeRabbit

* **New Features**
  * Added detection of missing FTS (Full-Text Search) triggers with actionable repair recommendations during index readiness checks

* **Bug Fixes**
  * Enhanced FTS index lifecycle management to track trigger presence and automatically restore missing triggers during repair operations
  * Improved diagnostics to distinguish missing triggers from other readiness issues

* **Tests**
  * Updated test stubs for improved FTS trigger verification accuracy

### DIFF
--- a/docs/plans/file-size-budgets.yaml
+++ b/docs/plans/file-size-budgets.yaml
@@ -45,7 +45,7 @@ exceptions:
     reason: "tracked split into per-provider model contract suites"
 
   - path: tests/unit/storage/test_fts5.py
-    ceiling: 1500
+    ceiling: 1510
     until: "#419"
     reason: "tracked split for FTS escaping vs hybrid ranking"
 
--- a/polylogue/readiness/__init__.py
+++ b/polylogue/readiness/__init__.py
@@ -164,12 +164,20 @@ def _message_index_check(conn: sqlite3.Connection, *, exact_counts: bool) -> Rea
 
     indexed_rows = int(index_readiness["indexed_rows"])
     total_rows = int(index_readiness["total_rows"])
+    triggers_present = bool(index_readiness.get("triggers_present", True))
     if bool(index_readiness["ready"]):
         return ReadinessCheck(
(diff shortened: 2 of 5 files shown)


Sinity:
@codex review

PR #557, opened 2026-04-28
fix: harden blob store against path injection, partial writes, and temp file leakage (#545)

### LINKED ISSUE #545: fix(security): blob_store path injection, sandbox escape, and data integrity
## Cluster: Storage security and data integrity

Findings from the post-hoc bot review audit (#544).

### CONFIRMED

1. **`blob_store.py:38` — path injection via `hash_hex`** (CodeRabbit, Critical)
   `blob_path()` constructs filesystem path from `hash_hex` with zero validation. A malformed `raw_id` containing `../` or `/` would escape the blob root directory.

2. **`blob_store.py:127` — partial write corruption** (Copilot)
   `os.write(fd, chunk)` can perform partial writes. For large chunks, a partial write would silently corrupt the blob since the hasher was already updated.

3. **`blob_store.py:171` — temp file leakage** (CodeRabbit, Major)
   `iter_all()` uses `prefix_dir.iterdir()` without filtering `.blob.*` temp files. Crash-left temp files are yielded as valid blobs.

4. **`blob_store.py:72` — heartbeat exception aborts write** (CodeRabbit, Major)
   Exception in `heartbeat()` callback propagates and aborts the entire blob write.

5. **`lib/security.py:55-56` — sandbox escape via absolute paths** (Copilot)
   `sanitize_path()` preserves absolute paths like `/etc/passwd` — no allowlist for safe directories.

6. **`storage/blob_store.py:88,94` — TOCTOU race** (Both, UNCERTAIN)
   File could change between hash check and write. Appears partially mitigated in HEAD but needs verification.

Ref #544

Sinity:
## Summary
Five security and data-integrity fixes to blob store and path sanitization.

## Problem
- `blob_path()` had no validation on `hash_hex`, allowing path injection via `../`
- `os.write()` partial writes could silently corrupt blobs
- Crash-left temp files (`*.blob.*`) were yielded as valid blobs
- Heartbeat callback exceptions aborted blob writes
- `sanitize_path()` passed absolute paths like `/etc/passwd` through unmodified

## Solution
- Validate hash_hex against `^[0-9a-f]+$` regex
- Retry partial `os.write()` in a loop
- Filter temp files from `iter_all()`
- Suppress heartbeat exceptions with `contextlib.suppress`
- Block absolute paths in `sanitize_path()`, returning `_blocked_<hash>`

Closes #545

### DIFF
--- a/polylogue/lib/security.py
+++ b/polylogue/lib/security.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import hashlib
 from pathlib import Path
 
 from polylogue.logging import get_logger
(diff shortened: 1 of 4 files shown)


Sinity:
@coderabbitai full review

PR #558, opened 2026-04-28
fix(pipeline): wire embed stage dispatch, add webhook status check, fix upsert data loss (#546)

Sinity:
## Summary
Seven pipeline fixes from the post-hoc audit: embed stage dispatch, webhook status checking, content_blocks upsert data loss, write ordering, stage sequences, set rebuild optimization, and stale docstrings.

## Problem
- `execute_embed_stage` was imported but never called — `stage="embed"` raised ValueError
- Webhook responses were never checked — non-2xx silently treated as success
- `_write_conversation` upserted content_blocks without deleting old ones
- `write_run_json` could fail after `record_run` committed, leaving inconsistent state
- Reprocess sequence omitted the site stage

## Solution
Each fix is minimal and targeted at the specific call site.

Closes #546

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #561, opened 2026-04-28
fix(pipeline): add NotebookEdit dispatch to extract_file_changes (#552)

### LINKED ISSUE #552: fix: FTS triggers permanently lost if BEGIN IMMEDIATE fails after suspend
## Cluster: Missing symmetry (suspend without restore on error path)

Found by generalization sweep during audit #544.

### CONFIRMED

1. **`ingest_batch.py:828-829`** — `suspend_fts_triggers_sync(conn)` runs BEFORE the try block. If `BEGIN IMMEDIATE` at line 829 fails (e.g., database locked), the finally block never runs, triggers stay dropped permanently, and the connection leaks. All subsequent INSERTs silently bypass FTS.

Note: my own FTS fix earlier this session added restore in finally but left suspend+begin outside the try block. This patch corrects that.

2. **`semantic_capture.py:154-180`** — `extract_file_changes` defined with tests but never imported by production code. Also missing `NotebookEdit` dispatch branch (only handles Read/Write/Edit).

3. **`run_stages.py:438-537`** — `execute_embed_stage` fully implemented with tests but never dispatched from any production call path. CLI bypasses it via `_run_embed_standalone`. Dead code.

### Already fixed

4. **FTS trigger restore** — The main suspend/restore asymmetry in `_process_ingest_batch_sync` was fixed earlier this session (restore + repair added to finally block).

Ref #544

Sinity:
## Summary
Add NotebookEdit tool dispatch to `extract_file_changes` in semantic_capture.py.

## Problem
The function handled Read, Write, and Edit tool invocations but silently dropped NotebookEdit operations from Claude Code sessions.

## Solution
Added NotebookEdit branch that extracts `new_source` content and records it as an edit operation.

Note: the other two items from #552 (FTS suspend outside try block, missing embed stage dispatch) were already fixed in #554 and #558.

Closes #552

### DIFF
--- a/polylogue/pipeline/semantic_capture.py
+++ b/polylogue/pipeline/semantic_capture.py
@@ -177,6 +177,12 @@ def extract_file_changes(tool_invocations: Sequence[Mapping[str, object]]) -> li
             if new_string is not None:
                 entry["new_content"] = new_string[:200]
             changes.append(entry)
+        elif tool_name == "NotebookEdit":
+            new_source = optional_string(input_data.get("new_source"))
+            entry: FileChangeSummary = {"path": path, "operation": "edit"}
+            if new_source is not None:
+                entry["new_content"] = new_source[:500]
+            changes.append(entry)
     return changes
 
 


Sinity:
@coderabbitai full review

PR #564, opened 2026-04-28
feat(search): degrade gracefully on small FTS gaps instead of hard-failing (#499)

### LINKED ISSUE #499: fix(query): graceful degradation on FTS gap; make doctor --repair --target dangling_fts actually visible
## Problem

Read-only query commands abort with a hard error when the FTS index is even slightly behind the messages table. Reproducer:

```
$ polylogue --plain "anything" --since 2026-04-15 --until 2026-04-23 list --limit 5
Query still running after 2.0s (route: summary-list; retrieval: auto).
Error: Search index is incomplete. Run `polylogue doctor --repair --target dangling_fts`, or `polylogue run all`.
```

The actual gap is small. `polylogue doctor` reports:

```
WARN index: messages indexed: 1,862,988/1,863,031
```

That's 43 unindexed messages out of 1.86M — 0.002%. For that gap, every read-only search verb refuses to execute. `count`, `list`, `stats`, `--similar`, etc. all abort with the same message.

The same pattern holds for `action_event_fts` (327,002/329,106 — 2,104 missing) and `transcript_embeddings` (0/10,657 — pending).

### Why this hurts

1. **Agents and operators can't make progress on a healthy archive.** A drifted FTS index is a maintenance condition, not a corruption. The remaining 99.998% of indexed content is fine to search; one missing message is not a reason to refuse all queries.

2. **The suggested repair command produces no visible output in `--preview` mode** — it runs `polylogue check` instead of showing what would be repaired. Without `--preview`, presumably it repairs, but the operator has no way to confirm scope before executing.

3. **The error fires at search time, not at archive boundary time.** Ingestion happily keeps the gap open without warning anyone. The first symptom is a query failure that doesn't name a specific cause.

4. **The threshold for "incomplete" is binary** — a gap of 1 fails as hard as a gap of 100,000.

### Degrade gracefully on small gaps

Read-only verbs (`list`, `count`, `stats`, `--similar`, `--sort`, `show`, `open`) should:

1. Detect the FTS gap.
2. If the gap is below a threshold (default 1%, configurable via `polylogue config set`), emit a one-line warning to stderr and execute the query against the available FTS rows. The warning carries the missing count and the repair command.
3. If the gap exceeds the threshold, behave as today: refuse with the error and the repair command.

Mutating verbs (`delete`, `--add-tag`, `polylogue run all`) and explicit consistency operations (`check`, `audit`) keep the strict gate.

### Make the suggested repair actually repair (and confirm)

`polylogue doctor --repair --target dangling_fts`:

1. Without `--preview`: re-indexes the missing rows and prints `repaired N messages` on stdout.
2. With `--preview`: prints the rowid range or count that would be repaired, not the full health-check output. Today `--preview` produces the same output as bare `polylogue check`, which is misleading.
3. Should support `--format json` like other commands (today it returns `No such option: --format`).

### Surface the gap before query time

Add a non-blocking warning in `polylogue --version` or any first-of-session command output: `archive: 1,862,988/1,863,031 messages indexed (0.002% pending)`. So the operator sees the state continuously, not only when a query fails.

## Why it matters for verifiability

The catalog has a maintenance subject for `dangling_fts` repair. Today there is no obligation that catches "FTS gap drifted to a degree that blocks queries" before it actually blocks one. A hard-gate for 0.002% drift is a false-positive of the verification surface — it doesn't reflect a real correctness failure, only a hygienic backlog. Soft-degrade-with-warn aligns the surface with the actual risk.

## Acceptance (executable)

```
# Reproduces today
polylogue --plain "anything" list --limit 1 2>&1 | grep -F "Search index is incomplete"

# After fix, with default threshold (gap < 1%)
polylogue --plain "anything" list --limit 1 2>&1 | grep -E "WARN: archive index is.*pending"
# query returns results

# Repair scope is visible
polylogue doctor --repair --target dangling_fts --preview --format json | jq '.would_repair'
# integer count of rows that would be repaired

# Repair runs and reports
polylogue doctor --repair --target dangling_fts | grep -E "repaired [0-9]+ messages"
```

## Related

- #467, #484 — closed FTS trigger fixes (different issue: this one is about UX of the gap state, not the trigger correctness).
- #494 (evidence freshness SLA) — analogous shape: maintenance state should have a soft-warn-then-hard-fail SLA, not an instant hard fail.

Sinity:
## Summary
When the FTS index is ≤1% behind the messages table, emit a warning instead of refusing all queries. Gaps above the threshold still hard-fail.

## Problem
Read-only query commands aborted with a hard error when the FTS index was even 0.002% behind. For 43 unindexed messages out of 1.86M, every search verb refused to execute.

## Solution
- Added `check_fts_readiness()` in `fts_lifecycle.py` — shared sync/async helper that computes gap ratio and either warns (≤1%) or raises (>1%)
- Updated `search_messages_impl` (runtime.py) to use the helper with `exact_counts=True`
- Updated both async search paths in `conversations_search.py`

The threshold (`FTS_GAP_THRESHOLD = 0.01`) is a module-level constant for now.

Closes #499

### DIFF
--- a/polylogue/storage/backends/queries/conversations_search.py
+++ b/polylogue/storage/backends/queries/conversations_search.py
@@ -21,14 +21,10 @@ async def search_conversation_hits(
     limit: int = 100,
     providers: list[str] | None = None,
 ) -> ConversationSearchResult:
-    from polylogue.errors import DatabaseError
-    from polylogue.storage.fts.fts_lifecycle import message_fts_readiness_async
+    from polylogue.storage.fts.fts_lifecycle import check_fts_readiness, message_fts_readiness_async
 
-    readiness = await message_fts_readiness_async(conn)
-    if not bool(readiness["exists"]):
-        raise DatabaseError(f"Search index not built. {_MESSAGE_SEARCH_REPAIR_HINT}")
-    if not bool(readiness["ready"]):
-        raise DatabaseError(f"Search index is incomplete. {_MESSAGE_SEARCH_REPAIR_HINT}")
+    readiness = await message_fts_readiness_async(conn, exact_counts=True)
+    check_fts_readiness(readiness, _MESSAGE_SEARCH_REPAIR_HINT)
 
     from polylogue.storage.search import build_ranked_conversation_search_query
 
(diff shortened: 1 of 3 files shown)


Sinity:
@coderabbitai full review

Issue #750, opened 2026-05-03
storage: lift queryable facts out of JSON before web reader grows annotation surface


Sinity:
## Problem

Three v2 storage decisions remain irreversible-by-default and are cheaper to fix before the daemon web reader (#717) starts accumulating annotations, stars, saved views, and reader marks:

1. **Tags are JSON metadata, not many-to-many rows.** Cannot be efficiently filtered, joined, or counted at archive scale; cannot enforce uniqueness or hierarchical structure.
2. **Insights live in the same SQLite as raw archive data.** Mixed write contention, mixed backup semantics, no clean wipe-and-rebuild for derived state.
3. **Queryable fields keep landing as new JSON blobs.** Each one becomes an expensive scan and a future migration.

## Why now

#717 ships the local web reader. The reader will quickly want stars, marks, saved searches, annotations, and view state. If those land as JSON metadata, the migration trap doubles in size and these v2 corrections become much harder.

### Tags → many-to-many

```sql
CREATE TABLE tags (
  id   INTEGER PRIMARY KEY,
  name TEXT    UNIQUE NOT NULL
);
CREATE TABLE conversation_tags (
  conversation_id TEXT    NOT NULL,
  tag_id          INTEGER NOT NULL,
  PRIMARY KEY (conversation_id, tag_id)
);
```

Plus explicit `add_tag` / `remove_tag` / `list_tags` API surface (already in the #723-lite proposal on #730).

### Insights → `insights.db`

Separate SQLite file. Rebuildable from archive + raw. Different backup cadence. Different write profile (batch rebuilds vs trickle ingest). Cleaner wipe semantics for the assurance-ontology evidence work too.

### Queryable fields → columns

Any new field the CLI, MCP, or API exposes for filtering or sorting must be a real column with an index, not a JSON path. Add a verifier (per the breaker rule in #594) that fails if a query descriptor field maps to a JSON path on the hot read path.

## Acceptance criteria

- [ ] Tags M2M schema landed; old JSON tag field migrated and removed (fresh-only schema discipline applies — schema version bump + reset-and-rebuild diagnostic).
- [ ] `insights.db` extracted; archive DB no longer contains insight tables; rebuild path documented.
- [ ] Lint or test fails when a query descriptor field is added without a backing column + index.
- [ ] Schema version bumped; CHANGELOG documents the wipe-and-rebuild migration.

## Sequencing

[...]

Issue #722, opened 2026-05-02
feat(storage): implement blob store garbage collection


Sinity:
## Problem

The blob store is content-addressed and immutable at rest. When a source file grows (append-only), the daemon verifies the new version is a valid extension of the old, stores the new blob, and updates the `raw_conversations` reference. But old blobs that are no longer referenced by any conversation accumulate indefinitely.

Currently the blob store is 41G — some fraction of that is unreferenced old versions that could be reclaimed.

## Design

Blob GC identifies blobs with zero `raw_conversations` references and deletes them. It must be safe:

- Never delete a blob referenced by any `raw_conversations` row.
- Never delete a blob that an in-flight daemon ingestion is about to reference.
- Run as bounded background maintenance, not as a stop-the-world operation.
- GC is triggered by the daemon after a batch of replacements accumulates, or on a schedule.

## Acceptance criteria

- `polylogue status` reports: total blob store size, estimated reclaimable size (blobs with zero references).
- Daemon runs GC after N blob replacements or on a time schedule. GC is bounded (time limit or count limit per run).
- GC never deletes a blob referenced by an active `raw_conversations` row.
- GC is safe with respect to in-flight daemon ingestions.
- Manual GC trigger available via daemon API or CLI flag.
- GC is tested: unreferenced blobs deleted, referenced blobs preserved.
- Reset interacts correctly: `reset --all` followed by re-ingestion should not create GC windows where blobs are prematurely deleted.

Ref #717
Ref #618

Sinity:
## Streaming append-only verification algorithm (2026-05-03 agent analysis)

### Goal
Verify that `file_v2` on disk is a valid extension of the blob identified by `old_hash` — i.e., `file_v2` starts with the exact bytes of the old blob, plus additional content.

### Algorithm

Uses SHA-256 collision resistance: hash the first `old_size` bytes of `file_v2`. If the hash matches `old_hash`, the prefix is provably the old content (no need for byte-by-byte comparison or reading the old blob).

```python
def is_valid_extension(old_hash: str, current_path: Path, blob_store: BlobStore) -> bool:
    old_size = blob_store.blob_path(old_hash).stat().st_size
    current_size = current_path.stat().st_size
    if current_size <= old_size:
        return False  # must be strictly larger
    hasher = hashlib.sha256()
    with open(current_path, "rb") as f:
        remaining = old_size
        while remaining > 0:
            chunk = f.read(min(128 * 1024, remaining))
            if not chunk:
[...]

Issue #618, opened 2026-05-01
fix(storage): unify schema, FTS, cache, and write side effects


Sinity:
## Scope

Composite issue for archive-storage correctness and compactness around schema bootstrap, direct write paths, FTS/search side effects, legacy product payloads, and storage helper duplication.

## Problem

Several storage issues share one root: archive writes and schema side effects are split across multiple paths that are not governed by one contract.

### Sync ingest FTS/cache side effects (#707 / `fix(pipeline): commit sync ingest side effects`)
- `_process_ingest_batch_sync()` now restores FTS triggers, repairs changed-conversation message/action FTS rows, commits that teardown work, then invalidates the search cache when conversations changed.
- Regression test: caches an empty search, sync-ingests a conversation through the direct sqlite batch path, then verifies from a fresh connection that FTS rows persisted and the same search query returns the new conversation.

### Schema version mismatch (#709 / `fix(storage): block schema version mismatch`)
- Schema-version mismatch now blocks with the reset/re-import diagnostic instead of running additive DDL and advancing `PRAGMA user_version`.
- Sync and async schema setup share that behavior.
- `open_read_connection()` performs a non-mutating compatibility check for mismatched versions and known legacy inline raw layouts.
- Removed the unused metadata export/import path that belonged to the misleading in-place recreate story.

### Force-write propagation (#711 / #714)
- `run --reparse` propagates `force_write` through run workflow, pipeline execution, parsing services, and sync ingest writes.
- `_write_conversation(..., force_write=True)` replaces same-hash message rows (corrected in #714 — earlier version left old message rows behind).

## Remaining acceptance criteria

1. **Broader write-path unification**: repository writes, sync ingest/batch writes, and other direct archive write paths should share a post-write contract for FTS repair, trigger restoration, cache invalidation, and product/readiness side effects.
2. **FTS trigger/table SQL single source of truth**: FTS lifecycle SQL is still split across `schema_ddl_archive.py` and `fts_lifecycle.py`. Unify.
3. **Hot search readiness**: exact row-count FTS readiness should not be on the hot search path. Exact counts belong in doctor/readiness/repair, or behind bounded freshness caching.
4. **Legacy session-product payload support**: delete from hot read/write paths or move into an explicit documented migration/repair path.
5. **Unsafe asset IDs**: accepted through the advertised `asset_path()` contract without failing in temp-file creation.
[...]

PR #754, opened 2026-05-03
feat(storage): ArchiveWriteGateway, tags M2M, insights.db, blob GC (#618, #722, #750)

Sinity:
Storage v2 corrections. Schema v3→v5. 10 files, 716 insertions.


Sinity:
Recreating from rebased branch.

Issue #839, opened 2026-05-05
fix(archive): store context and protocol artifacts semantically, not via projection fallbacks


Sinity:
## Problem

PR #836 taught projections how to avoid treating Claude Code / Codex protocol artifacts as ordinary user prose, but existing archive rows still contain many artifacts only as inline message text or legacy `message_type` values. Runtime classifiers then have to compensate when stored rows are missing semantic type information.

That is the wrong layer. The raw source blob already preserves exact provider JSON/text. Archive tables should represent the semantic entities Polylogue understands: human/user prose, assistant prose, tool use/results, context dumps, protocol/runtime notices, pasted bodies, provider events, and similar categories.

Concrete observed artifacts include:

- `<system-reminder>...</system-reminder>` blocks;
- task/task-notification wrappers;
- local command wrapper text;
- pasted skill bodies;
- Claude Code rows stored with provider role `user` but semantically `tool_result` or protocol/context content.

## Scope

- Define the archive-level ontology for context/protocol artifacts that are currently embedded in message text or recovered at projection time.
- Parse these artifacts into stable message/content/artifact rows during ingestion or rematerialization.
- Rematerialize existing rows so `message_type=context|protocol|tool_result|...` and/or dedicated artifact rows are authoritative in storage.
- Remove projection/runtime fallbacks that infer semantic type from raw text because the DB row is deficient.
- Keep raw provider blobs as the exact raw preservation layer; do not preserve provider wire quirks as canonical archive semantics.

## Acceptance Criteria

1. Stored rows, not projection heuristics, classify context/protocol artifacts for Claude Code and Codex conversations.
2. Existing DB rows are rematerialized or migrated so current archive reads do not need text-based type fallback for known artifact classes.
3. `content_projection` and message model helpers stop using runtime classification to compensate for missing `message_type` rows.
4. `<system-reminder>` and similar structured blocks are either represented as first-class artifact/content rows or explicitly classified in a documented semantic field; they are not silently erased from canonical storage.
5. Tests cover at least system reminders, task notifications, command wrappers, pasted skill bodies, and role/type asymmetry where provider role is `user` but semantic type is not ordinary user prose.
6. The PR includes before/after counts for artifact classes in the live or fixture archive.

## Related

- #837 handles provider-level protocol events such as context compactions.
- #838 depends on this cleanup so context packs can suppress protocol noise without verbatim text heuristics.
- #800 tracks broader provider parser/schema drift.

Sinity:
[...]

PR #990, opened 2026-05-09
feat(maintenance): add message_type backfill for pre-#839 rows

Sinity:
## Summary
Adds an idempotent message_type backfill repair operation that classifies pre-#839 rows (context/protocol markers) using the same `classify_text_message_type` logic as new-ingest materialization.

## Problem
#894 gave new rows CONTEXT/PROTOCOL message types and #977 dropped the runtime classifier fallback. But rows ingested before #894 still have the default `message_type = 'message'`, even when their text matches known context or protocol markers. The existing v2-to-v3 backfill only handles thinking/tool_use/tool_result types.

## Solution
Added `repair_message_type_backfill` in `polylogue/storage/repair.py` that:
- Batches over messages with `message_type = 'message'` and non-empty text
- Applies `classify_text_message_type()` — the same function used at materialization time
- Updates matched rows to `context` or `protocol`
- Is idempotent: second run is a no-op (classified rows no longer match the WHERE clause)

Registered as `message_type_backfill` maintenance target. Accessible via:
```bash
polylogue doctor --repair --target message_type_backfill
```

## Verification
```
pytest tests/unit/pipeline/test_message_type_backfill.py tests/unit/core/test_maintenance_targets.py tests/unit/storage/test_repair.py tests/unit/archive/test_model_runtime_no_runtime_classify.py -q
31 passed
```
Devtools verify --quick passes (pre-existing file-size overages in daemon/status.py and operations/archive.py remain).

Refs #839

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/docs/plans/file-size-budgets.yaml
+++ b/docs/plans/file-size-budgets.yaml
@@ -58,8 +58,8 @@ exceptions:
     reason: "tracked split into routing/execution/output suites"
 
   - path: polylogue/storage/repair.py
-    ceiling: 1200
-    reason: "added blob orphan detection, cleanup, and integrity verification (#818)"
+    ceiling: 1300
+    reason: "added blob orphan detection, cleanup, integrity verification (#818), and message_type backfill (#839)"
 
   - path: polylogue/storage/sqlite/schema_bootstrap.py
     ceiling: 1300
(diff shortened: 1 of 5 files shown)


Sinity:
Superseded by #988 — same branch, agent coordination race.

Issue #1310, opened 2026-05-18
audit(governance): closed-issue workload that fell through


Sinity:
## Summary

Audit of ~110 closed issues from 2026-05-01 → 2026-05-18 (the last 17 days of closures, capturing the full window since the prior governance reset in #944/#970/#971/#972). Most closures are clean — the assured-close discipline introduced after #1002/#1004 has clearly raised the floor. Decomposition umbrellas (#994, #995, #1058, #998) close with explicit per-AC matrices and verified successors. The 2026-05-01 mass-NOT_PLANNED batch (#626–#633) cleanly folds into open composite owners (#614, #621, #623, #624). However, four cases violate the close-discipline contract and one chain-fold leaks scope across three issues.

### Counts
- Clean closures: ~95
- **PARTIAL** (closed with explicit unfinished AC, no successor filed): **3**
- **ABANDONED** (closed with "incremental / remaining / etc." and no `#NNN`): **2**
- **ORPHANED-FOLD** (scope folded through a chain to a target that doesn't cover it): **1** (one chain, one root)

---

### PARTIAL — closed with unfinished acceptance and no successor

| # | Title | What fell through | Recommended remediation |
|---|-------|-------------------|-------------------------|
| **#1012** | Pre-existing test failures discovered during #1007 verification | The 2026-05-15 audit comment explicitly states "the issue is not closable yet" — `tests/unit/cli/test_command_aux_runtime.py::test_tags_command_plain_paths_cover_empty_hint_and_tabular_counts` remains XFAIL for provider-aware empty hint and Rich table renderer; stale xfails on the two XPASS rows were also flagged. Closed the next day by PR #1095 which addressed harness staleness only. The tags XFAIL is not in #1180 or any open issue. | File a successor issue: "fix(cli): provider-aware empty hint + Rich table for tags command (#1012 remainder)" — include the exact test node and the two stale xfails that should be removed. |
[...]

Issue #1243, opened 2026-05-17
perf(storage): compound action_events indexes + ANALYZE schedule (#817 slice D)


Sinity:
## Problem

#817 evidence: missing compound indexes on action_events; ANALYZE not scheduled in daemon maintenance cycle; per-call FTS-readiness COUNT(*) on every search.

## Scope

- [ ] Add `idx_action_events_conv_kind` and `idx_action_events_conv_tool` compound indexes
- [ ] Run ANALYZE on schema bootstrap; schedule periodic ANALYZE in daemon maintenance cycle
- [ ] Remove per-call FTS-readiness COUNT(*); daemon already checks at startup
- [ ] Increase `_SESSION_INSIGHT_REBUILD_PAGE_SIZE` from 1 to 50-200 (avoids ~17K round-trips for 4K conversations)
- [ ] Pre-compute provider metrics from `conversation_stats` instead of scanning messages

## Acceptance

- [ ] Before/after query-plan + benchmark evidence in PR body
- [ ] Daemon convergence benchmark shows reduced per-cycle FTS-readiness overhead

Ref #817.

Sinity:
Closing as realized by the merged #1314/#1334 perf rescue and subsequent daemon boundedness work.

Acceptance evidence checked on current `master` (`34fae42d`):

- `idx_action_events_conv_kind` and `idx_action_events_conv_tool` exist in canonical DDL and runtime index ensure paths:
  - `polylogue/storage/sqlite/schema_ddl_actions.py`
  - `polylogue/storage/sqlite/runtime_indexes.py`
- Live archive query plans use the compound indexes:
  - `SEARCH action_events USING INDEX idx_action_events_conv_kind (conversation_id=? AND action_kind=?)`
  - `SEARCH action_events USING INDEX idx_action_events_conv_tool (conversation_id=? AND normalized_tool_name=?)`
- Provider metrics are sourced from `conversation_stats` for precomputed per-conversation aggregates; live plan shows:
  - `SCAN conversation_stats USING INDEX idx_conv_stats_provider`
- `_SESSION_INSIGHT_REBUILD_PAGE_SIZE` is now `50` and guarded by `tests/unit/storage/test_perf_rescue_1314.py`.
- Search no longer performs archive-scale FTS `COUNT(*)` probes on the hot retrieval path when the daemon freshness ledger marks `messages_fts` ready; that behavior is guarded by `tests/unit/storage/test_perf_rescue_1314.py`.
[...]

Issue #1183, opened 2026-05-17
test(perf): tiered scale fixtures separating local, lab, and CI/nightly performance gates


Sinity:
## Problem

#997 status update flagged **performance/regression** as remaining: \"#1086 / #1072 cover the catalog/campaign side, but the comprehensive scale-fixture tier separation (cheap local vs lab vs CI/nightly) per the AC is not fully concluded.\" Today, \`tests/benchmarks/\` mixes fixture sizes and there is no documented contract for which tier runs where. \`devtools verify --quick\` skips benchmarks, but operators have no way to run \"quick perf sanity\" vs. \"lab perf evidence\" vs. \"nightly perf gate\" without reading source.

## Goal

Define and implement a tier model for scale fixtures (small / medium / large) with explicit corpus targets and gate placement. Wire the tiers through pytest markers and \`devtools\` so the right subset runs in the right context.

## Acceptance Criteria

- [ ] \`tests/infra/scale_fixtures.py\` (or extend \`tests/infra/corpus_fixtures.py\`) exposes \`tier_small_db\`, \`tier_medium_db\`, \`tier_large_db\` factories with documented row counts (e.g. 100 / 10k / 100k messages)
- [ ] Pytest markers \`@pytest.mark.scale_small\`, \`@pytest.mark.scale_medium\`, \`@pytest.mark.scale_large\` registered in \`pyproject.toml\` and \`tests/conftest.py\`; the existing \`tests/benchmarks/\` files are migrated to use them
- [ ] \`devtools verify --lab\` runs \`scale_medium\`; CI nightly job (or explicit \`devtools benchmark-campaign run --tier large\`) runs \`scale_large\`; quick gate stays free of scale tests
- [ ] Each tier has at least one benchmark per measured surface: search (FTS5 + hybrid), list_conversations, daemon catch-up, ingest throughput
- [ ] \`docs/test-quality-workflows.md\` documents the tier model and how to add a new scale benchmark
- [ ] Growth-shape assertions where applicable: large-tier latency must be within a documented factor of medium-tier (not absolute milliseconds), to keep machine-portable

## Size

M (2 days). Mostly fixture/marker plumbing plus one round of migration on existing benchmark files.

## Dependencies

Builds on #1072 / #1086 (benchmark catalog and pytest-benchmark integration) which are already shipped.

Ref #997

Issue #817, opened 2026-05-04
refactor(storage): reduce FTS bloat and stale schema/read-path drift


Sinity:
## Current status

Later comments supersede several original dead-schema claims:

- `live_cursor` is active daemon state, not a dead table.
- `conversation_tags` is active schema/query/write surface, not dead; the actionable gap is stale JSON tag reads that should use `conversation_tags` + `tags`.
- `ArchiveWriteGateway` is not dead after #847; it is part of the write/effects boundary.

Use this issue for storage bloat, stale read paths, and FTS/index safety, not for deleting live daemon/write infrastructure.

## Scope

- Verify and, if still valid, reduce `messages_fts` content bloat.
- Migrate remaining JSON tag read paths to M2M tag tables.
- Remove or justify unused indexes / low-value columns with query evidence.
- Review unsearched insight/session FTS tables and triggers: drop, repurpose, or add actual MATCH/query usage.
- Fix FTS trigger/side-effect safety where bulk writes can leave FTS stale after failure or SIGKILL.
- Keep schema changes fresh-first and explicit; if a live archive upgrade is needed, include the exact version transition plan.

## Non-goals

- Do not delete `live_cursor`.
- Do not delete `conversation_tags` merely because current user data has zero tags.
- Do not remove `ArchiveWriteGateway` without new evidence after #847.
- Do not use this issue to restore legacy run-stage assumptions.

## Acceptance criteria

- [ ] Each removed index/table/column has query evidence or a documented replacement.
- [ ] Remaining tag reads use the canonical M2M tag model or have an explicit compatibility reason.
- [ ] FTS bloat is measured before/after or the claim is marked misframed.
- [ ] FTS trigger suspension/repair behavior has regression coverage for failure/restart paths.
- [ ] Schema upgrade/fresh-init implications are recorded in the PR and verification notes.

Refs #841 #845 #852.

Sinity:
## Acceptance Criteria

### Dead code removal
1. `conversation_tags` table dropped or wired (currently 0 rows, never SELECTed)
2. `live_cursor` table dropped (0 rows, not in any DDL)
3. 8 unused JSON-extract indexes on attachments/attachment_refs dropped
4. `idx_content_blocks_conv_semantic` and `idx_content_blocks_semantic_type` dropped or justified
5. `media_type` column on content_blocks dropped (1 row set out of 2.5M)

### Storage optimization
6. `messages_fts` converted to `content='messages'` (content-sync) — saves ~6 GB
[...]

PR #1332, opened 2026-05-18
perf(storage): rescue perf items from #809 fold (#1314)

### LINKED ISSUE #1314: perf(storage): rescue perf items from #809→#815→#817 fold (from #1310 audit)
## Problem

Fall-through ORPHANED-FOLD from #1310 audit. Chain #809 → #815 → #817 (current open umbrella).

Perf items from #809 that are absent from #817's body and AC:

1. \`_SESSION_INSIGHT_REBUILD_PAGE_SIZE = 1\` at \`rebuild.py:82\` causes ~17K SQL round-trips for 4K conversations. Fix: increase to 50-200.
2. Per-call FTS readiness \`COUNT(*)\` on every search (daemon already checks at startup — redundant).
3. Provider metrics scan messages table instead of pre-computing from \`conversation_stats\`.
4. \`model_copy()\` in hydration where lazy content block lookup would suffice.

#817 slice D (#1243) covers the compound-index and ANALYZE work, but the page-size and readiness-cache items are not in any current scope.

## Scope

Fold all four items into a single perf-cleanup PR. Each is independently small but they cluster around the session-insight rebuild hot path.

## Acceptance criteria

- [ ] \`_SESSION_INSIGHT_REBUILD_PAGE_SIZE\` raised with benchmark evidence
- [ ] Per-call FTS \`COUNT(*)\` removed; startup check is canonical
- [ ] Provider metrics read from \`conversation_stats\`, not messages scan
- [ ] Lazy content-block hydration; \`model_copy()\` removed from hot path
- [ ] Before/after benchmark in PR body (use #1183 scale_small fixture)

## Reference

#809 (closed parent), #815 (closed pivot), #817 (current umbrella, missing these items), #1243 (sibling: indexes + ANALYZE), #1310 (fall-through audit).

Sinity:
## Summary

Four small storage-side performance fixes that fell through the
#809 → #815 → #817 fold. Each is a localized substitution; together they
cut session-insight rebuild SQL round-trips ~2x, drop per-search FTS
readiness cost ~4x, speed provider-metric reads ~24%, and trim a wasted
`model_copy` allocation from message hydration.

## Problem

The #1310 fold audit surfaced four perf items from #809 that #817's
acceptance criteria do not cover:

1. `_SESSION_INSIGHT_REBUILD_PAGE_SIZE = 1` at `rebuild.py:90` forces the
   full-rebuild path through one-conversation-per-chunk iteration. On a
   ~4K-conversation archive that is ~17K SQL round-trips for the load
   batch alone.
2. The per-search FTS readiness check (`message_fts_readiness_sync` /
   `_async` with `verify_total_rows=True`) runs three `COUNT(*)`
   queries on every search call (`messages_fts_docsize`,
   `action_events_fts_docsize`, `messages WHERE text IS NOT NULL`),
   redundant with the daemon's startup/repair-time exact check.
3. `get_provider_metrics_rows` aggregated everything from the large
   `messages` table even though `conversation_stats` already carries
   the per-conversation totals.
4. `SQLiteQueryStoreArchiveMixin.get_messages*` always rebuilt every
   `MessageRecord` via `model_copy`, even when the message has zero
   content blocks — the field already defaults to `[]`.

#817 slice D (#1243) covered the compound-index and `ANALYZE` work but
not the items above. This PR closes that gap as one focused perf change.

## Solution

- `polylogue/storage/insights/session/rebuild.py`: raise
  `_SESSION_INSIGHT_REBUILD_PAGE_SIZE` from 1 → 100 and route both the
  full-rebuild iterator and the async targeted iterator through the
  existing message-budget chunker
  (`_iter_budgeted_conversation_chunks_{sync,async}`,
  `_load_message_counts_async`). The 5 000-message RSS guard now applies
  on every code path; previously only the targeted-sync path enforced it.
- `polylogue/storage/sqlite/queries/conversations_search.py` and
  `polylogue/storage/search/runtime.py`: switch per-search readiness
  calls to `verify_total_rows=False` so each search runs the light probe
  (existence + sample + triggers) instead of three `COUNT(*)` sweeps.
- `polylogue/storage/fts/fts_lifecycle.py`: fix the light readiness
  probe to read `messages_fts_docsize` (and the async equivalent)
  instead of `messages_fts`. The external-content FTS5 table proxies
  reads through to `messages`, so `SELECT 1 FROM messages_fts` is not a
  faithful index-population signal. The docsize sidecar is. This fix is
  required for the per-search switch to keep detecting truncated FTS
  state (caught by the existing
  `test_search_with_empty_fts_rows_raises_descriptive_error`).
- `polylogue/storage/sqlite/queries/stats.py`: split
  `get_provider_metrics_rows` into two queries. Per-conversation totals
  (count, message_count, tool/thinking counts, with-tool/with-thinking
  conversation counts) come from `conversation_stats` keyed on
  `provider_name`; only the role-grouped figures (user/assistant message
  counts and word sums) still come from `messages` because
  `conversation_stats` does not carry role splits. Sorted in Python to
  preserve the existing `ORDER BY conversation_count DESC` contract.
- `polylogue/storage/sqlite/query_store_archive.py`: make all three
  hydration paths (`get_messages`, `get_messages_paginated`,
  `get_messages_batch`) skip `model_copy` when a message's
  `content_blocks` lookup is empty.
- `docs/topology-status.md`: regenerated (master had drift; this PR
  rolls it forward).

## Verification

`devtools verify --quick` is green:

```
nix develop -c devtools verify --quick
... total_duration_s: 31.34, exit_code: 0
```

Focused unit suites all pass with the changes:

```
nix develop -c pytest tests/unit/storage/ tests/unit/insights/ \
    tests/unit/operations/ tests/unit/cli/ tests/unit/mcp/ \
    tests/unit/daemon/
... 2869 passed, 1 xfailed
```

### Before / after benchmarks (scale_small from #1183)

Fixture: ~100 conversations / ~1 000 messages, seeded via
`tests/benchmarks/conftest.py::_seed_realistic_db(target_messages=1000,
seed=1183)`. Latency numbers are the median of five runs; the
session-insight rebuild SQL round-trip count is exact.

| Metric                                  | Before  | After   | Note                          |
|-----------------------------------------|---------|---------|-------------------------------|
| session_insight_rebuild_full_s          | 0.78 s  | 0.67 s  | ~14% faster                   |
| fts_readiness_200_calls_s               | 0.030 s | 0.007 s | ~4.3x faster (light path)     |
| provider_metrics_30_calls_s             | 0.033 s | 0.025 s | ~24% faster                   |
| hydration_get_messages_batch_20_calls_s | 0.83 s  | 0.74 s  | ~11% faster (noisy at small N)|

SQL round-trip count for `rebuild_session_insights_async` (full
rebuild path):

| Tier                      | page_size=1 | page_size=100 |
|---------------------------|-------------|---------------|
| scale_small (37 convs)    | 431 calls   | 215 calls     |
| scale_medium (155 convs)  | 1733 calls  | 825 calls     |

The asymptotic ceiling on the call reduction is dominated by the
per-conversation insight write helpers (`replace_session_profile`,
`replace_session_work_events`, `replace_session_phases`), which remain
one-per-bundle; the load-batch portion of the work reduces ~50x once
chunks hold 50–100 conversations instead of 1.

Closes #1314


Sinity:
Superseded by #1334 which landed first and covers the same perf items with measured wins. Closing without merge to avoid duplicate work.

Issue #1737, opened 2026-05-28
refactor(verifiability): delete ceremonial verification infrastructure


Sinity:
## Context

Audit of verification infrastructure found multiple systems that describe verification without performing it, reimplement standard tools, or are dead code. This issue removes them.

### Delete: proof catalog (~3,861 lines)
`polylogue/proof/` — catalog.py, models.py, subjects.py, runners.py, diffing.py, rendering.py, corpus.py, witnesses.py, suppressions.py, coverage_manifests.py, generated_scenarios.py. A classification taxonomy for verification metadata. Produces reports consumed only by devtools. Zero CI enforcement. The custom predicate language (Kind, AttrEq, AttrIn, And, Or, Not) reimplements Python built-ins.

Also delete all consumers:
- `devtools/render_verification_catalog.py`
- `devtools/verification_impact.py` + `verification_impact_cli.py`
- `devtools/evidence_report.py`
- `devtools/semantic_axis_evidence.py`
- `devtools/verify_suppressions.py` (suppression expiry checking)
- `devtools/verify_witness_lifecycle.py`
- `devtools/witness_discover.py`, `witness_minimize.py`, `witness_promote.py`
- `polylogue/showcase/invariants.py` (proof-catalog-driven invariant checking)
- All proof-related tests under `tests/proof/`, `tests/unit/proof/`
- `tests/unit/devtools/test_render_verification_catalog.py`
- Witness files under `tests/witnesses/` (all 5 are synthetic, zero bug reproductions)
- `tests/conftest_witness.py`

### Delete: disabled CI workflows
- `.github/workflows/proof-comment.yml` — disabled since May 2026, never coming back
- `.github/workflows/pr-policy.yml` — same
- Update CLAUDE.md references to "Polylogue Verification Impact" comments (they don't exist)

### Delete: verify-test-ownership
`devtools/verify_test_ownership.py` — pytest-testmon supersedes this. testmon tracks per-test-function affected-by-code-change at the coverage level. "Module imported by any test" is a strictly weaker signal.

### Delete: verify-file-budgets
`devtools/verify_file_budgets.py` + `docs/plans/file-size-budgets.yaml` — circumventable LOC ceilings that get bumped rather than triggering refactors.

### Delete: CATALOG-ONLY commands
`devtools/verify_cluster_cohesion.py` and `devtools/verify_migrations.py` — registered in command catalog, never invoked by verify.py or CI. If they're useful, they should have been wired in. They weren't.

[...]

PR #1766, opened 2026-05-31
ci: restore push/pull_request triggers accidentally removed by #718

Sinity:
## Summary

Restores the `push` / `pull_request` CI triggers that were **accidentally removed** by #718, re-enabling automatic test/lint/nix gating on `master` and PRs.

## Problem

#718 (`refactor(cli): remove foreground resource policy`) — a PR whose scope was entirely CLI resource policy, with **no mention of CI** — silently reverted the trigger blocks on four workflows, leaving each with only `workflow_dispatch:` (manual-only):

| Workflow | Before #718 | After #718 |
|----------|-------------|------------|
| `ci.yml` (lint + test ×3 Python) | push + pull_request | **dispatch only** |
| `nix.yml` (nix build + flake check) | push + pull_request | **dispatch only** |
| `actionlint.yml` | pull_request | **dispatch only** |
| `dep-audit.yml` | push + pull_request + weekly schedule | **dispatch only** |

Consequence: the CI test gate has not run automatically since **2026-05-02** (the last `ci.yml`/`nix.yml` runs are on `9029a37a`). Across the ~20 commits merged since, **~61 unit-test failures accumulated undetected** — see #1765. This is the root cause of the test debt.

(`pr-policy.yml` and `proof-comment.yml` were also gutted by #718, but those stay deleted — that was the intentional ceremonial-verification cleanup in #1737, not collateral. `codeql.yml` retains a working trigger and is unaffected.)

## Solution

Restore the exact pre-#718 `on:` blocks (verified against `git show 8b2b0fae~1:.github/workflows/<f>.yml`):
- `ci.yml` / `nix.yml`: `push: [master]` + `pull_request: [master]` (opened/reopened/synchronize/ready_for_review)
- `actionlint.yml`: `pull_request: [master]` (paths: `.github/workflows/**`, `.github/actions/**`)
- `dep-audit.yml`: push/pull_request (paths: `pyproject.toml`, `uv.lock`, the workflow) + weekly `cron: 0 6 * * 1`

## Verification

```
python3 -c "import yaml; ..."  # all four on: blocks parse with push/pull_request restored
```

## ⚠️ Expected: re-enabling CI surfaces the pre-existing failures

Once merged, `ci.yml` will run on `master` and PRs and will be **red** until the inherited test failures are cleared. That is the intended signal, not a regression introduced here. The known failures are partly fixed in **#1765** (collection error, keyset, config, stale tests) and partly still open (a handful of real-bug candidates triaged in #1765's body). Recommended order: merge #1765, work the remaining real failures, then this fix's CI will go green. Merging this first is also fine — it just makes the existing red state visible.

## Summary by CodeRabbit

* **Chores**
  * Enhanced workflow triggers to automatically run validation checks on pull requests and commits targeting the master branch, improving consistency of continuous integration checks.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -2,6 +2,11 @@ name: actionlint
 
 on:
   workflow_dispatch:
+  pull_request:
+    branches: [master]
+    paths:
+      - ".github/workflows/**"
+      - ".github/actions/**"
 
 permissions:
   contents: read
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,15 @@ name: CI
 
 on:
   workflow_dispatch:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
 
 permissions:
   contents: read
--- a/.github/workflows/dep-audit.yml
+++ b/.github/workflows/dep-audit.yml
@@ -2,6 +2,20 @@ name: Dependency Audit
 
 on:
   workflow_dispatch:
+  push:
+    branches: [master]
+    paths:
+      - "pyproject.toml"
+      - "uv.lock"
+      - ".github/workflows/dep-audit.yml"
+  pull_request:
+    branches: [master]
+    paths:
+      - "pyproject.toml"
+      - "uv.lock"
+      - ".github/workflows/dep-audit.yml"
+  schedule:
+    - cron: "0 6 * * 1"
 
 permissions:
   contents: read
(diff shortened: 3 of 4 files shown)


Sinity:
**Holding this PR (do not merge yet) — GitHub Actions minutes.**

The trigger restoration here is correct (it reverts the accidental #718 clobber), but re-enabling `push`/`pull_request` on `ci.yml`/`nix.yml` is only useful if Actions minutes are available. With minutes constrained, merging this would queue/fail a test×3-Python + nix run on every push and PR — pure noise. The local `devtools verify` + pre-push hook remain the effective gate in the meantime.

Options when minutes allow:
- merge as-is (full matrix on push + PR), or
- a cheaper variant: PR-only (drop the `push:` trigger), single Python version, or lint-only on PR with the full matrix on a weekly `schedule:`.

Leaving open so the fix is ready; the trigger config is the bug-free version regardless of when it's enabled.

PR #1798, opened 2026-06-09
fix(daemon): classify transient write-lock contention as retryable, not error

Sinity:
## Summary

Live batch ingest now distinguishes a transient `database is locked` write-lock
loss (self-healing, retried) from a real write failure, and records a
diagnosable reason in `ingest_attempts.error_message` instead of a source-path
dump.

> **Stacked on #1797** — shares the live batch ingest surface (`batch.py`,
> `test_live_watcher.py`). The diff currently shows both commits; once #1797
> squash-merges this rebases to the single `fix(daemon): classify transient
> write-lock contention…` commit.

## Problem

Two defects in the daemon's post-#1743 live ingest:

1. When an archive/insight write lost the write-lock race to a concurrent bulk
   write, `_ingest_full_records_archive` logged the `database is locked` failure
   at **WARNING with a full traceback** (`exc_info=True`) keyed by source path.
   The condition is self-healing — the path is re-queued and retried on the
   next cursor pass — so this floods journald with scary noise for a
   non-fault. (Journald even truncated the huge Rich traceback before the
   exception summary, making it useless *and* noisy.)
2. `finish_ingest_attempt` stored a `"; "`-joined list of up to three source
   **paths** in `ingest_attempts.error_message`. The diagnostic column carried
   no failure reason at all.

## Solution

- The per-record write loop now catches `sqlite3.OperationalError` separately
  and checks `is_transient_sqlite_lock`. Transient locks log at **DEBUG** with
  no traceback and record `"database is locked (retryable)"`; all other
  failures keep WARNING+traceback and record `str(exc)`.
- Per-record errors are collected on `_ArchiveFullWriteResult.errors`, threaded
  through `_FullIngestResult.materialization_errors`, accumulated per batch, and
  summarized by `_summarize_attempt_error` (deduped, capped at 3 distinct
  messages, `+N more` suffix). `error_message` now holds error text; with no
  captured text it stores a retry-path *count* — never a path dump.

## Verification

```
devtools test tests/unit/sources/test_live_watcher.py   # 66 passed
mypy --strict polylogue/sources/live/batch.py polylogue/sources/live/batch_support.py \
  tests/unit/sources/test_live_watcher.py   # clean
```

Two new tests monkeypatch `ArchiveStore.write_raw_and_parsed` to raise a lock
and a generic error, asserting the recorded `error` is
`"database is locked (retryable)"` and the real exception text respectively —
and never the source path.

### DIFF
--- a/polylogue/sources/live/batch.py
+++ b/polylogue/sources/live/batch.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import sqlite3
 import time
+import zipfile
 from collections.abc import Callable, Iterable
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
@@ -15,6 +16,7 @@ from json import loads as json_loads
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
+from polylogue.config import Source
 from polylogue.core.degraded import is_degraded
 from polylogue.core.memory import release_process_memory
 from polylogue.core.metrics import (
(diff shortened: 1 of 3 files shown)


Sinity:
Closing as **superseded by #1853** (merged 2026-06-13, `78954aaf0`). The core intent here — classify transient SQLite write-lock contention as retryable rather than error — is implemented on master in `polylogue/sources/live/sqlite_locking.py:is_transient_sqlite_lock()` (matches `database is locked` / `table is locked` / `database is busy`), wired through the live-ingest retry path that #1853 hardened against write thrash and retry storms. This branch is from 2026-06-09 and predates that work. Reopen if there's residual batch-level behavior in `batch.py`/`batch_support.py` not covered by #1853.

Issue #1810, opened 2026-06-12
Schema-v1 terminology sweep: session/origin only


Sinity:
# Schema-v1 terminology sweep: public `session` / `origin` only

## Purpose

Finish the public vocabulary migration. Polylogue’s public model is:

```text
session = normalized archived conversation/run/thread unit
origin  = normalized source family, e.g. claude-code, chatgpt, codex
```

Public CLI/API/web/MCP/docs must not expose `provider` or `conversation` as Polylogue concepts. Raw parser/capture code may preserve third-party names only when describing raw input material.

## Current reality

Recent work already removed much of the worst public drift:

- `ingest` public command was renamed to `import` in the first #1815 slice.
- `provider_meta` is no longer present in the main parser-output model; remaining occurrences are mostly raw-capture internals or historical/prose contexts.
- Public docs still need a deliberate sweep because stale words recur in generated docs, examples, tests, and old command comments.

The implementation must distinguish:

```text
public normalized vocabulary     -> session/origin only
raw third-party export vocabulary -> allowed in parser fixtures/comments when exact source format requires it
internal daemon process language  -> ingestion/ingests may describe daemon work, not public command names
historical migration docs         -> allowed only if marked historical
```

## Files and surfaces to inspect

```text
polylogue/sources/
polylogue/pipeline/
polylogue/archive/
polylogue/api/
polylogue/daemon/
polylogue/mcp/
polylogue/cli/
polylogue/browser_capture/
docs/
README.md
tests/
```

## Implementation slices

1. Add or update a stale-term audit with explicit allowlists for raw fixture/source-format contexts.
2. Sweep public CLI help, docs, schemas, DTOs, web routes, MCP tools, and Python API names.
3. Keep raw provider names only at parser/capture boundaries and mark them as raw input vocabulary.
4. Replace `conversation_id` in public-visible code with `session_id`; keep provider-native IDs under explicit raw/native names.
5. Ensure generated docs and snapshots are regenerated after command/docstring changes.
6. Add regression tests so new public surfaces cannot reintroduce `provider`/`conversation` names.

## Required tests

- Stale-term audit over public paths.
- Parser fixture tests proving normalized `origin` and `session_id` fields.
- API/MCP schema tests proving public DTOs use session/origin terminology.
- CLI help snapshot tests after command/doc updates.
- Negative audit test: `provider`/`conversation` allowed only inside explicit raw/third-party contexts.

## Acceptance criteria

[...]

Issue #1775, opened 2026-06-01
Stabilize devtools verify --all under xdist load


Sinity:
## Problem

`devtools verify --all` is currently not a stable completion gate under the full xdist run. During schema-v1 #1743 work, repeated broad full runs failed on different load-sensitive tests, while exact reruns of the failed nodes passed.

This creates a bad verification boundary: the normal PR baseline (`devtools verify`) is green, but the full diagnostic suite can return red for timing, xdist, database-lock, benchmark-budget, or UI/server timeout reasons that do not reproduce when isolated.

## Evidence

On branch `feature/refactor/schema-v1-safety-net` at `d59b8ff0`, `devtools verify --all` failed after the static/generated gates passed:

- `tests/benchmarks/test_fts_trigger_amplification.py::test_content_blocks_fts_trigger_growth_is_subquadratic`
- `tests/unit/core/test_schema_privacy.py::TestSafeValueNeverResemblesPII::test_urls_always_rejected`
- `tests/unit/storage/test_scale.py::TestPerformanceBudget::test_fts_search_budget`
- `tests/unit/cli/test_machine_contract.py::TestExtractCommand::test_extract_command_contract`
- `tests/unit/storage/test_scale.py::TestPerformanceBudget::test_get_many_performance_budget`
- `tests/unit/archive/test_filter_executor.py::TestFilterChainProperties::test_sql_pushdown_params_never_crashes`
- `tests/unit/daemon/test_web_reader.py::TestQueryNoResultsDiagnosticPath::test_search_with_bad_date_returns_query_error`
- `tests/unit/ui/test_tui.py::test_browser_node_selection`
- `tests/unit/sources/test_null_guard_properties.py::TestGeminiNonePartsProperty::test_gemini_text_extraction_never_crashes`

Exact rerun of those 9 nodes passed together:

```bash
pytest tests/benchmarks/test_fts_trigger_amplification.py::test_content_blocks_fts_trigger_growth_is_subquadratic \
  tests/unit/core/test_schema_privacy.py::TestSafeValueNeverResemblesPII::test_urls_always_rejected \
  tests/unit/storage/test_scale.py::TestPerformanceBudget::test_fts_search_budget \
  tests/unit/cli/test_machine_contract.py::TestExtractCommand::test_extract_command_contract \
  tests/unit/storage/test_scale.py::TestPerformanceBudget::test_get_many_performance_budget \
[...]

Issue #1003, opened 2026-05-10
live.watcher: gate ingest loop on schema-version preflight to prevent IOPS storm on stale runtime


Sinity:
## Problem

Observed on `sinnix-prime` 2026-05-10: polylogued (PID 2712788) sustained ~7000 reads/s and 80–90% NVMe utilisation for 15h, accumulating ~1 TB in `read_bytes` from `~/.local/share/polylogue/polylogue.db`. PSI `io full avg10=80%` system-wide — interactive work degraded.

Root cause: the installed runtime expected `SCHEMA_VERSION=9` while the on-disk DB was at v12 (newer build had migrated it). Each filesystem event in `~/.claude/projects/` and `~/.codex/sessions/` triggered a batch attempt that:

1. Opened the DB (multiple connections — `lsof` showed 20+ fds on `polylogue.db`)
2. Hit `_ensure_schema` → `version_mismatch` → `DatabaseError` from `polylogue/storage/sqlite/schema.py:152-153`
3. Got logged once per batch at `polylogue/sources/live/batch.py:232-233` as a generic warning
4. Retried on the next file change (no backoff, no circuit, no health degradation)

Under memory pressure (Rust build concurrent), DB header pages were evicted between attempts → physical NVMe reads instead of cache hits. 4K-avg request size × 7000/s = the IOPS-bound storm.

## Why the existing handling is wrong

- `except Exception as exc` in `batch.py` cannot distinguish \"transient parse error on one file\" from \"database is structurally unusable for this binary.\" The first deserves retry; the second is permanent until the binary or the DB changes.
- There is no preflight: the daemon successfully starts, opens HTTP on :8765, and reports healthy via systemd, but is incapable of ingesting anything.
- `daemon/health.py` already has a tiered alert system (`HealthSeverity.CRITICAL`, `HealthTier.FAST`) — schema-version mismatch is exactly the kind of CRITICAL/FAST liveness condition it is shaped for, but it is not wired in.

## Proposed design

**1. Preflight on startup (`daemon/cli.py` `run` path):**
Open the DB read-only, read `PRAGMA user_version`, compare against `SCHEMA_VERSION`. On mismatch:
- Emit a CRITICAL `HealthAlert` (`check_name=\"schema_version\"`, tier=FAST) with both versions and a hint (\"runtime is stale; rebuild/redeploy polylogue or rollback DB\").
[...]

PR #1877, opened 2026-06-14
fix(ci): restore green test gate (daemon deadlock + masked failures)

Sinity:
## Summary

Restores a working `test` CI gate for Polylogue. The gate was non-functional in
two compounding ways: a daemon-shutdown **deadlock** hung every `test` job to the
90-minute ceiling (so recent merges all rode `cancelled` test jobs), and that
hang **masked pre-existing failures** in everything that sorts after
`tests/unit/daemon`. This PR fixes the deadlock and every now-visible failure.

## Problem

1. **Deadlock.** `run_daemon_services`'s `finally` called `server.shutdown()`
   synchronously on the event-loop thread. `socketserver.shutdown()` blocks on an
   Event that only `serve_forever` sets — but `serve_forever`'s `to_thread` task
   can never start while the loop thread is blocked, so the Event is never set.
   The timeout dump shows only `Stack of MainThread` parked in `shutdown()`, no
   `serve_forever` worker. CI carried no per-test timeout, so it ran to the
   90-minute kill.
2. **Masked failures.** Because the suite died at ~45% (deep in
   `tests/unit/daemon`), everything after it never ran in CI and accumulated
   failures. Once the suite could complete, the full coverage gate enumerated the
   exact set; all are now resolved.

## Solution

**Deadlock + hang guard**
- `polylogue/daemon/cli.py`: `_shutdown_server_if_serving` is now async and runs
  `server.shutdown()` via `await asyncio.to_thread(...)`, freeing the loop to
  start the `serve_forever` worker. Both call sites `await`.
- `pytest-timeout` added; `devtools coverage-gate` passes
  `--timeout=600 --timeout-method=thread`; the two full-path daemon tests carry
  `@pytest.mark.timeout(30)`. A future hang now fails fast with a thread dump.

**Real product/contract bug**
- `polylogue/cli/output_assurance.py` (+ regenerated `docs/cli-reference.md`):
  added missing `OutputAssurance` entries for the `analyze`/`mark` verbs (#1872),
  which the assurance-matrix contract test requires.

**Stale test expectations from intentional, never-run-in-CI changes**
- CLI help snapshots regenerated (added analyze/mark; additive only).
- MCP query-term tests updated to the shared-compiler split contract (#1860).
- devtools verify-shape tests updated to current composition (#1775 lane split).
- CLI enumeration/`delete_verb` signature tests; visual reader deeplink
  (`getConvIdFromURL` → `getSessionIdFromURL`, #1810).
- `test_session_insight_profiles.py` codex-topic test decoupled from the checkout
  directory name: it now seeds a deterministic fake repo (`tmp_path/polylogue/.git`)
  so repo-name inference (which probes the real FS for a `.git` root) resolves
  `polylogue` in CI and in agent worktrees alike.

**Correction within this PR**
- An earlier commit on this branch removed the FTS readiness poison-recompute in
  `fts_status.py`, mislabeling it a #1003 "never scans" violation. On review it is
  intended, tested behavior: the expensive recompute fires *only* in the rare
  poisoned `ready|0|0`-over-populated-index window (detected via a cheap `LIMIT 1`
  existence probe), never on healthy probes. That commit is reverted; the narrow
  "recompute each probe during the poison window vs. heal the freshness record
  once" tension is real but separate, and is tracked rather than resolved by
  deleting tested behavior.

**CI reliability**
- `coverage-gate` ignores `tests/benchmarks` (performance tests with their own
  `--benchmark-enable` runner; flaky under random-ordered coverage due to
  cross-test pollution — tracked in #1878; no unique coverage lost).

## Verification

- `uv run pytest tests/unit/daemon/test_daemon_cli_remote_bind.py` → 9 passed in
  0.75s (previously hung indefinitely).
- Full local `devtools coverage-gate` (exactly what CI runs) → enumerated 2
  failures (fts poison-recompute regression + checkout-coupled codex-topic test);
  both fixed; `10718 passed`, coverage 83.59% (floor 82%).
- `devtools verify --quick` → exit 0 (format, lint, mypy --strict, render-all, static gates).
- This PR's `test` jobs across 3.11/3.12/3.13 are the authoritative green proof.

Ref #1878

## Summary by CodeRabbit

* **New Features**
  * Added `analyze` and `mark` CLI commands to the interface.

* **Bug Fixes**
  * Improved test reliability with per-test timeout guards to prevent hangs.
  * Optimized freshness state checks to avoid unnecessary table scans.
  * Fixed delete verb behavior to properly handle force flag in dry-run mode.

* **Documentation**
  * Updated CLI reference to document new `analyze` and `mark` commands.


Sinity:
Addressed the Codex P2 on `daemon/cli.py` (`d03b25d27`): `server.shutdown()` now runs in a dedicated single-worker `ThreadPoolExecutor` instead of `asyncio.to_thread`. The original fix could re-deadlock under default-executor pressure (both `serve_forever` calls + an embedder `max_workers=1` or maintenance `to_thread` jobs) by queuing shutdown behind the `serve_forever` it must stop; the dedicated executor guarantees shutdown runs immediately and sets socketserver's flag. Teardown uses `wait=False` so a stuck shutdown never blocks, with the caller's `server_close()` as the socket backstop. 24 daemon shutdown/bind tests pass; mypy clean.

pip-audit failures are pre-existing dependency CVEs (idna/pip/pyjwt) unrelated to this PR — a dependabot/lockfile concern, advisory here.

Sinity:
Addressed both CodeRabbit findings (`26c4cd451`):
- **Major (cli.py shutdown):** correct — `ThreadPoolExecutor.shutdown(wait=False)` joins its worker at interpreter exit, so a wedged `server.shutdown()` would still hang. Switched to a dedicated **daemon** thread that signals completion to the loop via `call_soon_threadsafe`; it's never queued behind `serve_forever` (the Codex P2 concern) and is abandoned at exit if genuinely stuck (your concern).
- **Nitpick (test):** applied — create the intermediate `polylogue/` dir for defensive robustness.

Issue #1818, opened 2026-06-12
Enforce single machine-output contract: JSON finite, NDJSON streaming/bulk


Sinity:
# Machine-output contract: finite JSON, NDJSON streams, error and mutation envelopes

## Purpose

Enforce one machine-output contract for the final command/action/API/MCP surfaces:

```text
--format json   => exactly one finite JSON document on stdout
--format ndjson => one JSON object per line on stdout
human progress/warnings/hints => stderr
errors => stable JSON error envelope when machine output requested
ANSI => forbidden in machine output
```

## Current reality

The old body overstates the need to invent all envelopes from scratch. Recent recon found:

- list/search/detail success envelopes are already broadly aligned across CLI/daemon/MCP;
- divergence remains in **error envelopes** and **mutation envelopes**;
- `read --view messages --format json` has emitted NDJSON-like output, which violates finite JSON semantics;
- contracts should be applied after #1842 prunes the command zoo and #1816 defines action contracts.

Sequencing: #1842 Phase A first, #1816 second, this issue third. Otherwise this issue spends effort specifying outputs for commands that will be deleted.

### Result-set envelope

```json
{
  "status": "ok",
  "mode": "results",
  "query": {},
  "items": [],
  "total": 0,
  "limit": 20,
  "cursor": null,
  "next_cursor": null,
  "caveats": []
}
```

### Singleton/read envelope

```json
{"status":"ok","mode":"session","query":{},"item":{}}
```

### Mutation envelope

```json
{"status":"ok","operation":"mark.tag.add","matched":3,"changed":3,"dry_run":false,"caveats":[]}
```

### Error envelope

```json
{"status":"error","code":"ambiguous_selection","message":"...","query":{},"details":{}}
```

## Implementation slices

1. Enumerate surviving actions after #1842.
2. Add common success/error/mutation envelope DTOs or schema helpers.
3. Fix known finite-json violations such as `read --view messages --format json`.
4. Ensure human warnings/progress are routed to stderr in machine-output modes.
5. Add schema files or generated schemas for machine envelopes.
6. Wire CLI, daemon HTTP, MCP, and Python facade to the same DTO vocabulary where applicable.
7. Add regression tests for empty result, success result set, singleton read, mutation dry run, destructive confirmation, unsupported query, ambiguity, and interrupted stream.

## Required tests

- Every declared JSON action emits exactly one JSON document.
- Every declared NDJSON stream emits one object per line and no banners.
[...]

PR #1897, opened 2026-06-15
fix(cli): messages --format ndjson emits JSON-per-line, not markdown (#1818)

Sinity:
## Summary
`read --view messages --format ndjson` emitted human markdown lines instead of JSON; it now emits one self-contained JSON document per line, matching the repo-wide ndjson contract (#1818 finiteness).

## Problem
#1818 requires `--format json` → a single finite JSON value; `--format ndjson` → one JSON document per line. Investigating the reported "messages --format json emits N documents" violation: it did **not** reproduce — the json path (`cli/messages.py:run_messages`) already emits one finite envelope `{session_id, messages:[...], total, limit, offset}` parseable by a single `json.loads`. But the **ndjson** side of the same view was broken: it fell through to the `else:` markdown branch (`[role type] text` lines), not JSON at all, contradicting the format help and `query_output.py`'s ndjson definition.

## Solution
`cli/messages.py`: added an explicit `elif fmt == "ndjson":` branch emitting one JSON document per message (each line self-contained with `session_id`), factoring the per-message dict into `_message_document()` shared by the json and ndjson branches. The finite-json envelope is unchanged; non-JSON formats untouched.

## Verification
```
devtools test tests/unit/cli/test_messages.py   # 5 passed (2 new)
devtools verify --quick   # exit 0 (ruff/mypy/render-all --check)
```
New tests pin both halves of the contract: json parses as one finite document; ndjson yields N independently-parseable lines.

Ref #1818

## Summary by CodeRabbit

* **New Features**
  * Added `--format ndjson` option for streaming one JSON object per message line with session identifiers included.

* **Bug Fixes**
  * JSON output format now emits raw JSON directly to stdout for improved output handling.

* **Tests**
  * Enhanced test coverage for JSON and NDJSON output formats with stdout capture assertions.

### DIFF
--- a/polylogue/cli/messages.py
+++ b/polylogue/cli/messages.py
@@ -61,25 +61,35 @@ def run_messages(
 
             fmt = output_format or "markdown"
 
+            def _message_document(m: object) -> dict[str, object]:
+                return {
+                    "id": str(m.id),  # type: ignore[attr-defined]
+                    "role": str(m.role),  # type: ignore[attr-defined]
+                    "message_type": str(m.message_type.value),  # type: ignore[attr-defined]
+                    "text": m.text or "",  # type: ignore[attr-defined]
+                }
+
             if fmt == "json":
                 import json as _json
 
+                # Finite machine-output contract (#1818): one JSON value.
                 payload = {
                     "session_id": session_id,
-                    "messages": [
-                        {
-                            "id": str(m.id),
-                            "role": str(m.role),
-                            "message_type": str(m.message_type.value),
-                            "text": m.text or "",
-                        }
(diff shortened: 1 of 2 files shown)


Sinity:
Fixed in the latest push: machine output (json + ndjson) now goes through raw `click.echo` instead of `env.ui.print`, so Rich never interprets `[bold]`-style sequences inside message text — the emitted bytes match `json.dumps` exactly. Added markup-survival assertions (text `[bold]first[/bold]` round-trips byte-for-byte through both json and ndjson). Note: the pre-existing json branch had the same latent issue; it's fixed too.

Issue #1842, opened 2026-06-12
Polylogue: final hard-pruned CLI surface


Sinity:
# Finish the public CLI command floor without losing capability

## Purpose

Polylogue's public CLI should stop teaching a flat command zoo. The durable product model should be:

```text
polylogue find QUERY
polylogue find QUERY then read|mark|analyze|delete|continue
polylogue 'QUERY' then ACTION        # quoted shorthand only
polylogue import PATH...
polylogue config ...
polylogue ops ...
```

This is a command-surface cleanup, not a compatibility project. Do not keep old roots as hidden aliases. Do not add deprecation shims. A command disappears only when its capability has a real home, or when the issue/PR explicitly records that the capability is being cut as non-product.

## Current state

This issue was closed prematurely. Several real slices landed, but the live root command tree still exposes retained/deferred command families and the final floor is not done.

Current master root commands are approximately:

```text
analyze blackboard config continue cost dashboard delete feedback import init insights mark ops read resume resume-candidates select tags tutorial user-state
```

PR #2140 moves the root `diagnostics` group under `polylogue ops diagnostics` and moves archive-facing workload/space diagnostics out of flat `devtools` commands. After that lands, `diagnostics` should no longer be a retained product root.

## Landed work that remains part of this issue

These slices are done and should not be reimplemented:

- `find` keyword, strict bare-query floor, and dead `export.py` removal: #1885.
- `facets` absorbed into `analyze --facets [--no-idf]`: #1891.
- `neighbors` absorbed into `read --view neighbors`: #1893.
- `correlate` absorbed into `read --view correlation`: #1894.
- `near:id:<ref>` session-seeded similarity plumbing/execution: #1899 and #1904.
- `context` and `context-pack` absorbed into read views, fixing the old passthrough bug: #1909.
- Query/action contracts and machine-output envelope work that constrain the surviving surfaces: #1816/#1818 follow-up PRs.
- Devtools command-sprawl regrouping lives under #2126; only product CLI consequences belong here.

## Remaining command families

| Surface | Target decision |
| --- | --- |
| `cost` | Retain top-level unless an actually equivalent billing/subscription workflow exists elsewhere. It is not just result-set analysis. |
| `insights` | Retain until transform/status/read-model ownership is redesigned. Do not delete it merely because `analyze` exists. |
[...]

PR #1899, opened 2026-06-15
feat(query): near:id:<ref> session-seeded similarity plumbing (#1842)

Sinity:
## Summary
Adds the `near:id:<ref>` query form (session-seeded similarity) through the DSL compiler, spec, plan, and field registry, with a typed-rejection guard at execution. This is the plumbing half of the operator's "both" decision for #1842 PR-3b; the stored-embedding execution path is a tracked follow-up.

## Problem
`near:` only supported free text (`near:"prompt"` → `similar_text`, re-embedding a string). There was no way to query "sessions similar to session X" by its stored embedding.

## Solution
- `archive/query/expression.py`: `near:` value of form `id:<ref>` (conservative: only the unambiguous `id:` prefix) → `similar_session_id`; everything else stays free-text → `similar_text`; empty ref raises `ExpressionCompileError`.
- Threaded `similar_session_id` through `spec.py`, `plan.py`, and added a parallel `QueryFieldDescriptor` in `fields.py` (required by the field-catalog parity test).
- `archive_execution.py` / `plan_execution.py`: `_reject_unexecutable_session_seed` guard — until the execution path lands, a session-seeded query fails **typed** (`ExpressionCompileError`, `field="near"`, referencing #1842), never silently broadening to a full listing.

### Follow-up (out of scope, flagged by the implementer)
Execution needs a new `VectorProvider.query_by_session(session_id, limit)` primitive in `protocols.py` + `storage/search_providers/sqlite_vec_queries.py` (fetch the seed's `message_embeddings`, run a vec0 MATCH excluding the seed's own messages, fail typed when unembedded). The task's STOP clause kept this PR within `archive/query/`. `near:@latest` also deferred (needs an execute-time DB read).

## Verification
```
devtools test test_query_expression test_near_session_seed_execution test_query_descriptor_parity test_query_fields   # 144 passed
devtools test test_read_surface_coherence test_facade_contracts test_query_search_runtime test_query_exec_laws         # 304 passed
devtools verify --quick   # exit 0
```

Ref #1842


Sinity:
Fixed in the latest push: `search_hits_for_plan` now calls `_reject_unexecutable_session_seed(plan)` before the `plan_has_search_hit_evidence` check, so a `near:id:` plan fails typed there too instead of silently returning no hits. Added a regression test (`test_search_hits_for_plan_rejects_session_seed_not_silently_empty`) covering this exact path. Good catch — this was a real hole in the 'fail typed everywhere' contract.

PR #1901, opened 2026-06-15
refactor: drop provider vocabulary from public read surfaces (#1810)

Sinity:
## Summary
Removes `Provider`-token vocabulary from the public CLI/MCP/daemon read surfaces, leaving them origin-only (#1810). Provider-wire surfaces (schema packages) are intentionally retained.

## Problem
`Provider` (lab/product-wire identity) leaked onto public read surfaces that should speak `Origin` (source-origin: claude-code-session, chatgpt-export, …). #1810's anti-goal: no provider wording on public filters/payloads.

## Solution (3 sites)
- `cli/archive_query.py`: dropped the dead `provider`/`exclude_provider`→origin fallback in `_resolve_origins`/`_resolve_excluded_origins` (no `--provider` flag exists on the query surface — the fallback was API/dead). Public root query is origin-only.
- `mcp/server_insight_tools.py`: provider param → origin on the public read tool.
- `daemon/http.py`: provider facet label → origin (label-only change).
- `core/sources.py`: small helper for the origin/provider boundary mapping where internal contracts still need a provider token.
- Added `tests/unit/core/test_public_surface_origin_vocabulary.py` — guard asserting no provider vocabulary on these public read surfaces.

**Retained (correctly):** `schema.py`'s `--provider` flags — schema generate/promote/audit operate on provider-wire schema packages, a #1810-exempt surface.

## Verification
```
devtools test test_archive_query test_public_surface_origin_vocabulary   # 116 passed
devtools verify --quick   # exit 0 (mypy --strict, render-all --check, ...)
```

Ref #1810

### DIFF
--- a/polylogue/cli/archive_query.py
+++ b/polylogue/cli/archive_query.py
@@ -36,7 +36,6 @@ from polylogue.cli.shared.helpers import load_effective_config
 from polylogue.cli.shared.machine_errors import error_no_results
 from polylogue.cli.shared.types import AppEnv
 from polylogue.config import Config
-from polylogue.core.enums import Provider
 from polylogue.paths import archive_file_set_root_for_paths
 from polylogue.storage.search_providers import create_vector_provider, reciprocal_rank_fusion
 from polylogue.storage.sqlite.archive_tiers.archive import (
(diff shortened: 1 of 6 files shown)


Sinity:
Fixed in the latest push: `source_name_to_origin` now maps source-family tokens (`gemini-export`, `drive-takeout`, …) via `_FAMILY_TO_PROVIDER` *before* the provider-wire fallback, so those sessions project to their real origin (e.g. aistudio-drive) instead of `unknown-export`. Added a regression test asserting family tokens resolve to the same origin as their provider equivalent and never to `unknown`. Good catch.

Issue #1830, opened 2026-06-12
Tiered backup, blob GC, and archive performance artifacts


Sinity:
# Tiered backup, blob GC, and archive performance artifacts

## Purpose

Make Polylogue’s local archive durable and inspectable: tier-aware backup/export, safe content-addressed blob GC, and performance artifacts that measure import/search/write behavior without turning CI into a wallclock lottery.

## Current model

Polylogue uses multiple SQLite tiers and a blob store:

```text
source.db       raw acquisition / rebuild root
index.db        parsed sessions/messages/blocks/FTS/read models
embeddings.db   semantic vectors/status
user.db         user overlays / future KV assertions
ops.db          daemon cursors / transient ops state
blob/           content-addressed large payloads
```

Backups and GC must respect the durability class of each tier. `user.db` is irreplaceable; index/embeddings/ops are rebuildable to different degrees; blobs are source evidence when referenced.

## Implementation slices

1. Document canonical archive root and blob root semantics; fix any mismatch between code and docs.
2. Add backup profiles: full evidence backup, user-overlays backup, rebuildable-cache exclude, diagnostics bundle.
3. Add blob GC generation rules using `blob_refs`, `pending_blob_refs`, and source/index reachability.
4. Add dry-run and report modes before destructive GC.
5. Add benchmark artifact scripts for import/search/embedding/write amplification, storing results as artifacts not CI pass/fail wallclock gates.
6. Connect write-amplification findings from #1851 into the performance report.

## Required tests

- Backup contains `user.db` and all referenced blobs.
- GC never deletes pending or referenced blobs.
- Dry-run GC reports exact candidate counts and refs.
- Restore smoke test into an isolated archive root.
- Benchmark scripts produce machine-readable artifacts with environment metadata.

## Acceptance criteria

- [ ] Backup/restore story is documented and tested.
- [ ] Blob GC is generation-safe and dry-run-first.
- [ ] Performance artifacts exist for import/search/write paths without CI wallclock gates.
- [ ] `user.db`/KV future is treated as irreplaceable in backup docs.

## Non-goals

- Do not move off SQLite without a measured bottleneck.
- Do not make benchmark wallclock thresholds required CI gates.
- Do not prune blobs based only on filesystem age.

## PR handoff checklist

Every implementing PR must include this exact information in the PR body:

```text
Issue slice:
Files inspected:
Files changed:
Old surface removed or retained:
Contracts/tests added:
Generated artifacts updated:
Verification commands and results:
[...]

PR #1963, opened 2026-06-15
feat(backup): support named archive profiles (#1830)

Sinity:
## Summary

Adds named backup profiles to the concrete `polylogue backup` execution path so the profile taxonomy documented by `maintenance backup-plan` can drive actual archive tier selection.

## Problem

#1830 had documented backup semantics and a planning surface, but the runtime backup command still always copied the historical source/user/embeddings set and omitted index/ops. Operators could inspect profiles but could not execute them.

## Solution

`polylogue.daemon.backup` now accepts a `profile` argument with four named profiles: `rebuildable_cache_exclude` (default, preserving existing behavior), `full_evidence`, `user_overlays`, and `diagnostics_bundle`. The manifest records `profile`, `included_tiers`, and `omitted_tiers`, verification reads those manifest fields, and `polylogue backup --profile ...` passes the selected profile through to the archive backup operation.

This slice intentionally keeps blob copying tied to profiles that include `source.db`; there is not yet a distinct user-owned blob reference index to make `user_overlays` copy only human-owned blobs truthfully.

## Verification

- `uv run devtools test tests/unit/daemon/test_backup.py tests/unit/cli/test_backup_command.py` → 9 passed
- `uv run mypy polylogue/daemon/backup.py polylogue/cli/commands/backup.py tests/unit/daemon/test_backup.py tests/unit/cli/test_backup_command.py` → success
- `uv run devtools verify --quick` → exit_code 0

### DIFF
--- a/polylogue/cli/commands/backup.py
+++ b/polylogue/cli/commands/backup.py
@@ -6,7 +6,7 @@ from pathlib import Path
 
 import click
 
-from polylogue.daemon.backup import backup_archive, format_backup_result
+from polylogue.daemon.backup import BACKUP_PROFILES, BackupProfile, backup_archive, format_backup_result
 from polylogue.logging import configure_logging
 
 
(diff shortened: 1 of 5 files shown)


Sinity:
Review resolution update:

- CodeRabbit: `full_evidence` should remain usable when optional `ops.db` is absent. Fixed in 8fc4de45 by treating profile-optional tiers as non-fatal and omitting absent optional tiers from the copied set/manifest. Added `test_backup_archive_full_evidence_profile_treats_ops_as_optional`.
- CodeRabbit: non-default profile omissions should not be labelled rebuildable/disposable. Fixed in 8fc4de45 by keeping the rebuildable/disposable suffix only for `rebuildable_cache_exclude`; other profiles now render `Omitted by profile`. Added `test_backup_result_formats_non_default_omissions_neutrally`.
- CodeRabbit: backup help baseline drift. Fixed in b23b1c91 by regenerating `tests/baselines/showcase/help-backup.txt` and verifying `devtools lab-scenario verify-baselines`.

Local verification after fixes: backup focused tests passed (11), mypy passed on touched backup files, and `uv run devtools verify --quick` exited 0.

Issue #1849, opened 2026-06-12
Replace proof theater with code-coupled tests


Sinity:
## Purpose
Replace ceremonial proof/QA/showcase confidence layers with code-coupled tests, fixture-world tests, and benchmark artifacts that catch real regressions. A test or artifact is valuable only if it fails when shipped behavior drifts.

## Rationale
Polylogue accumulated several reporting/proof surfaces that looked like rigor but could drift away from implementation. The useful pieces should survive as executable contracts, generated-surface checks, fixture worlds, schema validation, and benchmark artifacts. The rest should be deleted or internalized.

This is also the home for the fanout rule: when a change requires many hand-authored updates for the same fact, prefer generating one side, deriving docs/tests from the owning registry, or deleting stale prose. Do not normalize accidental duplication as a permanent planning system.

## Current implementation reality
A lot of the old proof layer has already been removed or replaced. Current useful executable surfaces include:

- `devtools render-all --check` and generated-surface checks;
- topology/layering/manifest/doc-command/lane-assertion gates;
- `ACTION_CONTRACTS` and contract-derived CLI behavior tests (#1816);
- generated CLI output schemas and live fixture-output validation;
- demo fixture-world convergence tests (#1843);
- recovery transform benchmark/render artifacts (#1880);
- benchmark-campaign comparison smoke tests;
- public-surface stale-vocabulary audits.

The previous live cleanup pressure around `polylogue/cli/output_assurance.py` is resolved: the old hand-authored output-assurance registry has been deleted. The durable lesson is behavioral, not file-name-based: do not recreate static confidence ledgers that are detached from executable command, schema, fixture, benchmark, or generated-surface checks.

The remaining work is to delete or fold stale proof/QA/showcase vocabulary where it is public-facing, redundant, or detached from behavior. Internal devtools/lab terms are fine when they name an executable gate or fixture harness; they are noise when they become a second non-executable product story.

## Evidence-substrate cleanup rule
The new evidence primitives are subject to the same anti-ceremony test as proof tooling. Keep an evidence artifact only if a shipped workflow consumes it for drilldown, recovery, query, web/API display, MCP/API parity, missed-review/context detection, or user-state lifecycle. Delete or fold generated proof-like evidence reports that merely demonstrate internal consistency without changing behavior.

## Keep / Delete Rules
Keep a verification artifact only when it is wired to code and would fail on real drift:

- Keep generated docs/schemas/topology when they are rendered from source and checked by `render-all --check`.
[...]

PR #1965, opened 2026-06-15
test: register archive maintenance benchmarks (#1830)

Sinity:
## Summary

Adds an archive-maintenance benchmark campaign covering backup-plan, blob-GC dry-run, and archive space-report paths, with catalog and operation-spec coverage.

## Problem

#1830 had backup profile semantics and planning surfaces, and #1849 had benchmark infrastructure, but archive maintenance still lacked a durable benchmark artifact lane tying backup/GC/space-report behavior into the campaign registry.

## Solution

Registers an `archive-maintenance` benchmark campaign, adds benchmark coverage for the maintenance paths, declares the benchmark operations in the artifact graph, and refreshes the campaign coverage/generated quality docs. The blob-GC dry-run benchmark asserts blob files are preserved, so the benchmark also guards the non-destructive proof boundary.

## Verification

- `uv run devtools test tests/unit/devtools/test_benchmark_campaign.py tests/unit/operations/test_specs.py tests/benchmarks/test_archive_maintenance.py -q` → `18 passed`
- `uv run devtools verify --quick` → passed

Ref #1830
Ref #1849


Sinity:
Review fix pushed in 876dc59a: `ScenarioProjectionSource.to_projection_entry()` now propagates `maintenance_targets`, with regression coverage for the archive-maintenance campaign projection and generic scenario projection metadata. Regenerated `docs/test-quality-workflows.md` so the campaign row now carries the maintenance target instead of dropping it.

Issue #1883, opened 2026-06-14
Unify user overlays as evidence-linked assertions


Sinity:
## Purpose
Collapse Polylogue's user and agent overlay mini-systems into one evidence-linked Assertion substrate. Assertion is the #1807 primitive for authored or transform-produced claims about a target ref.

## Rationale
Polylogue needs one lifecycle model for user/agent claims instead of a separate semantic subsystem for every overlay. Marks, annotations, corrections, suppressions, saved views, recall packs, workspaces, blackboard notes, transform candidates, RunState, decisions, caveats, blockers, and lessons all share the same deeper shape: an author asserts a value about a target, with evidence, status, visibility, staleness, and context policy.

An Assertion is:

```text
author/system/transform asserts value V about target ObjectRef T in scope S,
with kind, evidence refs, status, visibility, confidence, staleness, and context policy.
```

Raw transcripts, tool output, blobs, deterministic read models, and ObservedEvents do not belong in Assertions.

## Current implementation reality
The assertion substrate is now the live user-overlay storage path, not only a mirror table:

- `polylogue/storage/sqlite/archive_tiers/user.py` defines `USER_SCHEMA_VERSION = 2` and the canonical `assertions` table in `user.db`.
- #2058 removed the old canonical overlay mini-tables for marks, annotations, corrections, suppressions, saved views, recall packs, workspaces, and blackboard notes. Those surfaces now persist through assertion helpers.
- `session_tags` and `session_metadata` remain table-backed because their shape is still simple session metadata, not lifecycle claims.
- `polylogue/storage/sqlite/archive_tiers/user_write.py` owns assertion upsert/read/list helpers for the user overlay surfaces.
- `polylogue/storage/sqlite/archive_tiers/archive.py` exposes the archive-tier facade over assertion-backed marks, annotations, suppressions, corrections, saved views, recall packs, workspaces, and blackboard notes.
- Delete/status transitions landed for the public overlay deletes: deleted overlays become inactive assertion rows instead of lingering as live state.
- Saved views and workspaces preserve user-visible name uniqueness by retiring any previous active same-name assertion before writing the new one.
- Transform-candidate assertions are private, evidence-linked, `status=candidate`, and not injected into context by default.
- `polylogue/readiness/capability.py` reports assertion capability/readiness when the table is present.
[...]

Issue #1839, opened 2026-06-12
Polylogue: typed sidecar/blackboard notes with evidence and staleness


Sinity:
> **v8 issue-pack applied 2026-06-13.** This issue is the **typed sidecar/blackboard notes with evidence refs and staleness** within the v8 cluster *agent work packets, typed notes, workflow observability & evidence refs* (`PLG-05-agent-work-packets-notes-workflows-evidence.md`).
> Coordinated cluster siblings: #1838 #1839 #1840 #1845.
> The brief below is authoritative; where it conflicts with prior issue text, **v8 wins**. Source pack: `polylogue_sinex_issue_pack_v8_expanded_2026-06-13`.

# Polylogue agent work packets, typed notes, workflow observability, and evidence refs

Current GitHub issues covered: #1838, #1839, #1840, #1845.

## Purpose

Make Polylogue useful for real agent work, not only transcript search. The central product object is an **agent work packet**: a bounded, inspectable packet tying a session or workflow run to branch/diff refs, commands, checks, notes, claims, caveats, and outcome.

This gives future wiki/reporting ideas a grounded precursor without creating a new truth layer.

## Vocabulary

A **work packet** is a view/artifact over evidence. It is not canonical truth and it is not an issue-closure oracle.

A **typed sidecar note** is a bounded note from a human or agent scoped to a session, branch, workflow run, or packet. It has a kind, claim, evidence refs, suggested action, priority, and staleness condition.

A **workflow run** is an observed external orchestration artifact, such as Claude Code dynamic workflows. Polylogue records definitions/runs/phases/agent-runs only as evidence and analytics. It does not orchestrate workflows.

An **ObjectRef** is a tiny reference to an object in Polylogue, Sinex, Git, GitHub, a file, or an artifact. An **EvidenceRef** is an ObjectRef plus role/caveats/excerpt/check time. These are citation plumbing, not a new evidence core service.

## Target DTO sketches

```python
@dataclass(frozen=True)
class ObjectRef:
    kind: str              # polylogue.session, polylogue.message, git.commit, github.issue, file.range, artifact.blob
    id: str | None = None
    path: str | None = None
    anchor: Mapping[str, object] = field(default_factory=dict)

@dataclass(frozen=True)
class EvidenceRef:
    target: ObjectRef
    role: Literal['supports','contradicts','context','source','acceptance_criterion','implementation','test_result','caveat']
    excerpt: str | None = None
    caveats: tuple[str, ...] = ()
[...]

PR #1967, opened 2026-06-15
feat(storage): read user overlays through assertions (#1883)

Sinity:
## Summary

Moves user-overlay read parity forward by reading mirrored assertion rows for overlay metadata and routing blackboard list/read behavior through the assertion-backed adapter while preserving the public blackboard output shape.

## Problem

#1883 write-through work made marks, annotations, corrections, saved views, recall/workspace data, and blackboard notes mirror into assertion rows, but read paths could still bypass that substrate. #1839 specifically needed blackboard/sidecar notes to become evidence-bearing without breaking existing note IDs and public list output.

## Solution

Adds assertion-backed user-overlay readers in the user tier, gives mirrored assertion rows stable named IDs, and adapts `ArchiveStore.list_blackboard_notes()` to decode blackboard rows through the assertion-backed envelope. Focused tests cover overlay parity and the blackboard facade path.

The phase intentionally keeps legacy blackboard rows as the source of public `note_id` and target identity. Assertion-only blackboard notes are not exposed yet because the assertion row does not currently carry a first-class public blackboard note identity.

## Verification

- `uv run devtools test tests/unit/storage/test_archive_tiers_assertion_write_through.py tests/unit/storage/test_blackboard_facade.py` → `17 passed`
- `uv run devtools verify --quick` → `exit_code: 0`

Ref #1883
Ref #1839

## Release Notes

* **Improvements**
  * Blackboard notes now use a more consistent storage mechanism with improved data deduplication and integrity handling.
  * Enhanced note retrieval and listing with better support for legacy data compatibility.

### DIFF
--- a/polylogue/storage/sqlite/archive_tiers/archive.py
+++ b/polylogue/storage/sqlite/archive_tiers/archive.py
@@ -85,6 +85,7 @@ from polylogue.storage.sqlite.archive_tiers.user_write import (
     assertion_id_for_recall_pack,
     assertion_id_for_saved_view,
     assertion_id_for_workspace,
+    list_archive_blackboard_note_envelopes,
     mark_assertion_status,
     upsert_annotation,
     upsert_assertion,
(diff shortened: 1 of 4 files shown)


Sinity:
Review fix pushed in 290448e1: suppression reads now fall back to the legacy `suppressions.reason` when the mirrored assertion exists but has `body_text = None`, while still accepting structured `value.mode`. Added a regression for a value-only suppression mirror.

Issue #2006, opened 2026-06-16
feat(query): implement the full query DSL substrate


Sinity:
## Purpose
Implement Polylogue's full query DSL as one typed grammar, AST, metadata, and lowering pipeline over the existing archive engines. The Lark grammar in `polylogue/archive/query/expression.py` is the query language; compact field/text clauses and explicit Boolean predicates are entry shapes in that same grammar, not a "floor" language plus a separate future language.

This is a base product feature. Polylogue is an evidence cockpit; if users and agents cannot ask precise questions about sessions, messages, blocks, tools, actions, runs, reviews, assertions, lineage, and bundles, every higher-level view becomes a one-off report.

## Current implementation reality
The current implementation is real code and should be extended in place, not bypassed:

- `polylogue/archive/query/expression.py` owns the Lark grammar, field registry, structural unit registry, typed parser errors, AST/explain output, and the historical `compile_expression` public entrypoint. `compile_expression` is only a compatibility function name used by existing CLI/MCP/API callsites; it is not a separate compiler, grammar, or execution path. Its body must delegate to the canonical Lark parser, typed AST, and lowerer.
- `polylogue/archive/query/spec.py` defines `SessionQuerySpec`, query planning, and the shared query selection contract used by current read surfaces.
- `polylogue/archive/query/predicate.py` carries the typed Boolean/structural/semantic/sequence predicate nodes that are already lowerable where supported.
- `polylogue/storage/repository/archive/queries.py` and storage search providers lower supported predicates into SQLite, FTS, vector, recursive traversal, and EXISTS legs.
- `polylogue/cli/query_group.py`, `polylogue/cli/archive_query.py`, `polylogue/cli/query.py`, MCP, and API paths wrap the shared grammar/AST/lowerer rather than owning separate query semantics.
- `polylogue/cli/shell_completion_values.py` consumes `EXPRESSION_FIELD_REGISTRY` and structural query-unit metadata for completion/query-builder candidates.

Current executable behavior includes:

- compact field/text clauses such as `repo:polylogue tag:active "json envelope"`;
- in-field alternatives such as `origin:(codex-session|claude-code-session)`;
- negated field/text clauses such as `-tag:stale` and `-"bad phrase"`;
- count comparisons for `messages` and `words`;
- explicit Boolean predicates such as `sessions where (repo:polylogue OR origin:chatgpt-export) AND NOT tag:stale`;
[...]

PR #2012, opened 2026-06-17
feat(query): execute Boolean FTS predicates (#2006)

Sinity:
## Summary
Adds executable lexical predicates to the #2006 Boolean DSL. `~term` and `~"quoted phrase"` now parse into a typed `QueryTextPredicate` and lower into a correlated FTS5 filter against `messages_fts` joined through `blocks`.

## Problem
After #2011, the DSL could execute session fields, structural `exists(...)`, and action `seq(...)`, but lexical text inside Boolean predicates was still not representable. That meant users could not write queries such as `sessions where origin:chatgpt-export AND ~"timeout failure"` without dropping back to the older flat ranked FTS path.

## Solution
Introduced `QueryTextPredicate`, extended the Lark Boolean grammar with `~text` / `~"phrase"`, and wired the archive Boolean lowerer to `messages_fts MATCH ?`. Because `messages_fts` is contentless and its unindexed columns read back as NULL, the lowerer joins `blocks` by rowid for session correlation, matching the existing search path's storage reality. This PR gives Boolean FTS filter semantics; the existing flat query-term path remains the ranked lexical search path.

## Verification
- `nix develop --command devtools test tests/unit/cli/test_query_expression.py` -> `190 passed in 24.52s`
- `nix develop --command devtools verify --quick` -> `exit_code: 0`
- pre-push `devtools verify --quick` on pushed branch -> `exit_code: 0`

Ref #2006

### DIFF
--- a/polylogue/archive/query/expression.py
+++ b/polylogue/archive/query/expression.py
@@ -75,6 +75,7 @@ from polylogue.archive.query.predicate import (
     QueryNotPredicate,
     QueryPredicate,
     QuerySequencePredicate,
+    QueryTextPredicate,
 )
 from polylogue.archive.query.spec import (
     QUERY_ACTION_TYPES,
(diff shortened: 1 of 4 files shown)


Sinity:
@codex review

PR #2013, opened 2026-06-17
feat(query): execute semantic DSL predicates (#2006)

Sinity:
## Summary

Adds semantic/vector predicates to the canonical Lark query DSL. `semantic:"..."` and the explicit alias `near:text:"..."` now parse into a typed `QuerySemanticPredicate`, compile into the existing `SessionQuerySpec.similar_text` vector seed, and execute through the existing semantic archive path with any remaining Boolean predicate applied as a structured filter.

## Problem

The #2006 DSL had executable SQL, structural, sequence, and FTS predicates, but semantic search still lived only on the older `near:` / `similar_text` path. A semantic predicate cannot honestly lower to a storage `WHERE` clause: vector search produces ranked message candidates first, then archive filters resolve those candidates to sessions.

## Solution

The grammar now accepts `semantic:<text>` and `near:text:<text>` at session scope. The AST bridge lifts a single positive semantic predicate out of a top-level conjunction into `similar_text`, leaving the residual predicate tree as normal Boolean filtering. Unsupported ranked Boolean forms (`OR`, `NOT`, nested structural semantic predicates, or multiple semantic seeds) fail with typed `ExpressionCompileError` messages instead of broadening to lexical search or pretending vector ranking is ordinary set logic.

## Verification

- `nix develop --command devtools test tests/unit/cli/test_query_expression.py` → `196 passed in 27.05s`
- `nix develop --command devtools verify --quick` → `exit_code: 0`

Ref #2006

### DIFF
--- a/polylogue/archive/query/expression.py
+++ b/polylogue/archive/query/expression.py
@@ -74,6 +74,7 @@ from polylogue.archive.query.predicate import (
     QueryFieldPredicate,
     QueryNotPredicate,
     QueryPredicate,
+    QuerySemanticPredicate,
     QuerySequencePredicate,
     QueryTextPredicate,
 )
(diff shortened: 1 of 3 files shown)


Sinity:
@codex review

PR #2014, opened 2026-06-17
feat(query): execute lineage DSL predicates (#2006)

Sinity:
## Summary

Adds a concrete lineage/topology predicate to the canonical query DSL. `lineage:id:<session-id>` now parses into `QueryLineagePredicate` and executes against the existing archive topology materialization, returning sessions that share the seed session's logical root.

## Problem

#2006 requires traversal coverage, not only scalar session filters. Polylogue already materializes logical lineage in `sessions.root_session_id`; the query DSL could not use that evidence directly.

## Solution

The parser registry now exposes `lineage` as a session-scope predicate. The storage Boolean lowerer resolves the seed root with `COALESCE(root_session_id, session_id)` and compares candidate sessions against that same root. This keeps the implementation on the existing topology/read-model contract and avoids inventing a separate traversal engine.

## Verification

- `nix develop --command devtools test tests/unit/cli/test_query_expression.py` → `200 passed in 27.89s`
- `nix develop --command devtools verify --quick` → `exit_code: 0`

Ref #2006

### DIFF
--- a/polylogue/archive/query/expression.py
+++ b/polylogue/archive/query/expression.py
@@ -72,6 +72,7 @@ from polylogue.archive.query.predicate import (
     QueryBoolPredicate,
     QueryExistsPredicate,
     QueryFieldPredicate,
+    QueryLineagePredicate,
     QueryNotPredicate,
     QueryPredicate,
     QuerySemanticPredicate,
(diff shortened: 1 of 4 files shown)


Sinity:
@codex review

Issue #1844, opened 2026-06-12
Build completion on shared query grammar


Sinity:
## Purpose
Build shell completion, fuzzy selection, and query-builder metadata on top of the same query grammar, AST, action contracts, and read-view metadata used by CLI, MCP, daemon, Python, and web routes.

Completion is not a separate language. It is a read-only projection of the parser/lowerer/action/view registries, plus bounded archive-backed value providers.

## Why this exists
The query language is now powerful enough that mistyped fields, stale examples, and unsupported future syntax are expensive. Completion should make valid fields, structural units, operators, values, read views, formats, and `then` actions discoverable without adding a natural-language layer or a second parser.

The product outcome is fewer invalid commands, fewer hidden surface mismatches, and a shared metadata feed that future MCP/web query builders can consume without reimplementing query semantics.

## Current implementation reality
Polylogue already has useful completion plumbing:

- `polylogue completions` emits bash/zsh/fish completion setup.
- `polylogue/cli/shell_completion_values.py` contains static and archive-backed completion providers.
- `polylogue/archive/query/fields.py` defines `QueryFieldDescriptor`, `CompletionSource`, and `query_completion_sources()` for spec-field metadata.
- `polylogue/archive/query/expression.py` owns the Lark-backed query grammar, `EXPRESSION_FIELD_REGISTRY`, and structural query-unit metadata for `exists message(...)`, `exists action(...)`, and `exists block(...)`.
- Public action contracts live in `polylogue.operations.action_contracts`; the CLI re-exports them for compatibility. `CliActionContract.completion_context` and `action_completion_contexts()` expose action-contract completion contexts.
- `SessionViewProfile` and the read-view profile APIs from #1997 expose read-view metadata to CLI, Python, and MCP. `read --view` choices and completion come from `READ_VIEW_PROFILES`; `read --format` choices and completion come from `SessionViewProfile.formats` (#2034).
- `polylogue/cli/select.py` owns the read-only query-backed selector with fzf/UI fallback; `select` is now registered as a query verb and action contract (#2036).
- `polylogue.archive.query.completions` owns the shared `QueryCompletionCandidate` payload and registry-backed candidate providers. `polylogue query-completions`, `Polylogue.query_completions(...)`, MCP `query_completions`, and `GET /api/query-completions` all expose the same structured query/action completion metadata (#2037/#2038/#2039), including registry provenance, danger/stale fields, replacement fields, full descriptions, and readable count/date field discoverability.
[...]

PR #2016, opened 2026-06-17
feat(cli): complete query fields from grammar registry (#1844)

Sinity:
## Summary

Adds the first grammar-backed completion layer for #1844: root query shell completion now suggests DSL field names from `EXPRESSION_FIELD_REGISTRY`, and the shell adapters for bash, zsh, and fish are covered by tests.

## Problem

Completion was still centered on option value providers. The query language field names themselves were not projected from the shared grammar registry, so shell completion could drift from #2006's parser metadata or fail to make query-first field syntax discoverable.

## Solution

Added a structured `QueryCompletionCandidate` model in `shell_completion_values`, generated field candidates directly from `EXPRESSION_FIELD_REGISTRY`, and wired the query-first root group's `shell_complete` hook to append those candidates lazily. The hook preserves the lightweight query-group import boundary by importing completion metadata only during completion.

## Verification

- `nix develop --command devtools test tests/unit/cli/test_command_aux_runtime.py tests/unit/cli/test_completion_matrix.py` -> `54 passed`

Ref #1844

### DIFF
--- a/polylogue/cli/query_group.py
+++ b/polylogue/cli/query_group.py
@@ -188,6 +188,15 @@ def _detect_subcommand_and_verb(group: click.Group, args: list[str]) -> tuple[st
 class QueryFirstGroupBase(click.Group):
     """Custom Click group that routes to query mode by default."""
 
+    def shell_complete(self, ctx: click.Context, incomplete: str) -> list[click.shell_completion.CompletionItem]:
+        items = list(super().shell_complete(ctx, incomplete))
+        from polylogue.cli.shell_completion_values import complete_query_expression_fields
+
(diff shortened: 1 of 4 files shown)


Sinity:
@codex review

Sinity:
Addressed Codex review: structured candidates still carry `kind="query-field"`, but `to_click_item()` now emits Click-supported `type="plain"` so installed bash/zsh/fish scripts do not discard the completion. Added a regression on the adapter output.

Verification:
- `nix develop --command devtools test tests/unit/cli/test_command_aux_runtime.py tests/unit/cli/test_completion_matrix.py` -> `54 passed`

PR #2017, opened 2026-06-17
feat(cli): complete query field values from descriptors (#1844)

Sinity:
## Summary

Adds value completion for query DSL field prefixes. `origin:cla<TAB>` now completes supported origin values through the real shell adapters, and archive-backed fields such as `repo:` route through the same existing completion providers used by root options.

## Problem

The first #1844 completion layer exposed field names from `EXPRESSION_FIELD_REGISTRY`, but it stopped once a field prefix was typed. Completion still had to know, separately, that `origin:` should use origin values or `repo:` should use archive repo values.

## Solution

Mapped `EXPRESSION_FIELD_REGISTRY[field].spec_field` onto `QUERY_FIELD_DESCRIPTORS` and reused the descriptor `completion_source` to dispatch to existing completion providers. The shell adapter preserves the DSL field prefix around returned values, so the inserted candidate remains a valid query expression token.

## Verification

- `nix develop --command devtools test tests/unit/cli/test_command_aux_runtime.py tests/unit/cli/test_completion_matrix.py` -> `58 passed`

Ref #1844

### DIFF
--- a/polylogue/cli/shell_completion_values.py
+++ b/polylogue/cli/shell_completion_values.py
@@ -18,7 +18,7 @@ from click.shell_completion import CompletionItem
 
 from polylogue.archive.message.types import MessageType
 from polylogue.archive.query.expression import EXPRESSION_FIELD_REGISTRY
-from polylogue.archive.query.fields import CompletionSource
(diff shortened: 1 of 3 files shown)


Sinity:
@codex review

PR #2018, opened 2026-06-17
feat(query): suggest close field names in DSL errors (#1844)

Sinity:
## Summary

Adds typo suggestions to query DSL unknown-field errors. A mistyped field such as `origon:chatgpt-export` now reports `did you mean: origin` while preserving the existing recognized-field list and typed `ExpressionCompileError.field` value.

## Problem

#1844 requires useful unknown-field suggestions. The DSL parser already failed loudly, but it forced users and agents to scan the whole recognized-field list even for obvious typos.

## Solution

Added a shared unknown-field error formatter backed by `EXPRESSION_FIELD_REGISTRY` and `difflib.get_close_matches`. Both the flat compatibility accumulator and Boolean predicate validation path use the same formatter, so suggestions are consistent across current query syntaxes.

## Verification

- `nix develop --command devtools test tests/unit/cli/test_query_expression.py` -> `202 passed`

Ref #1844

### DIFF
--- a/polylogue/archive/query/expression.py
+++ b/polylogue/archive/query/expression.py
@@ -63,6 +63,7 @@ from __future__ import annotations
 import json
 import re
 from dataclasses import dataclass, field, replace
+from difflib import get_close_matches
 from typing import Any, Literal, cast
 
 from lark import Lark, Token, Transformer, v_args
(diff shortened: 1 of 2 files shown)


Sinity:
@codex review

PR #2020, opened 2026-06-17
refactor(query): collapse DSL parsing into one Lark grammar (#2006)

Sinity:
## Summary
Collapse the query DSL parser implementation from two grammar strings into one Lark grammar with explicit `flat_query` and `boolean_query` entry points, and remove the private `_lex()` compatibility projection. Current behavior remains covered, but tests now inspect flat clauses through the canonical AST entry point.

## Problem
#2006 defines one canonical query grammar and AST/lowering path. The code already used Lark, but `expression.py` still carried `_QUERY_GRAMMAR` for flat clauses and `_BOOLEAN_QUERY_GRAMMAR` for Boolean predicates. Tests also imported `_lex()`, preserving a flat-parser artifact as if it were a real floor grammar. That split made the old flat path look like retained architecture instead of compatibility behavior absorbed by the Lark DSL.

## Solution
`polylogue/archive/query/expression.py` now has a single `_QUERY_GRAMMAR` and a single `_QUERY_PARSER` configured with two starts. The flat and Boolean transformers still own their respective AST projections, but both parse from the same grammar definition. `_lex()` is deleted; tests use `parse_expression_ast(...).clauses` directly.

## Verification
- `nix develop --command devtools test tests/unit/cli/test_query_expression.py tests/unit/cli/test_click_app.py -k 'query_expression or query_explain or BooleanQueryExpression or Lexer or ExplainExpression'` -> `205 passed, 60 deselected`
- pre-push `devtools verify --quick` -> `exit_code 0`

Ref #2006

### DIFF
--- a/polylogue/archive/query/expression.py
+++ b/polylogue/archive/query/expression.py
@@ -348,31 +348,16 @@ class QueryExpressionExplanation:
 # ---------------------------------------------------------------------------
 
 _QUERY_GRAMMAR = r"""
-    start: clause*
+    flat_query: flat_clause*
 
-    ?clause: COUNT_CLAUSE       -> count_clause
+    ?flat_clause: COUNT_CLAUSE       -> count_clause
         | FIELD_CLAUSE          -> field_clause
         | NEG_QUOTED_TEXT       -> neg_quoted_text
         | QUOTED_TEXT           -> quoted_text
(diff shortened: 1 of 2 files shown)


Sinity:
@codex review

Sinity:
@codex review latest commit 6b678bd6 as well

PR #2034, opened 2026-06-17
feat(cli): complete read formats from view profiles (#1844)

Sinity:
## Summary
Makes `read --format` choices and shell completions come from `SessionViewProfile.formats`, matching the existing profile-backed `read --view` behavior.

## Problem
#1844 says completion metadata should come from the parser/action/view registries rather than parallel lists. `read --view` already used `READ_VIEW_PROFILES`, but `read --format` still used an ad hoc static tuple, so a view-profile format rename/add/delete could drift from the CLI option and shell completions.

## Solution
Derive the read format choice set from the union of `READ_VIEW_PROFILES[*].formats`, add a read-format completer that narrows candidates to the selected `--view` when available, and test both direct Click metadata and bash/zsh/fish completion behavior. The CLI reference was regenerated for the changed option surface.

## Verification
- `nix develop --command devtools test tests/unit/cli/test_query_verbs_runtime.py -k 'read_format or read_view'` → 7 passed, 16 deselected
- `nix develop --command devtools test tests/unit/cli/test_completion_matrix.py -k 'read_format or static_completers'` → 24 passed, 64 deselected
- `nix develop --command devtools verify --quick` → exit_code 0

Ref #1844

## Summary by CodeRabbit

* **New Features**
  - Shell completion for `read --format` now intelligently filters available formats based on the selected `--view`.

* **Documentation**
  - Updated CLI reference and help text to display all supported output formats for the `read` command, including plaintext and text options.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -282,7 +282,7 @@ Options:
                                   [default: summary]
   --to [terminal|stdout|browser|clipboard|file]
                                   Output destination.  [default: terminal]
-  -f, --format [text|markdown|json|ndjson|yaml|html|obsidian|org|csv]
+  -f, --format [csv|html|json|markdown|ndjson|obsidian|org|plaintext|text|yaml]
                                   Output format (where applicable).
   --out PATH                      File path for --to file.
   --all                           Apply to all matched sessions (bulk export).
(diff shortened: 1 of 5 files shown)


Sinity:
Addressed the showcase-baseline finding in commit 5ce405ff: refreshed tests/baselines/showcase/help-read.txt and verified with `devtools lab-scenario verify-baselines` plus `devtools verify --quick`.

PR #2078, opened 2026-06-18
test(storage): cover assertion overlay parity (#1883)

Sinity:
## Summary

Adds public-surface assertion parity coverage for #1883 and fixes the verification harness gap that surfaced while running the default baseline. The storage tests now prove blackboard writes through `Polylogue.post_blackboard_note()` preserve assertion metadata, and they prove ordinary session tags/session metadata remain the intentional table-backed user metadata exception rather than lifecycle assertions.

## Problem

#1883 had helper-level assertion write-through coverage and readiness audit coverage, but the durable public-facade proof was incomplete for two important cases: blackboard assertion metadata and the explicit tags/metadata exception. While validating the branch, `devtools verify` selected a very large pytest-testmon set, ran for 45 minutes, then timed out at 82%. Because pytest-json-report never flushed, verify reported `report_path: null` and lost node-level failure evidence despite many visible failures. That made the broad gate hard to act on.

## Solution

- Extend `tests/unit/storage/test_user_state_contracts.py` with public-facade assertions for blackboard author/evidence/staleness/context-policy metadata.
- Add explicit table-backed coverage for `Polylogue.add_tag()` and `Polylogue.set_metadata()`, matching #1883's current design: simple tag/metadata rows are user metadata, not assertion lifecycle claims.
- Add `devtools.pytest_progress_plugin`, loaded by every `devtools verify` pytest lane, which writes per-test call outcomes incrementally to `.cache/verify/current-pytest-events.jsonl`.
- Persist captured pytest stdout/stderr to `.cache/verify/current-pytest-output.log` and expose both paths in verify metadata so timeout/stall terminations leave usable evidence.

## Verification

- `nix develop --command devtools test tests/unit/storage/test_user_state_contracts.py` -> `9 passed`.
- `nix develop --command devtools test tests/unit/devtools/test_verify.py tests/unit/storage/test_user_state_contracts.py` -> `52 passed`.
- `nix develop --command devtools verify --quick` -> `exit_code: 0` at `8c0a2e94`.
- Attempted default `nix develop --command devtools verify` before the harness fix: static gates passed, pytest-testmon timed out after 2700s at 82% with `report_path: null`. This PR fixes that evidence-loss path; I did not rerun another 45-minute default baseline after the harness patch.

Ref #1883
Ref #2006

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for assertion metadata persistence and user state validation.
  * Added verification for test execution artifact tracking and metadata propagation.

* **Chores**
  * Enhanced test execution monitoring with improved artifact generation for diagnostics.

### DIFF
--- a/devtools/pytest_progress_plugin.py
+++ b/devtools/pytest_progress_plugin.py
@@ -1,0 +1,39 @@
+"""Incremental pytest event ledger for ``devtools verify``.
+
+The pytest-json-report and JUnit artifacts are written at session end. When a
+long verify run is killed for timeout or output stall, those reports may never
+flush. This plugin writes one JSON object per completed test call so the
+operator still has node-level failure evidence after an interrupted run.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import Any
+
+_EVENTS_ENV = "POLYLOGUE_PYTEST_EVENTS_PATH"
+
+
+def pytest_runtest_logreport(report: Any) -> None:
+    """Append one test-call event when configured by ``devtools verify``."""
+    if getattr(report, "when", None) != "call":
+        return
+    raw_path = os.environ.get(_EVENTS_ENV)
+    if not raw_path:
+        return
+    payload = {
+        "event": "test_report",
+        "updated_at": datetime.now(UTC).isoformat(),
+        "nodeid": str(getattr(report, "nodeid", "")),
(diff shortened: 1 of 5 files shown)


Sinity:
Addressed both review findings in 011b8280: the pytest event plugin now records failed setup/teardown reports as well as call reports, and event-ledger writes are best-effort on OSError so telemetry cannot fail pytest. Added direct plugin tests for both behaviors. Pre-push quick gate passed at 011b8280.

Issue #1838, opened 2026-06-12
Render work packets as on-demand evidence bundles


Sinity:
## Purpose
Render work packets as bounded, on-demand evidence bundles for coding-agent work. A WorkPacket is not a task tracker and not a new durable truth store. It is `Bundle(kind=work_packet)` assembled over existing evidence: sessions/runs, issues, PRs, branches, tests, reviews, assertions, caveats, and raw refs.

## Why this exists
Fast agent work often spans several sessions, subagents, branches, reviews, CI checks, and local verification runs. The next agent usually gets either a noisy transcript or a final assistant claim. Neither is enough to continue safely.

A WorkPacket should answer the handoff question: what happened, what remains, and what evidence supports that? It should make existing evidence inspectable without replacing GitHub, Beads, the raw archive, or the issue set.

## Current implementation reality
The first storage-free DTO and recovery-read exposure have landed under #1880:

- `RecoveryWorkPacket` and `RecoveryWorkPacketEntry` live in `polylogue.insights.transforms`.
- `RecoveryDigest.work_packet()` builds a packet from recovery events, subagent reports, RunState, tool summaries, and decisions.
- `render_work_packet(packet)` renders Markdown, and entries require typed `EvidenceRef` support.
- `read --view recovery --report work-packet` renders the packet through the facade-backed recovery report path.
- `read --view recovery --report work-packet --format json` exposes the DTO inside the normal success envelope.
- `Polylogue.recovery_report(session_id, "work-packet")` renders Markdown and `Polylogue.recovery_work_packet(session_id)` returns the DTO for Python consumers.
- The recovery read-view profile advertises work-packet output as part of the successor-agent recovery surface.

That means this issue must not start with a new model package, storage table, or standalone renderer. The remaining value is making the packet richer and more truth-preserving as a Bundle: target refs, missing-evidence markers, review/check/read/addressing distinctions, assertion/caveat links, and broader fixtures over real coding-agent work.

## Primitive mapping
Governed by the #1807 agent-evidence primitive contract:

- task/PR/session/file references use ObjectRef/EvidenceRef (#1845);
- decisions, caveats, lessons, and run state are Assertions (#1883);
- session/run/review/check events come from Run/ObservedEvent when available (#1882/#1840), or from deterministic recovery transforms until storage is justified;
- recovery digests feed packet rendering (#1880);
- the packet itself is a Bundle, not a new database table by default.

## V0 output must answer
A v0 work packet should answer:

[...]

PR #2082, opened 2026-06-18
feat(recovery): mark work-packet evidence support (#1838)

Sinity:
## Summary
Add explicit support markers to recovery work-packet rows and enrich facade-built packets with session-targeted assertion claims from `user.db`.

## Problem
Work packets already carried evidence refs, but API consumers and Markdown readers could not tell whether a row was raw evidence, authored RunState, inferred decision, caveat, or missing evidence. Sparse packets represented missing evidence only as renderer prose (`none extracted`). Separately, #1883 had assertion lifecycle claim reads, but recovery work packets did not consume assertion-backed decisions/caveats/blockers/lessons/transform candidates that targeted the recovered session.

## Solution
`RecoveryWorkPacketEntry` now has a typed `support` marker. The recovery packet builder classifies raw events/tools/subagents, RunState assertions, blocker caveats, inferred decisions, and explicit `evidence_gaps` rows for missing handoff evidence. The facade enriches `recovery_work_packet()` with target-session assertion claims from `user.db`, keeps the transform layer storage-free, and makes `recovery_report(..., "work-packet")` render the same enriched DTO used by the API. Markdown now shows support markers and assertion-claim rows while preserving evidence refs on every packet entry.

## Verification
- `nix develop --command devtools test tests/unit/insights/test_transforms.py -k 'work_packet or continue_report'` -> 3 passed
- `nix develop --command devtools test tests/unit/api/test_facade_contracts.py -k recovery_report_renders_seeded_session_presets` -> 1 passed
- `nix develop --command devtools verify --quick` -> exit_code 0

Closes part of #1838 and advances #1883 recovery/work-packet consumer wiring.

### DIFF
--- a/polylogue/api/archive.py
+++ b/polylogue/api/archive.py
@@ -64,7 +64,12 @@ if TYPE_CHECKING:
     from polylogue.insights.export_bundles import InsightExportBundleRequest, InsightExportBundleResult
     from polylogue.insights.readiness import InsightReadinessQuery, InsightReadinessReport
     from polylogue.insights.resume import ResumeBrief, ResumeCandidate
-    from polylogue.insights.transforms import RecoveryDigest, RecoveryReportPreset, RecoveryWorkPacket
+    from polylogue.insights.transforms import (
+        RecoveryDigest,
+        RecoveryReportPreset,
+        RecoveryWorkPacket,
+        WorkPacketSupport,
+    )
     from polylogue.operations import ArchiveStats
     from polylogue.protocols import ProgressCallback, SessionQueryRuntimeStore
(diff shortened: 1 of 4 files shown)


Sinity:
Addressed the Codex review findings in `9f86f70d`:

- Updated the stale run-state Markdown assertions to expect the new `[assertion]` support markers.
- Updated the stale evidence footer assertion to the new support-marker footer.
- Ran the full transform test file, not only the narrowed work-packet selection: `devtools test tests/unit/insights/test_transforms.py` -> 14 passed.
- Re-ran the facade recovery packet test and `devtools verify --quick` -> clean.

PR #2098, opened 2026-06-18
feat(assertions): export user-tier assertions (#1883)

Sinity:
## Summary

Adds a durable assertion export surface for `user.db`, moves split-archive learning-correction storage onto assertion rows, and fixes the verification harness issues exposed while proving the broad seed run.

## Problem

#1883 still had a backup/export/reset gap: assertions are irreplaceable user-tier state, but there was no direct export surface that included inactive/candidate/deleted rows. The split archive had also removed the old correction table while the async feedback storage path could still try to use it. During broad verification, the seed run exposed unrelated harness fragility: progress writes could race in parallel tests, xdist basetemp cleanup could interrupt report flush, and the memory-budget probe could report parent RSS larger than total tree RSS for very short commands.

## Solution

- Add `ops maintenance assertion-export` with JSONL/JSON output, `--out`, `--kind`, `--status`, and non-negative `--limit` filters.
- Add `list_assertions_for_export(...)` and `assertion_envelope_to_payload(...)` over the typed assertion envelope so export output is deterministic and includes all statuses by default.
- Store archive-root learning corrections in `user_tier.assertions` when split `user.db` is present, and tombstone correction assertions on delete/clear.
- Fix assertion-backed blackboard note decoding for rows whose `key` is absent but `target_ref` carries the note id.
- Update route/status snapshots for `list_assertion_claims` and current CLI command inventory.
- Harden verify/test harness behavior: unique pytest progress temp files, no xdist basetemp deletion during session-finish teardown, and process-tree RSS never below parent RSS.

## Verification

- `devtools test` focused assertion/export/user-state/harness node set: `29 passed`.
- `devtools test` broad-run regression nodes: `7 passed`.
- `ruff check` and `mypy --strict` over changed files: passed.
- `devtools verify --quick`: `exit_code: 0`.
- `devtools verify --seed-testmon --skip-slow`: pytest seed completed with `11280 passed, 1 xfailed`; the enclosing first seed command failed only because `tests/unit/test_pytest_temp_policy.py` needed formatting, then formatting and quick gate passed.
- `devtools verify`: `exit_code: 0`, affected pytest `7 passed`.

Ref #1883

## Summary by CodeRabbit

* **New Features**
  * Added `maintenance assertion-export` CLI command to export durable assertions from the archive in JSON or JSONL formats with optional filtering by kind and status
  * Feedback corrections now use assertions-based persistence for improved durability and consistency
  * Archive status reporting includes assertion claims routing information

* **Documentation**
  * Updated scenario projection catalog with new exercise projections

### DIFF
--- a/devtools/query_memory_budget.py
+++ b/devtools/query_memory_budget.py
@@ -80,6 +80,7 @@ def run_memory_budget(command: list[str], *, max_rss_mb: int, poll_interval_s: f
 
     peak_parent_rss_kb = max(peak_parent_rss_kb, _read_vm_rss_kb(proc.pid))
     peak_rss_kb = max(peak_rss_kb, _read_process_tree_rss_kb(proc.pid))
+    peak_rss_kb = max(peak_rss_kb, peak_parent_rss_kb)
     exit_code = int(proc.returncode or 0)
     peak_parent_rss_mb = round(peak_parent_rss_kb / 1024, 1)
     peak_rss_mb = round(peak_rss_kb / 1024, 1)
--- a/devtools/verify.py
+++ b/devtools/verify.py
@@ -233,10 +233,11 @@ def _pytest_metadata_from_report(report: dict[str, Any], *, report_path: Path) -
     summary = report.get("summary")
     metadata: dict[str, Any] = {"report_path": str(report_path)}
     if isinstance(summary, dict):
-        # `total` is collected count; sum the executed outcomes for the
-        # number we previously scraped from the terminal.
+        # Prefer pytest-json-report's explicit total when present. Older or
+        # reduced reports still get a stable executed count from outcome keys.
         outcome_keys = ("passed", "failed", "error", "skipped", "xfailed", "xpassed")
-        executed = sum(int(summary.get(k, 0) or 0) for k in outcome_keys)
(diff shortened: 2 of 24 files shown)


Sinity:
Review follow-up at dd48264a:\n\n- Fixed Codex P2: async correction listing now preserves valid flat assertion values as payloads, with regression coverage in tests/unit/insights/test_feedback.py.\n- Tightened the SQL interpolation audit per CodeRabbit: removed broad trusted names and moved assertion/session query interpolation to line-specific _AUDITED_SITES entries with rationale.\n- I did not apply the SessionBuilder fixture suggestion literally for user-tier assertion tests: these tests exercise the split user.db assertion helpers directly, and SessionBuilder is session/archive fixture infrastructure rather than the owner of user_tier.assertions. The tests still use isolated tmp/workspace roots and direct user-tier initialization where that is the unit under test.\n\nVerification after the fixes:\n- devtools test tests/unit/storage/test_no_string_interpolated_sql.py tests/unit/insights/test_feedback.py -k 'no_unaudited_string_interpolated_sql or audited_sites_are_real_violations or FeedbackStorage' -> 8 passed.\n- devtools verify --quick -> exit_code 0.

Issue #2178, opened 2026-06-19
feat(import): explain provider detection, parsing, and produced archive rows


Sinity:
## Metadata

Parent: #1807  
Related cleanup umbrella: #2177  
Cross-pollination source: Sinex admission/explainability doctrine from `/tmp/ideas.md` and Sinex source-identity map.  
Owning surfaces: import pipeline, source parsing, raw/session evidence, CLI, Python API, MCP, daemon HTTP, docs/generated schemas.  
Status: partially implemented. PR #2186 added the first CLI ImportExplain vertical for local paths (`polylogue import PATH --explain --format json|ndjson`) and the shared payload models. Remaining scope is prior-explanation retrieval by raw/source ref plus any Python API, MCP, and daemon/web surfaces that should consume the same payload.

## Purpose

Add a first-class import explanation surface that answers:

> Why did this raw artifact import this way, and what did it produce?

Polylogue already detects providers by payload shape, lowers raw payloads into provider/session parse modes, stores raw/session/message/block/action data, and reports some failures through status/debt surfaces. What is missing is a durable, structured explanation object that connects one raw artifact or import attempt to the detector, parser, parser mode, produced rows, skipped rows, caveats, and raw evidence refs.

This is the Polylogue analogue of Sinex admission explainability, but it should stay import-shaped. Do not import Sinex's event/admission model wholesale.

## Closure audit — 2026-06-20

This issue was closed with a real landed baseline, but a broad audit found it is not fully complete against its own acceptance criteria. Keep the landed CLI/API implementation; do not discard it. Remaining scope is:

- MCP must expose `ImportExplainPayload` or the issue must explicitly decide that path-based import explanation is intentionally not safe/useful over MCP.
- Daemon/web shell must expose a local/shell-supported import-explain route or explicitly defer with route metadata that names this issue.
- A raw-ref or source-path read path must retrieve an explanation from already-observed archive evidence when `source.db` / `index.db` rows exist. Re-running detector logic over a path is not enough for this criterion.
- The current generated CLI output schema is good, but OpenAPI/route docs should appear only for exposed daemon routes.
- Privacy coverage must prove raw bytes and absolute local paths are omitted/redacted consistently across every exposed surface, not just the CLI/API preflight path.

Treat the existing `polylogue import PATH --explain`, `ImportExplainPayload`, and `Polylogue.explain_import(...)` as the foundation. The remaining work is to make the surface broad and archive-backed enough to justify closure.

## Landed baseline

[...]

PR #2186, opened 2026-06-19
feat(import): explain parser decisions for local imports (#2178)

Sinity:
## Summary

Adds a first ImportExplain vertical: `polylogue import PATH --explain --format json|ndjson` now runs the existing detector/parser stack without scheduling daemon ingest, and returns a bounded structured explanation of artifact classification, provider/origin detection, parser mode, produced-row counts, skips, and caveats.

## Problem

`polylogue import` is truthful about daemon scheduling, but it previously could not answer why a local raw artifact would be detected as a provider/artifact/parser mode or what archive rows parsing would produce. The new ImportExplain/provider-completeness/debt issues need a concrete payload and surface instead of another prose-only status message.

## Solution

- Added shared ImportExplain payload models in `polylogue.surfaces.payloads`.
- Added `polylogue.sources.import_explain.explain_import_path()` over the existing path resolver, artifact taxonomy, provider detector, and parser dispatch.
- Added non-mutating CLI output for JSON and NDJSON; it reads local bytes but does not stage daemon work, write raw blobs, or print raw payload bytes.
- Handles files, directories, supported ZIP entries, skipped sidecars, malformed JSON, bad ZIPs, and parser failures as finite payload rows.
- Published the CLI output schema and regenerated CLI/topology surfaces.

This intentionally does not introduce a parallel import/admission engine.

## Verification

- `nix develop --command devtools test tests/unit/cli/test_import_explain.py` -> 4 passed
- `nix develop --command devtools verify --quick` -> exit_code 0
- `nix develop --command devtools render all --check` -> sync OK
- `git diff --check` -> clean

Ref #2178


Sinity:
Addressed the two Codex findings in ab1cb016b:\n\n- ZIP import explain now routes supported members through open_bounded_zip_entry and rejects oversized/suspicious members before parsing.\n- ZIP member decode/parser skips are propagated into the top-level ImportExplainPayload.\n\nVerification: devtools test tests/unit/cli/test_import_explain.py; devtools verify --quick; showcase-verify is green in CI.

PR #2238, opened 2026-06-20
feat(query): sort terminal aggregate pipelines (#2006)

Sinity:
## Summary

Adds explicit aggregate ordering to SQL-backed terminal query pipelines. Queries such as `messages where text:aggregate | group by role | count | sort by count asc | limit 2` now parse, execute, and report their aggregate sort stage through the shared `QueryUnitAggregateEnvelope` pipeline metadata.

## Problem

#2006 already had executable `group by FIELD | count` aggregate pipelines, but ordering was hidden inside storage as a fixed `ORDER BY count DESC, group_key`. Users and query-builder surfaces could not request ascending counts or key ordering, and the stage list could not explain the aggregate order that shaped a page.

## Solution

`QueryUnitSort` now covers row-time sorting and aggregate `count`/`key` sorting. The parser accepts aggregate sort stages only after `count` and before `limit`/`offset`, while row `sort by time` remains pre-aggregate only. `ArchiveStore.query_unit_counts()` takes a closed sort selector and direction and builds the SQL order clause from trusted literals. The aggregate envelope builder passes the parsed sort through, and the search docs describe the shipped syntax.

The SQL interpolation audit line numbers were updated because the storage edit shifted existing audited f-string execute sites; the aggregate rationale now also names the closed aggregate order fragment.

## Verification

- `devtools test tests/unit/cli/test_query_expression.py -k "pipeline_group_by_count or aggregate_sort or terminal_aggregate"` -> `6 passed, 280 deselected`
- `devtools test tests/unit/storage/test_no_string_interpolated_sql.py` -> `2 passed`
- `devtools verify --quick` -> `exit_code: 0`
- `devtools verify --seed-testmon --skip-slow` -> `11697 passed`, peak tree RSS `3587.2 MiB`
- `devtools verify` -> `7 passed`, peak tree RSS `584.8 MiB`

Closes no issue; advances #2006.

## Summary by CodeRabbit

* **New Features**
  * Enhanced sorting for aggregated results: now supports sorting by count or key (ascending/descending) in addition to timestamp-based sorting.

* **Documentation**
  * Updated query and search reference documentation with extended sorting and aggregation capabilities.

### DIFF
--- a/docs/search.md
+++ b/docs/search.md
@@ -56,6 +56,9 @@ predicate          ::= predicate "OR" predicate
 
 structural-unit    ::= "message" | "action" | "block" | "assertion"
 pipeline-stage     ::= "sort" "by" "time" ["asc" | "desc"]
+                     | "group" "by" field
+                     | "count"
+                     | "sort" "by" ("count" | "key") ["asc" | "desc"]
                      | "limit" integer
                      | "offset" integer
 ```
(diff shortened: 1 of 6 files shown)


Sinity:
Addressed the aggregate-sort review finding in 8aed038: row-time sorting is rejected before aggregate stages, aggregate count ordering defaults back to descending unless an explicit aggregate sort is present, and the reviewed pipeline shape is covered by a parser regression test.

Issue #2248, opened 2026-06-20
chore(dev): make daemon, web shell, and browser extension loops branch-local


Sinity:
## Problem

Polylogue development still leans too much on the deployed user service and on manual browser state when the work touches the daemon web shell, browser-capture receiver, or browser extension. That makes debugging loops slower and less trustworthy than they should be:

- a local code branch can be tested by unit/integration tests, but it is not obvious how to run a branch-local `polylogued` with branch-local databases and ports without colliding with the systemwide deployed `polylogued.service`;
- daemon/web-shell development should be debuggable from the devshell without installing the branch into the live Nix profile or disturbing the production archive;
- browser-capture development needs the branch-local receiver and unpacked extension to be easy to point at from whatever browser tooling the local agent/operator already has; the repo should not pretend to own or verify the agent browser itself;
- the browser extension has release/package wiring and docs, but it has not had enough end-to-end development-loop verification against ChatGPT/Claude.ai DOMs and the local receiver;
- agents need a repeatable way to inspect the web UI, extension popup/background worker, receiver logs, and captured payloads while staying on a feature branch.

The target is a self-contained local development loop for daemon, web shell, browser-capture receiver, and extension work. It should not require deploying Polylogue systemwide before debugging the branch.

## Scope

Build and document a branch-local development workflow that covers:

1. **Deployed daemon separation**
   - Decide and document the default development posture for the systemwide `polylogued.service`.
   - Preferred starting assumption: the systemwide service is stopped while running branch-local daemon/web/browser-capture debugging, unless the workflow can prove port/archive isolation strongly enough to keep both running.
   - Add a cheap status/preflight command or documented checklist that reports:
     - whether the systemwide `polylogued.service` is running;
     - which process owns the daemon API port (`8766`) and browser-capture receiver port (`8765`);
     - which archive root each active process is using, when discoverable.

2. **Branch-local daemon databases**
   - Provide a devshell-local archive root for branch runs, e.g. `.local/dev-archive/` or `.cache/dev-archive/`, without touching the user's production archive.
   - Make it easy to initialize/reset that archive and run a branch-local daemon:
     - `polylogued run` or a devtools wrapper with explicit `POLYLOGUE_ARCHIVE_ROOT`;
     - branch-local `source.db`, `index.db`, `user.db`, `ops.db`, and `embeddings.db`;
     - independent API/browser-capture ports if the production service is alive.
   - The workflow must make the active archive root and ports visible in logs/status.

[...]

PR #2266, opened 2026-06-21
chore(dev): add CLI capture events to dev-loop timeline (#2248)

Sinity:
## Summary

Adds CLI capture events to the same `dev-loop.events.jsonl` stream used by branch-local daemon launch events. `devtools workspace dev-loop --capture-cli` now records command start and finish metadata alongside its stdout/stderr/transcript/env/summary artifacts.

## Problem

#2248 is about making one branch-local run diagnosable from local artifacts. The daemon launcher now writes lifecycle events, and CLI capture writes good per-command artifacts, but the CLI captures were not represented in the shared run timeline. That left a gap when trying to answer which CLI commands were run against the branch-local archive during the same debugging session.

## Solution

- Emit `cli_capture_requested` before running the captured command.
- Emit `cli_capture_finished` after artifacts are written.
- Include command, timeout, exit code, timeout flag, duration, and artifact paths in the event payload.
- Extend the existing CLI capture test to assert the event stream.
- Document that daemon and CLI capture events share one run timeline.

## Verification

- `devtools test tests/unit/devtools/test_dev_loop.py -q` -> 11 passed
- `devtools render all --check` -> sync OK
- `devtools verify --quick` -> exit_code 0

Ref #2248

### DIFF
--- a/devtools/dev_loop.py
+++ b/devtools/dev_loop.py
@@ -329,6 +329,7 @@ def run_cli_capture(
         raise ValueError("preflight payload is missing artifact paths")
     terminal_dir = Path(str(artifacts["terminal_dir"]))
     terminal_dir.mkdir(parents=True, exist_ok=True)
+    event_path = Path(str(artifacts.get("dev_events", Path(str(preflight["run_log_dir"])) / "dev-loop.events.jsonl")))
 
     env = os.environ.copy()
     suggested_env = preflight.get("suggested_env")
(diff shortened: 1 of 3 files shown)


Sinity:
Addressed the Codex inline finding: the CLI capture test now passes an isolated tmp_path-backed --log-dir instead of sharing the repo-root dev-loop cache, so appended event rows from other runs cannot pollute the assertion. Re-ran devtools test tests/unit/devtools/test_dev_loop.py -q, devtools render all --check, and devtools verify --quick.

PR #2272, opened 2026-06-21
fix(query): bind runtime unit fields to descriptors (#2006)

Sinity:
## Summary

Runtime-transform terminal query units now expose and validate only the session-scoped fields their summary projection can actually evaluate. The parser validates terminal-unit fields through `QueryUnitDescriptor` metadata instead of maintaining a parallel branch of hard-coded unit field sets. Bound field predicates also serialize their validated `field_ref` identity into machine-readable AST/explain payloads.

## Problem

#2006's descriptor work had landed, but parser validation still duplicated unit field truth in `expression.py`. That left runtime-transform units (`runs`, `observed-events`, `context-snapshots`) advertising SQL-backed session predicates that their in-memory recovery/run projection cannot evaluate. The old behavior was narrower than broad FTS fallback, but still misleading: some unsupported inline session predicates parsed and then produced empty results. Also, validated field identity existed only as an internal Python attribute, so explain/query-builder consumers could not see the descriptor-backed field binding execution uses.

## Solution

- Split terminal unit metadata between SQL-backed session-scoped fields and runtime-summary-backed session-scoped fields.
- Added `query_unit_field_names()` so parser validation uses the same descriptor field list as completions/docs/schema surfaces.
- Kept summary-backed counters/date fields available for runtime units while rejecting SQL-only fields such as `session.action`, `session.tool`, `session.path`, and `session.has`.
- Retained SQL-backed `session.path` inline scoping for messages/actions/blocks/assertions.
- Included `field_ref` in `QueryFieldPredicate.to_payload()` when parser binding has resolved the field scope/unit.
- Updated query docs to describe the runtime-transform boundary explicitly.
- Updated parser/metadata tests to cover descriptor-backed field refs, explain payloads, SQL `session.path`, and typed runtime rejection.

## Verification

- `devtools test tests/unit/archive/test_query_metadata.py tests/unit/cli/test_query_expression.py -q` → `338 passed in 88.58s (0:01:28)`
- `devtools verify --quick` → exit 0
- Push hook reran `devtools verify --quick` at `1fc3e5c8cfc3cf85a9fce14f9f87e9bc99b4997b` → exit 0

Ref #2006

### DIFF
--- a/docs/search.md
+++ b/docs/search.md
@@ -234,13 +234,16 @@ one child row matching the nested predicate.
 | `assertion` | `author`, `author_kind`, `author_ref`, `body`, `context`, `evidence`, `key`, `kind`, `scope`, `scope_ref`, `status`, `target`, `target_ref`, `text`, `value`, `visibility` |
 
 Structural units also accept `session.<field>` predicates for the owning
-session fields, such as `session.repo`, `session.origin`,
-`session.tag`, `session.title`, `session.date`, `session.since`, and
-`session.until`. Count and date session fields accept compact comparison
-prefixes such as `session.messages:>=2`, `session.words:<=500`, and
-`session.date:>=2026-01-02`. This lets a unit query carry its session scope
-inline instead of splitting selection between the query string and parallel
-parameters.
+session fields that their lowerer can evaluate. SQL-backed units can use the
+full session filter surface, including action/tool/path/feature predicates.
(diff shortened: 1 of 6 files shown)


Sinity:
Correction to my previous note: addressed the Codex review finding about SQL-backed `session.path`. The full SQL session-scoped descriptor set now includes `session.path`, runtime-transform descriptors still reject it, and `test_sql_terminal_unit_accepts_session_path_scope` covers the SQL path.

PR #2286, opened 2026-06-21
fix(cli): explain query terminal actions (#2006)

Sinity:
## Summary

Adds a side-effect-free explain bridge for query-action verbs. `polylogue --explain find QUERY then select|read|continue|delete|mark|analyze ... --format json` now returns the normal DSL parse/lowering payload plus terminal-action metadata instead of falling through into the action implementation.

## Problem

#2006 calls out the larger target shape where `find QUERY then ACTION` is represented as `Source(sessions) -> Filter(...) -> Terminal(action)`. Root query explain already described compact/Boolean expressions and terminal query-unit pipelines, but query verbs were opaque. In practice, `find repo:polylogue --explain then read --view messages` tried to resolve a session and failed, so explain could not inspect the action boundary at all.

## Solution

- Preserve parsed query terms in `QueryFirstGroupBase` even when dispatching to a query verb.
- Promote the query explain helper so query verbs can attach a `terminal_action` payload.
- Short-circuit `select`, `read`, `continue`, `delete`, `mark`, and `analyze` when `--explain` is present, before any selector, renderer, mutation, or delete path runs.
- Add focused CLI tests over the action floor, asserting the terminal action appears in both top-level explain output and the lowering plan.

This deliberately does not claim that terminal action execution has fully moved into a pipeline AST. It creates the first executable bridge in explain/diagnostics without changing normal action execution.

## Verification

- `devtools test tests/unit/cli/test_click_app.py -k explain -q` -> 12 passed.
- `devtools test tests/unit/cli/test_click_app.py tests/unit/cli/test_query_exec_laws.py -k 'query_action or unit_source or read_verb_messages_requires_id or assertion_unit_source or run_unit_source or context_snapshot_unit_source' -q` -> 13 passed.
- `devtools verify --quick` -> exit_code 0, run `20260621T160042Z-quick-523073-1350e9f2`.

Ref #2006

### DIFF
--- a/polylogue/cli/query.py
+++ b/polylogue/cli/query.py
@@ -76,7 +76,11 @@ def _create_query_vector_provider(config: Config, *, db_path: Path | None = None
         return None
 
 
-def _explain_query_request(request: RootModeRequest) -> None:
+def explain_query_request(
+    request: RootModeRequest,
+    *,
+    terminal_action: dict[str, object] | None = None,
+) -> None:
     if not request.query_terms:
         raise click.UsageError("--explain requires query terms. Use `polylogue find QUERY --explain`.")
     from polylogue.cli.query_explain import explain_query_expression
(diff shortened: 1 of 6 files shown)


Sinity:
Addressed the Codex inline finding about read explain format reporting in `fbd03f8de316410f5583b7f3208586fdbb58422c`: explain now shares the effective read-format helper with execution, so local `then read --format` wins and root `--format` only carries into `--view recovery`. Added tests cover default messages view, recovery root-format inheritance, and local read-format override.

Issue #2308, opened 2026-06-22
fix(ops): prove the deployed archive is coherent after rebuild and repair


Sinity:
## Outcome

The deployed Polylogue archive, daemon, CLI, browser-capture intake, and derived read models agree on one current state after the large rebuild/repair campaign. A restart does not regress authority, convergence, or queryability, and the operator can prove that with one bounded receipt set.

This is the public **live closure gate**, not a container for every historical deployment bug.

## Current reality

Many original implementation gaps are already fixed: version probes, deployment smoke, browser-capture spool discovery/materialization, receiver lifecycle states, route availability, package closure, source-row-missing recovery, bounded health probing, replay/repair machinery, and substantial scale hardening.

The remaining question is empirical and whole-system:

> Does the current deployed archive converge to a trustworthy fixed point, survive restart, and report the same truth through every operator surface?

Historical June measurements and one-off repairs remain evidence, but they are not current acceptance criteria unless reproduced on current code.

## Required receipt set

Capture one coherent run containing:

- expected and deployed commit/package identity;
- source/index/user/ops/embedding schema identity;
- raw-authority reconciliation result and unresolved conflicts/debt;
- latest per-origin acquisition and index convergence;
- browser-capture spool → raw → indexed identity/provenance;
- bounded query/read smoke against the live archive;
- daemon restart and post-restart convergence;
- stale/interrupted attempt handling;
- resource/time summary for startup, replay, derived rebuild, FTS finalization, and steady state;
- explicit caveats for any deferred or intentionally unresolved debt.

## Acceptance criteria

- [ ] The canonical raw-authority reconciler reaches fixed point or emits a finite named set of unresolved conflicts requiring operator judgment.
- [ ] Re-running the reconciliation is idempotent and does not replace accepted authority with a weaker/stale candidate.
- [ ] Source, index, and derived materialization debt converge or remain explicitly bounded and explained.
- [ ] A new browser capture is traced from receiver receipt through raw evidence to a nonzero-message indexed session with provenance identifying provider-native versus fallback capture where applicable.
- [ ] `polylogue`, `polylogued`, daemon HTTP status/health, and deployment smoke agree on deployed version, archive root, schema state, browser-capture state, and current debt.
- [ ] Stale running attempts are interrupted/recovered after process death; no phantom running work survives restart.
- [ ] Restart followed by the same receipt set remains green without manual cursor edits, forged raw rows, or ad hoc SQL repair.
[...]

PR #2321, opened 2026-06-23
fix: prevent silent raw materialization gaps

Sinity:
## Summary

This PR turns the acquired-but-not-materialized raw-session failure mode into an explicit ingest invariant. JSONL stream detection now handles drive-cache filenames such as `*.jsonl.txt.json`, session artifacts that parse to zero sessions are quarantined instead of stamped clean, reparse removes stale same-raw sessions in the same batch transaction, and raw-materialization debt recognizes duplicate raw aliases whose source filenames encode an already-materialized native session id.

## Problem

Live archive evidence showed raw acquisition could be marked clean while user-visible sessions/messages were absent or wrong. The concrete production case was AI Studio Drive cache files that were actually Claude Code JSONL streams behind a trailing `.json` suffix; they had materialized as zero-message `aistudio-drive:*` sessions instead of real Claude Code sessions. Treating this as manual repair would leave the same class of bug available to recur.

## Solution

- Route embedded `.jsonl.` / `.ndjson.` filenames through the stream parser in both the ingest worker and shared JSON decoder.
- Reject session parser outputs that contain zero sessions, so raw rows cannot become parse-success/no-session artifacts silently.
- Delete stale sessions for the same `raw_id` before writing replacement parse results, inside the ingest batch transaction.
- Suppress raw-materialization debt for parsed duplicate aliases whose source filename maps to an existing native session for the same origin, including split suffixes such as `_1`.
- Count detection-warning-only raw-state updates as real updates.

Live production replay with this source tree:

- Replayed 15 affected AI Studio JSONL-cache raw rows.
- Wrote 15 changed sessions and 6,418 messages with 0 parse failures.
- Replayed one missing `ac49922f...` raw row and restored 782 messages.
- Source-tree raw-materialization debt after replay: `0` rows.
- FTS coverage after replay: `messages_fts_docsize=3,987,982`, indexable blocks `3,987,982`.

## Verification

Focused tests:

```text
devtools test \
  tests/unit/operations/test_archive_debt.py::test_archive_debt_raw_materialization_ignores_source_path_native_aliases \
  tests/unit/operations/test_archive_debt.py::test_archive_debt_raw_materialization_ignores_native_id_matches \
  tests/unit/pipeline/test_ingest_batch.py::test_process_ingest_batch_sync_replaces_stale_sessions_for_same_raw_id \
  tests/unit/pipeline/test_ingest_batch.py::test_persist_batch_raw_state_updates_uses_one_typed_update_per_raw \
  tests/unit/pipeline/test_resilience.py::test_jsonl_stream_shape_beats_drive_cache_json_suffix \
  tests/unit/pipeline/test_resilience.py::test_ingest_worker_quarantines_session_artifact_with_no_sessions
# 6 passed
```

Static/generated quick gate:

```text
devtools verify --quick
# ruff format/check, mypy, render all, topology, layering, closure-matrix,
# schema roundtrip, manifests, ci-workflows, doc-commands,
# test-infra-currency, test-clock-hygiene all ok
```

Push hook reran `devtools verify --quick` at `33c998c57` and passed.

Default `devtools verify` was not run because this fresh worktree has no pytest-testmon seed and the command exits with guidance to run `devtools verify --seed-testmon` first.

## Follow-up

Targeted replay performance is not acceptable: a 2.2 MiB / 782-message single-row replay spent ~459s in commit/FTS work. That is now live-measured evidence for the dogfooding performance lane.

Ref #2308

## Summary by CodeRabbit

* **Improvements**
  * Session deduplication now automatically removes outdated entries when processing updated batches
  * JSONL file format detection improved to recognize additional naming patterns beyond standard extensions
  * Raw-materialization debt tracking now correctly handles native-alias source path mappings

* **Bug Fixes**
  * Added error handling for parse operations that produce no materializable sessions

### DIFF
--- a/polylogue/operations/archive_debt.py
+++ b/polylogue/operations/archive_debt.py
@@ -247,7 +247,12 @@ def _raw_materialization_rows(archive_root: Path) -> list[ArchiveDebtRowPayload]
                 """
             )
         )
-        missing_rows = [row for row in candidate_rows if not _raw_materialized_by_source_path_native(conn, row)]
+        missing_rows = [
+            row
+            for row in candidate_rows
+            if _raw_materialization_category(row, archive_root) != "parsed-without-session"
+            or not _raw_materialized_by_source_path_native(conn, row)
+        ]
     except sqlite3.Error:
         return []
     finally:
--- a/polylogue/operations/archive_debt.py
+++ b/polylogue/operations/archive_debt.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import re
 import sqlite3
 from collections.abc import Iterable, Mapping
 from datetime import UTC, datetime
@@ -215,7 +216,7 @@ def _raw_materialization_rows(archive_root: Path) -> list[ArchiveDebtRowPayload]
     except sqlite3.Error:
         return []
     try:
-        missing_rows = list(
+        candidate_rows = list(
             conn.execute(
                 """
                 SELECT
@@ -246,6 +247,7 @@ def _raw_materialization_rows(archive_root: Path) -> list[ArchiveDebtRowPayload]
                 """
             )
         )
+        missing_rows = [row for row in candidate_rows if not _raw_materialized_by_source_path_native(conn, row)]
     except sqlite3.Error:
         return []
     finally:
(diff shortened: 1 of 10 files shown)


Sinity:
Addressed the two Codex inline findings in commit 75d0c8959:\n\n- Alias suppression is now limited to clean parsed-without-session duplicate rows; parse failures, missing blobs, and pending parses remain visible debt. Added test: tests/unit/operations/test_archive_debt.py::test_archive_debt_source_path_aliases_do_not_hide_parse_failures.\n- Stale same-raw session deletion now manually applies the schema's sessions FK actions while bulk ingest has FK enforcement suspended. Added child-row coverage and PRAGMA foreign_key_check assertion in tests/unit/pipeline/test_ingest_batch.py::test_process_ingest_batch_sync_replaces_stale_sessions_for_same_raw_id.\n\nVerification after the fixes:\n- devtools test tests/unit/operations/test_archive_debt.py::test_archive_debt_source_path_aliases_do_not_hide_parse_failures tests/unit/pipeline/test_ingest_batch.py::test_process_ingest_batch_sync_replaces_stale_sessions_for_same_raw_id\n- devtools verify --quick

Issue #2383, opened 2026-06-24
feat(insights): mine agent-workflow failure-mode pathologies


Sinity:
## Outcome

Deterministic, rule-based detectors mine recurring failure-mode pathologies from your own
agent runs — surfaced as evidence-linked findings, queryable, with a distribution view. No
LLM-as-judge: every finding is reproducible from the same evidence.

## Shipped shape

`polylogue/insights/pathology.py` — versioned (`PATHOLOGY_DETECTOR_VERSION`) detectors over
the typed run projection the recovery digest already carries:

- `wasted_loop` — repeated `test_failed`/`check_failed` turns (edit→test-fail→edit churn);
- `missed_review` — a review surfaced but never acted on / acknowledged;
- `stale_context` — a resume/continuation boundary that re-inherited context lossily.

Each finding carries drillable `EvidenceRef`s. Outputs are consumed three ways: the
postmortem bundle's pathology fields (#2380); the `get_pathologies` MCP tool + the
`Polylogue.pathology_report(spec)` distribution view; and `Assertion(kind=pathology)`
candidate rows (queryable via the assertion lifecycle, #2006), emitted by
`materialize_pathology_assertions`. Candidates are private, non-injected, and require
explicit promotion.

## Relationship

Parent #1807 (analysis). Reads the recovery-digest run projections; feeds #2380. Deterministic
and versioned — a rebuild over identical evidence produces identical output.

Sinity:
## #2383 complete — all acceptance criteria satisfied

The pathology-mining workstream landed across three PRs and the full
detect → distribute → emit → promote loop is now realized and consumed.

### Acceptance criteria

- ✅ **≥3 detectors emit findings with `EvidenceRef`s.** `polylogue/insights/pathology.py` ships `wasted_loop`, `missed_review`, and `stale_context` detectors (#2448), each producing `PathologyFinding`s with drillable evidence. They emit `Assertion(kind=pathology)` candidate rows via `upsert_pathology_findings_as_assertions` (#2450).
- ✅ **Deterministic and versioned; rebuild → identical output.** `PATHOLOGY_DETECTOR_VERSION = 1`; findings sorted by `(kind, session_id, detail)`; assertion ids are deterministic over `(session, kind, detector_version, detail, evidence)`, so re-emission is idempotent (test: `test_upsert_pathology_findings_emits_queryable_candidates`).
[...]

Issue #2380, opened 2026-06-24
feat(analyze): add distilled agent-session postmortem bundle


Sinity:
## Outcome

`polylogue analyze --postmortem` renders a compact, single-artifact summary over a matched
session scope, for reviewing your own agent work. Every headline metric carries at least one
drillable `EvidenceRef`, and fields with no signal degrade honestly (explicit status +
reason) rather than fabricating a number.

## Shipped shape

`polylogue/insights/postmortem.py` — a pure aggregator (`compile_postmortem_bundle`) over
already-fetched session profiles + recovery digests, with plain/markdown/JSON renderers.
Headline fields: session count, wall-clock span (calendar span + summed per-session wall),
estimated cost with labelled token lanes (input/output/cache-read/cache-write), repos
touched, top-cost session, subagent-branch count, longest-tool-gap (degraded in v0), and the
`wasted_loop` / `failure_mode` pathology fields populated by #2383's detectors.

The owning API method `Polylogue.postmortem_bundle(spec, *, limit)` does the fetching and
constructs the scope; the CLI verb lives in `cli/query_verbs.py` (`analyze --postmortem`),
and the MCP `get_postmortem_bundle` tool exposes it to agents.

## Relationship

Parent #1807 (analysis). Consumes recovery-digest run projections; its pathology fields are
fed by #2383. A corpus-scope sibling exists in #2437.

Sinity:
Implemented in PR #2431 (branch `feature/feat/analyze-postmortem-bundle`).

`analyze --postmortem` lands as a typed `PostmortemBundle` (new `polylogue/insights/postmortem.py`) with a **pure** scope aggregator, a `Polylogue.postmortem_bundle(spec)` API method (bounded cap 200, truncation flagged + logged), and a CLI mode rendering plain / Markdown / JSON from one payload.

Real headline metrics (each with EvidenceRefs): session_count, wallclock_span (real `wall_duration_ms` — note: the per-session wallclock the issue lists as missing already exists on the profile), estimated_cost with separately-labelled input/output/cache-read/cache-write token lanes, repos_touched, top_expensive_session, subagent_branch_count.

Degraded honestly (null + explicit status, never fabricated): failure_mode & wasted_loop (`no_signal`, await #2383 detector — permitted by AC5); longest_tool_gap (`unavailable` — no per-tool-call timestamps in profile/digest in v0; refused to mislabel `latency_percentiles_ms`).

[...]

Issue #2379, opened 2026-06-24
docs(legibility): build the 7-minute first-contact funnel


Sinity:
First-contact onboarding docs for the repo: a layered README (quick skim → deeper), a glossary mapping internal nouns (`docs/glossary.md`), an architecture diagram, and a one-command demo. Goal is that a reader (including future-self) can orient quickly. Parent #1807.

Sinity:
Implemented in PR #2429 (branch `feature/docs/first-contact-funnel`).

Scope delivered (docs/presentation only, plus one thin command):
- README 30s/3min/30min skim funnel before any project-specific noun + a verified one-command demo (`demo seed --with-overlays` → `analyze --facets`, three providers in one archive).
- `docs/glossary.md` (20+ terms, entry layers), linked from README and registered in the docs map.
- Mermaid four-ring diagram in `docs/architecture.md`.
- `devtools render visual-tapes` — thin entrypoint over the existing tested `visual_vhs` engine, so the screencast media is regenerable by one documented command (`--capture`) instead of a committed binary.

All four ACs satisfied (see PR body). Did **not** touch runtime/archive behavior, and intentionally avoided any marketing-site redesign (#2304/#2307 guard). Verification: `verify doc-commands` clean, `render all --check` no drift, `verify --quick` exit 0, 32 targeted devtools tests pass.

Sinity:
Realized by PR #2429 (merged, commit de82f9232). All 4 ACs satisfied — see the implementation-trail comment above: README skim funnel, one-command demo, Mermaid architecture diagram + glossary linked from README, `verify doc-commands` clean. Train-1 item #1 of #1807.

Issue #1807, opened 2026-06-12
EPIC: Polylogue as local AI-work evidence cockpit


Sinity:
## Purpose

Polylogue is a **local evidence cockpit for the operator's own AI-work history** —
Claude / Codex / ChatGPT sessions and coding-agent runs imported into a queryable
archive. It answers, for *you* (and for a successor agent operating on your behalf):

- What sessions and work traces do I have?
- What did an agent do yesterday — what did it cost, touch, read, write, test, review, merge?
- Which tools, subagents, PRs, checks, reviews, branches, commits, and context boundaries were involved?
- What evidence backs a digest, work packet, assertion, or claim?
- What should resume next; what caveats / blockers are known?

It is **not** a proof catalog, static-site generator, QA dashboard, generic
observability platform, or project-management clone.

This epic coordinates the issue set. It is not one patch — child issues own code.

## Direction — two layers, both for the operator

1. **Understanding** — resume / digest / evidence depth, query, usage & cost,
   assertions. The cockpit over your own archive. This is the realized core.
2. **Substrate (a direction, confirm before a large build)** — Polylogue as the durable
   evidence substrate *underneath* a live work-generating agent (e.g. Hermes): the agent
   reads context/evidence in-loop and its work lands as durable evidence (#2384). Built
   **in-Polylogue** (no second source of truth; intake/writes route through the existing
   ingest/assertion substrates), capturing **recorded evidence**, not a live task store
   Polylogue mutates.

## Child issues (current, by theme)

**Query & contracts**
- **#2006** — full query DSL substrate. The Lark grammar in
  `polylogue/archive/query/expression.py` is the canonical query language; extend it in
  place. Sole owner of query semantics.
- **#1844** — shell completion as a read-only projection of the query/AST/field
  registries. Depends on #2006.
- **#2177** — make contracts own surfaces, delete parallel dispatch.
- **#1883** — user overlays unified as evidence-linked assertions. Substrate is the live
  path; remaining work is consumer wiring + lifecycle tightening.
- **#2246** — targeted archive-schema hygiene for read paths (unsafe joins, JSON checks,
  durable-row rules) — not a rewrite.

**Analysis**
- **#2380** (closed) — distilled postmortem bundle for session review.
- **#2383** (closed) — deterministic failure-mode pathology detection over your runs.
- **#2437** (closed) — corpus cost/scope/wallclock aggregation, surfaced via `read`/`analyze`.

**Substrate axis (confirm before a large build)**
[...]

PR #2431, opened 2026-06-26
feat(analyze): add --postmortem distilled session bundle (#2380)

Sinity:
## Summary

Adds `polylogue analyze --postmortem` (aka "blame") — one compact, shareable
artifact over a matched session scope, available as plain text, Markdown, and
JSON, all rendered from a single typed `PostmortemBundle`. Every real headline
metric carries ≥1 drillable `EvidenceRef`. This is Train 1, item 2 ("the blade")
of the #1807 epic.

## Problem

The source review behind #2380: *"the logs are the ore; do not mail them a
mountain — forge a blade."* #1807 lists these exact cockpit questions (what did
an agent do, what did it cost, what did it touch). The recovery-digest substrate
(`compile_recovery_digest`, the `blame` preset, `get_recovery_report`) was ~70%
of the way there but had no scope-level aggregator and no first-class CLI verb.

Note on the issue premise: it lists per-session **wallclock** as a missing gap.
It is not — `wall_duration_ms` already exists on the session profile
(`archive_models.py`), so the bundle uses real per-session wall time directly.

## Solution

- **`polylogue/insights/postmortem.py`** (new): the typed `PostmortemBundle`
  model set and a **pure** `compile_postmortem_bundle(profiles, digests, *, scope)`
  aggregator (no I/O — unit-testable in isolation), plus `render_postmortem_plain`
  / `render_postmortem_markdown` (both derive from the same bundle object).
- **`Polylogue.postmortem_bundle(spec, *, limit)`** (`api/archive.py`): resolves
  the matched session set via the same summary path `facets` uses, batch-fetches
  profiles, fetches per-session recovery digests for a bounded set (cap 200;
  truncation is flagged on the bundle and logged, never silent), delegates to the
  pure aggregator.
- **CLI** (`cli/query_verbs.py`): `--postmortem` as a mutually-exclusive analyze
  mode; plain (default), `--format json` (via the standard `success(...)`
  envelope), and `--format markdown`.

### Real vs degraded headline fields (honest accounting)

Real, each carrying `EvidenceRef`s: `session_count`, `wallclock_span` (earliest
`first_message_at` → latest `last_message_at`, plus summed real
`wall_duration_ms`), `estimated_cost` with **separately labelled token lanes**
(input / output / cache-read / cache-write — no undifferentiated "tokens", per
AC4) + `total_cost_usd`, `repos_touched`, `top_expensive_session`,
`subagent_branch_count` (counted from `run_projection.runs[].role == "subagent"`).

Degraded honestly (null + explicit `status` + reason, never fabricated):
- `failure_mode`, `wasted_loop` → `status: "no_signal"` — the #2383 pathology
  detector is not built yet (AC5 explicitly permits absent-or-degraded here).
- `longest_tool_gap` → `status: "unavailable"` — neither the session profile nor
  the recovery digest carries per-tool-call timestamps in v0. Using
  `latency_percentiles_ms` (inter-message latency) would have mislabeled it, so
  it degrades rather than fabricates. Tracked as a v0 follow-up.

## Verification

- `devtools verify --quick` → `exit_code 0` (ruff format/check, mypy --strict,
  render all --check).
- `devtools test tests/unit/insights/test_postmortem.py tests/unit/cli/test_plain_cli_snapshots.py`
  → `20 passed, 11 snapshots passed`. The unit tests cover a fully-populated
  multi-session case (cost / repos / subagent / top-expensive all exercised with
  constructed inputs incl. a subagent-bearing recovery digest), a degradation
  case (proves the three degraded fields are null-with-status, never fabricated),
  and a JSON-headline-key contract.
- CLI snapshots (`test_{plain,json}_analyze_postmortem_snapshot`) pin the
  rendered artifact against a deterministic, private-data-free corpus archive
  with insights materialized — `analyzed_session_count=4`, populated
  `wallclock_span`. This is the AC3 stable snapshot-able artifact.
- Live CLI run against a seeded archive returns exit 0 and renders all three
  formats.

## Known limitation / follow-up

`polylogue demo seed` alone does not materialize session-profile insights, so
`demo seed` → `analyze --postmortem` reports `analyzed=0` and an empty bundle
(the command degrades correctly). A populated bundle requires materialized
profiles — present after a daemon-backed ingest (`polylogue import --demo --wait`)
or in the corpus test fixture used by the snapshots. Wiring the #2379 README
first-contact demo to the daemon-backed demo path (so the postmortem renders
populated there) is a small follow-up, not part of this substrate slice.

## Acceptance criteria (#2380)

- [x] `analyze --postmortem` returns every named field; each real headline metric
  carries ≥1 drillable `EvidenceRef`.
- [x] JSON envelope contract-tested; plain + Markdown render from the same payload.
- [x] Runs against a no-private-data archive, producing a stable snapshot-able
  artifact (CLI snapshot test).
- [x] Cost fields label provider/cache/reasoning token lanes separately; wallclock
  is real per-session duration.
- [x] `failure_mode` / `wasted_loop` (and `longest_tool_gap`) are absent-or-degraded
  with explicit status, never fabricated.

Ref #2380. Ref #1807.

## Summary by CodeRabbit

* **New Features**
  * Added a new `analyze --postmortem` option to view a distilled session postmortem in plain text, Markdown, or JSON.
  * Introduced postmortem summaries with scope, timing, cost, repository activity, and diagnostic highlights.
* **Documentation**
  * Updated CLI reference and topology docs to include the new postmortem capability and refreshed project metrics.
* **Bug Fixes**
  * Improved handling of incomplete data by clearly marking unavailable values instead of showing fabricated results.


Sinity:
Addressed all six CodeRabbit findings in 56a422884:
1. **Drillable artifacts (Major):** both the plain-text and Markdown renderers now emit an Evidence section mapping each headline field to its `EvidenceRef`s — the shared bundle is drillable, not just the JSON.
2. **`--format plaintext` (Major):** accepted and routed through the plain renderer (was wrongly rejected).
3. **wallclock_span fallback ref (Major):** falls back to a session ref so a non-zero `summed_wall_duration_ms` stays drillable when boundary timestamps are absent.
4. **Bounded truncation log (Major/Security):** logs a 5-id preview instead of the full dropped-session tail.
5. **JSON contract on the shipped path (Major):** the unit test now asserts against `model_json_document(..., exclude_none=True)` (the CLI's actual serializer) and verifies degraded fields keep status+reason with no fabricated value.
6. **Markdown snapshot (Trivial):** added so the third renderer is pinned.

Verification: `devtools test tests/unit/insights/test_postmortem.py` → 3 passed; `-k postmortem` → 4 passed; `devtools verify --quick` → exit 0.

PR #2433, opened 2026-06-26
feat(export): add fail-closed sanitized shareable export (#2381)

Sinity:
## Summary

Adds `polylogue ... then export --sanitized <DIR>` — a fail-closed export that
produces a sanitized **dataset + redaction manifest + README**, refusing to
publish unless an independent leak-check gate passes. Train-1 item #3 of #1807.

## Problem

The source review makes redaction the **precondition** for any public/portfolio
sharing. The privacy primitives existed (`schemas/privacy.py`,
`privacy_config.py`, `redaction_report.py`, context-pack path redaction) but
there was no user-facing "export a shareable, sanitized bundle" command, so the
operator's own scale case-study could not ship.

## Solution

- `polylogue/export/sanitize.py`: a pure `sanitize_rows` (absolute/`$HOME` paths
  → `<redacted-path>/{basename}`, known secrets + opaque high-entropy tokens →
  `<redacted-secret>`, every decision recorded in a `SchemaReport` manifest that
  stores only the **redacted** value), an **independent** `verify_sanitized_export`
  gate that re-scans the written files, and an atomic temp→publish
  `write_sanitized_bundle` that runs the gate and refuses (cleaning up, raising
  `SanitizedExportError`) on any surviving path/secret.
- API `Polylogue.sanitized_export(spec, …)` + a CLI `export --sanitized` verb
  (`--privacy-level`, `-o`, `--with-postmortem`). The projection excludes raw
  repo/cwd/file paths and message text.
- Fail-closed by default: `--no-redact` is refused unless combined with an
  explicit `--i-accept-unredacted-leak` acknowledgement.

### Security gap caught during review (and fixed)

An adversarial probe found a **fail-closed violation**: a path containing a
space (`/home/me/Secret Project/notes.md`) only partially redacted — the
space-naive path regex stopped at the space, leaking the directory and filename,
and the same-pattern gate did not catch it. Fixed by allowing spaces within path
segments so the whole path reduces to `<redacted-path>/{basename}`
(over-redaction is the safe direction for a sanitizer). Regression tests added:
the directory structure no longer survives, and the gate refuses a raw
space-containing path.

## Verification

- `devtools verify --quick` → exit 0 (ruff, mypy --strict, render all --check).
- `devtools test tests/unit/export/` → 12 passed (pure redaction, gate,
  fail-closed, + space-path regressions).
- CLI-coupled slice (verb cardinality, help/completion snapshots) → 226 passed;
  the new `export` verb's help snapshot regenerated.
- Independent adversarial probe: planted absolute path, `$HOME` path,
  space-containing path, and `sk-…`/`gh*_`/SSH-URL secrets — none survive in
  `dataset.jsonl` or `redaction-manifest.json`; the gate refuses raw leaks.

## Acceptance criteria (#2381)

- [x] Export refuses to emit when redaction is incomplete (fail-closed) with
  typed diagnostics (`SanitizedExportError`, `SanitizedExportVerifyResult`).
- [x] Output includes a redaction manifest (what was removed, by which rule).
- [x] A test proves no private absolute path / known-secret pattern survives
  (incl. the space-path regression).
- [~] Accepts a postmortem Bundle (`--with-postmortem`). **Deferred:** the
  `spans-jsonl` OTel format, to keep the v0 leak surface bounded.

## Honest v0 limitations (follow-ups)

- **Raw message text is not exported** — the highest-leak surface; deferred.
- **Windows `\`-style paths and email addresses are not yet redacted** — the
  gate covers POSIX absolute paths + `$HOME` + known-secret value patterns. The
  basename is intentionally retained (`<redacted-path>/{basename}`), consistent
  with the context-pack redaction primitive.
- `spans-jsonl` export deferred.

Ref #2381. Ref #1807.

## Summary by CodeRabbit

* **New Features**
  * Added a new `export` command to create sanitized bundles from the matched scope.
  * Export output can include an optional postmortem, with controls for redaction, privacy level, overwrite, and result limits.

* **Documentation**
  * Updated CLI help and reference docs to list the new export command.

* **Bug Fixes**
  * Exported bundles now undergo leak checks to help prevent sensitive paths or secrets from appearing in published output.


Sinity:
Addressed all five CodeRabbit findings in 7ceee1e20:
1. **Single-segment roots (Major security):** added a known-root pattern (`/etc`, `/tmp`, `/home`, …) applied in both the redactor and the gate, with a negative lookahead so `/etcetera` and multi-segment paths are untouched.
2. **Raw `~/…` gate gap (Major security):** the gate now scans for tilde paths directly, so a surviving `~/private/key.txt` fails closed.
3. **`--limit 0` bypass (Major security):** non-positive limits are now rejected instead of falling back to the 1000-row default.
4. **Non-atomic overwrite (Major):** the writer moves the existing bundle to a backup, swaps the new one in, then drops the backup, restoring on failure.
5. **Root `--json`/`--format` (Minor):** the export verb now inherits the root format before choosing the renderer.

Adversarial re-probe: `/etc`,`/tmp`,`~/secret/x` redacted; `/etcetera` preserved; gate independently refuses raw `/etc` and `~/private/key.txt`. `devtools test tests/unit/export/` → 14 passed; `devtools verify --quick` → exit 0.

Sinity:
Re: the re-surfaced review thread — all five were addressed in 7ceee1e20 (the atomic-overwrite, root-format, limit, and tilde-gate items are fully resolved; verified by tests + adversarial probe). 

One documented residual on the single-segment item: the known-root pattern covers `/etc`, `/tmp`, `/home`, `/Users`, `/var`, `/opt`, `/mnt`, `/media`, `/srv`, `/usr`, `/private`, `/data`, `/realm`, `/persist`. An arbitrary single-segment path to a *non-standard* root (e.g. `/custommount`) is still not flagged, by deliberate tradeoff — flagging every `/word` would over-redact ordinary prose ("see section /intro") and break legitimate exports. This sits alongside the already-documented v0 limitations (Windows `\` paths, emails). The realistic private-path threat (multi-segment paths, known roots, `$HOME`/`~` paths, known-secret patterns) is fail-closed via the independent gate.

PR #2443, opened 2026-06-26
feat(mcp): expose postmortem bundle and sanitized export as agent tools

Sinity:
## Summary

Adds two MCP tools so agents reach the distilled postmortem bundle (#2380) and
the fail-closed sanitized export (#2381), instead of those products being
CLI+API-only. Ref #2436.

## Problem

Two recently-landed distillation products were unreachable from MCP, while the
only MCP exporters (`export_session` / `export_query_results`) render
**unredacted** content with no privacy gate — a footgun on the agent-facing
surface. Train-2 wants agent-consumable *distilled* bundles; MCP was the surface
left behind.

## Solution

Leaf-adapter tools delegating to the existing substrate (no new distillation
logic in MCP):

- **`get_postmortem_bundle`** (read role) — wraps `Polylogue.postmortem_bundle`,
  returns the typed `PostmortemBundle` over a matched filter scope.
- **`export_sanitized`** (write role) — wraps `Polylogue.sanitized_export`, runs
  the fail-closed leak gate; on gate failure returns a typed
  `sanitized_export_refused` envelope and publishes nothing. Redaction is
  mandatory — the MCP surface exposes **no** `redact`/`acknowledge_unredacted`
  affordance.

The raw unredacted exporters stay `admin`-gated, giving a clean privilege
ladder: read = postmortem < write = sanitized < admin = raw.

Both tools are registered in the enumeration gates (`EXPECTED_TOOL_NAMES`,
`TOOL_CONTRACT`). Docs: `mcp-integration.md` catalog + `architecture.md` MCP
note with the redacted-vs-raw distinction.

## Acceptance criteria (#2436)

- [x] `get_postmortem_bundle` and `export_sanitized` registered, delegating to
  the #2380/#2381 substrate (no reimplementation)
- [x] `export_sanitized` runs the fail-closed gate; a leak returns a typed
  refusal and publishes nothing
- [x] The MCP sanitized path has no redaction-disable affordance (signature test)
- [x] Edge-case tests under `tests/unit/mcp/`; enumeration/contract gates updated
- [x] `architecture.md` + MCP tool catalog reflect the new tools and the
  unredacted-vs-sanitized distinction
- [x] `devtools verify` green on the MCP slice

## Verification

`devtools test tests/unit/mcp/test_distilled_bundle_tools.py` — 5 passed.
`devtools test tests/unit/mcp/test_envelope_contracts.py tests/unit/mcp/test_server_surfaces.py`
[...]

PR #2449, opened 2026-06-26
feat(insights): expose pathology distribution via API + MCP tool (#2383)

Sinity:
## Summary

Surfaces the deterministic agent-workflow pathology detectors (#2448) as a
first-class queryable distribution: a new `Polylogue.pathology_report()` API
method and a `get_pathologies` MCP read tool that return every finding from all
three v1 detectors over a matched scope, with counts by kind and drillable
evidence refs.

## Problem

The v1 pathology detectors landed in #2448 but were only reachable through the
postmortem bundle's `wasted_loop` / `failure_mode` summary fields. There was no
way to pull the full pathology distribution — all three detectors
(wasted_loop / missed_review / stale_context), every finding, and the
per-kind counts — across a matched scope from the API or an agent surface.

## Solution

- `polylogue/api/archive.py`: add `Polylogue.pathology_report(spec, *, limit)`.
  It mirrors `postmortem_bundle`'s scope resolution but fetches only the
  recovery-digest run projections the detectors consume, then runs
  `compile_pathology_report` over them.
- `polylogue/mcp/server_insight_tools.py`: add the `get_pathologies` read tool.
  It forces `limit=None` on the resolved spec via `dataclasses.replace` so the
  matched scope is not silently capped at the `MCPSessionQueryRequest` default
  of 10 (the same footgun CodeRabbit caught in #2443).
- Register `get_pathologies` in `EXPECTED_TOOL_NAMES` (tests/infra/mcp.py) and
  `TOOL_CONTRACT` as `single_object` (tests/unit/mcp/test_envelope_contracts.py).
- Add a delegation/serialization test asserting counts_by_kind, findings, and
  that the delegated spec carries `limit is None`.
- Document the tool in `docs/mcp-integration.md`.

## Verification

- `devtools test tests/unit/mcp/test_server_surfaces.py` → 75 passed
- `devtools test tests/unit/mcp/test_distilled_bundle_tools.py tests/unit/mcp/test_envelope_contracts.py` → 40 passed
- `mypy` clean on all changed files
- `ruff format` / `ruff check` clean

## Follow-up

Remaining #2383 scope: emit `Assertion(kind=pathology)` candidate rows
queryable via the #2006 assertion lifecycle. `assertions.kind` is plain TEXT
(no DB CHECK), so the `AssertionKind` addition is schema-free — this is the
final slice.

Ref #2383

## Summary by CodeRabbit

* **New Features**
  * Added a new read-only tool to surface pathology insights for a session scope.
  * Results are returned as a single structured object and can be limited for larger scopes.

* **Documentation**
  * Updated the tool reference to include the new pathology insights command.

* **Tests**
  * Expanded coverage for tool registration, response formatting, and scope handling.

### DIFF
--- a/docs/mcp-integration.md
+++ b/docs/mcp-integration.md
@@ -54,6 +54,7 @@ Add to `~/.config/claude/claude_desktop_config.json`:
 | `get_session` | Get a single session by ID (supports prefix matching) |
 | `stats` | Archive statistics: totals, provider breakdown, database size |
 | `get_postmortem_bundle` | Distilled postmortem bundle over a matched scope (#2380): top sessions by cost, repos touched, tool/work-kind rollups, failure signals. Read-only. |
+| `get_pathologies` | Deterministic agent-workflow pathology distribution over a matched scope (#2383): wasted-loop, missed-review, and stale-context findings with drillable evidence. Read-only. |
 | `export_sanitized` | Write a fail-closed **redacted** shareable bundle (#2381) to a path; refuses (typed error, nothing published) if any private path or secret survives the leak gate. Write role. |
 
 > **Redacted vs. raw.** `get_postmortem_bundle` and `export_sanitized` are the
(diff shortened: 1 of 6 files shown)


Sinity:
Applied the parity warning in bcd69068c — `pathology_report` now emits the same `logger.warning` with dropped-preview as `postmortem_bundle` when matched scope exceeds the cap. Thanks.

Issue #2384, opened 2026-06-24
feat(substrate): prove a live evidence-to-context loop under an agent


Sinity:
## Outcome

A work-generating agent emits structured evidence into Polylogue while it runs, can query that evidence immediately, and later consumes an evidence-backed context/resume artifact that materially improves continuation of the same work.

That repeated bidirectional use, not the existence of more archive machinery, is the closure condition for moving Polylogue from a downstream evidence archive toward an in-loop substrate.

## Current reality

Polylogue already has substantial archive, query, lineage, assertion, recovery, context-pack, MCP, and evidence-ref machinery. It can continuously ingest and reconstruct work from several providers.

The missing outcome is narrower and more important:

- no production first-class provider-neutral work-event write leg is established;
- Hermes work topology is not yet reconstructed from the strongest available state/runtime evidence;
- hook-event materialization has advanced, but complete query projection and coverage are not proven;
- inbound OTLP remains primarily disposable runtime telemetry rather than durable work evidence;
- context/recovery surfaces exist, but repeated in-loop use that changes real work has not been demonstrated.

Do not interpret this issue as permission to build a new general event architecture before testing a terminal loop.

## Minimum proving loop

Use one real agent/runtime and one real project task:

1. **Emit:** the runtime records a bounded typed set of work events through Polylogue's current source/admission authority.
2. **Query:** events, lineage, tool outcomes, decisions/caveats, and raw support are immediately queryable with truthful coverage.
3. **Compile:** Polylogue produces a compact evidence-backed resume/context artifact.
4. **Consume:** a fresh agent invocation receives that artifact before acting.
5. **Measure:** compare the resulting continuation against a raw-transcript or ordinary-handoff baseline.

The first implementation may be intentionally narrow. Generalization follows observed failures, not the other way around.

## Acceptance criteria

- [ ] A non-synthetic agent run emits provider-neutral typed work evidence during execution through the current canonical source/write authority, with deterministic idempotency and no parallel source of truth.
- [ ] The emitted run is queryable before the original runtime session ends, including evidence refs and explicit unsupported/missing coverage.
- [ ] A compact context/resume artifact is generated from that evidence and consumed by a fresh continuation invocation.
- [ ] The trial records whether the artifact recovered non-obvious state, changed the next action, prevented duplicated work, or introduced misleading context.
- [ ] At least three real resumptions use the loop; one must include a failure, retry, delegation, or context boundary.
[...]

PR #2458, opened 2026-06-27
feat(insights): materialize run projection into queryable SQL tables

Sinity:
## Summary

Materialize the run projection (`Run` / `ObservedEvent` / `ContextSnapshot`) from a
storage-free, recomputed-every-query view into three derived index.db read-model tables
(`session_runs`, `session_observed_events`, `session_context_snapshots`), recomputed by the
session-insight materializer exactly like `session_profiles`. This flips the `run`,
`observed-event`, and `context-snapshot` query units (#2006) from `runtime_transform`
(no `exists`, no sort, full-archive scan per query) to durable SQL lowerers with
`exists`-predicate and sort support. This is the keystone storage layer of the substrate
program (#2384): it gives the agent-work evidence vocabulary a durable, queryable home.

## Problem

`polylogue/insights/run_projection.py` was explicitly *"storage-free … without inventing a
new durable table before there is measured query pressure."* The pressure arrived from two
directions:

- **#2006**: the `run`/`observed-event`/`context-snapshot` query units were stuck at
  `exists_supported=False`, `lowerer_kind="runtime_transform"`. Every `runs where …` query
  scanned the whole archive and ran `compile_recovery_digest(session)` per session at read
  time; `exists run(…)` and `sort` were rejected as unexecutable.
- **#2384 (substrate)**: the substrate write-leg needs a durable place for agent-work
  evidence to land. Projections-over-transcripts have nowhere to write.

## Solution

Run/ObservedEvent/ContextSnapshot become **derived materialized read-models** (rebuildable,
like `session_profiles`), NOT a new source of truth. The durable evidence stays
`session_events` + `messages` + `topology_edges`; the materializer recomputes the read-models
over them.

- **Schema** (`storage/sqlite/archive_tiers/index.py`, `common.py`): three new tables with typed
  columns for SQL filtering (`harness`/`role`/`status`/`kind`/`delivery_state`/`boundary`/…) +
  a `payload_json` for lossless model hydration. New `literal_check()` DDL helper derives enum
  CHECKs from the `run_projection` `Literal` aliases via `typing.get_args`. `INDEX_SCHEMA_VERSION`
  10 → 11.
- **Materializer** (`storage/insights/session/rebuild.py` + `run_projection_rows.py` + `storage.py`):
  `build_session_insight_records` computes `compile_recovery_digest(session, session_links=()).run_projection`
  (identical inputs to the old runtime query → byte-for-byte parity) and emits the three record
  lists; wired into the sync bulk write site, both async write sites, the full-rebuild DELETE
  set, and the materialization stamp loop.
- **Readers** (`storage/repository/insight/run_projection_reads.py`, query store, mappers):
  `query_runs`/`query_observed_events`/`query_context_snapshots` hydrate stored rows back to the
  exact projection models.
- **#2006 lowerers** (`archive/query/metadata.py`, `storage/sqlite/archive_tiers/archive.py`,
  `archive/query/unit_results.py`): descriptors flipped to `sql` + `exists_supported=True`; new
  per-unit field→column predicate clauses + EXISTS clauses + ArchiveStore terminal `query_*`
  methods mirroring the `assertion` unit; runtime-transform executor path removed (no second
  executor). Payload `from_row` reuses the prior row mapping so CLI/MCP/API/daemon payloads are
  unchanged.

### Schema-Touching: re-ingest plan (no in-place upgrade)

Per `docs/internals.md` § Schema Versioning Model, this bumps `INDEX_SCHEMA_VERSION` 10 → 11, so
an existing `index.db` is rejected on open. There is **no in-place upgrade** — it is a
deletes-then-defines edit of the index DDL. Re-acquisition:

```
polylogue ops reset --database && polylogued run
```

- This rebuilds `index.db` **from `source.db`** (NOT a provider re-acquire) — raw evidence is
  untouched.
- The new derived tables, FTS, and all existing insights (profiles/work-events/phases) rebuild
  automatically via the insight materializer.
- `user.db` (tags/marks/annotations/assertions) is unaffected.
- Operator impact: one index rebuild from source; no source loss, no provider re-download.

## Verification

```
nix develop -c devtools verify --quick          # ruff format+lint, mypy --strict, render all --check
nix develop -c devtools test tests/unit/insights/test_run_projection_materialization.py
nix develop -c devtools test tests/unit/storage/test_session_insight_refresh.py
nix develop -c devtools test tests/unit/cli/test_query_support_runtime.py   # #2006 query units
nix develop -c devtools render topology-projection && … render topology-status
```

Results:
- `devtools verify --quick` → **exit 0** (ruff format+lint, mypy --strict across 1571 files, `render all --check`, manifests, ci-workflows, doc-commands, test-infra-currency, test-clock-hygiene all green).
- `test_run_projection_materialization.py` → **1 passed** (stored rows hydrate equal to `compile_recovery_digest(session, session_links=()).run_projection`; typed-filter + re-materialization idempotence).
- `test_session_insight_refresh.py` → **19 passed** (sync + async rebuild write the new tables).
- `test_query_exec_laws.py` → **74 passed** (terminal rows, `exists run/observed-event/context-snapshot(...)`, sort-by-time, SQL limit/offset).
- Broad query-DSL + consumer sweep (`test_query_metadata`, `test_query_expression`, `test_completions_contract`, `test_web_reader`, `test_facade_contracts`, `test_server_surfaces`, `test_pathology`, `test_transforms`, …) → **passing**.

## Follow-ups

- #2384 Phase 1 write-leg (`record_work_event` push → `session_events` → this materialized layer).
- #2006 remainder: terminal/transform pipeline stage, docs unit/field/operator tables.

Ref #2384 Ref #2006

## Summary by CodeRabbit

* **New Features**
  * Added support for querying and listing run, observed-event, and context-snapshot insights.
  * Expanded query filtering, sorting, and pagination for these insight views.
  * Updated query explanations and examples to reflect the new supported query paths.

* **Bug Fixes**
  * Improved consistency between archived data, materialized insight tables, and query results.
  * Existing insight data can now be rebuilt and queried reliably after refreshes.


Sinity:
## Adversarial review + perf/re-ingest validation (independent agents)

**Correctness review** ran a field-by-field diff of the new SQL `_*_field_predicate_clause` lowerers against the removed runtime matchers. Findings and dispositions:

- **`search_text` / `text:` parity (fixed — 5b6e2f6e).** The broadened `search_text` diverged from the runtime `text` haystack: `run` dropped `confidence` and wrongly added `native_session_id`/`native_parent_session_id`; `observed-event` wrongly added `kind`/`delivery_state`/`run_ref`. Now matches the runtime `text` field set exactly (verified against the removed `_run_field_matches`/`_observed_event_field_matches` haystacks); `context-snapshot` already matched.
- **LIKE wildcard semantics (deliberately unchanged).** `_like_clause` treats `%`/`_` as LIKE metacharacters, so `text:50%` wildcards where the runtime did a literal Python substring. This is the established, parameterized (injection-safe) semantic shared by every SQL unit (`message`/`action`/`block`/`assertion`); making the new units consistent with their siblings is preferred over exact parity with the removed path for this edge case. Escaping would be a cross-cutting change to the shared helper — out of scope here. Flagging for the record.
- **Verified sound:** rebuild idempotence (DELETE-then-insert across sync + both async + empty-rebuild branches), `payload_json` round-trip (model_dump→model_validate lossless), EXISTS JOIN correlation, session-scoped field routing.

**Docs (fixed — 0e571fc9).** `docs/search.md` still described these units as runtime-transform / sort-rejecting / non-`exists` — now corrected to the SQL reality (sort/limit/offset + `exists` selectors + full session filter surface; `group by | count` still typed-unsupported).

**Perf + re-ingest validation** (360-session synthetic archive, cross-version timing):
- Materialization adds **+30%** to a full insight rebuild (~1.73 ms/session; `compile_recovery_digest` is 0.59 s of the 0.62 s — the 3 bulk table writes are 0.04 s). Linear floor on the ~24 K production archive ≈ **+42 s** to a full re-ingest. Acceptable: full rebuilds only run on version-bump re-ingest; the daemon's incremental path pays this only for changed sessions; and the work **moves** out of every read into once-per-rebuild (net read-latency win). No gating/incrementalization needed.
- Re-ingest path **confirmed end-to-end**: v10 index.db rejected under v11; `polylogue ops reset --database` preserved `user.db`; re-acquire + rebuild repopulated all three tables (360 runs / 568 observed-events / 360 context-snapshots). Note: `reset --database` re-acquires from the original **source artifacts on disk** (it also drops `source.db`), matching the `docs/internals.md` "re-acquire from source" wording.

Recommend a one-time observation of the real full re-ingest rebuild time after deploy, since `compile_recovery_digest` cost scales with session size (synthetic sessions are small).

Ref #2384 Ref #2006

Sinity:
## CodeRabbit dispositions (commit 5546ccc6)

- ✅ **Cross-session `sort by time`** (archive.py): was ordering by per-session `position` only, interleaving every session's position-0 rows. Now `ORDER BY COALESCE(source_updated_at,''), session_id, position` across all three units. Added a multi-session regression test (`test_terminal_run_sort_by_time_orders_across_sessions`) using native ids whose alphabetical order differs from chronological, so the old behaviour fails it.
- ✅ **`session_ids=[]` global delete** (rebuild.py async): aligned to the sync path — an empty (non-None) target list is now a no-op, not a wipe of the derived tables. Only `session_ids is None` does a full rebuild.
- ✅ **Unsupported `sort` value** (`_order_by`): now raises a typed error for anything other than `position`/`recency` instead of silently falling back to position.
- ↩️ **Add `event_ref` to the observed-event haystack** — declining. `search_text` backs the DSL `text:` predicate, whose contract (#2006) is parity with the removed runtime matcher; that matcher's observed-event `text` covered summary + subject/object/evidence refs only, NOT `event_ref`. Adding it would re-introduce the over-matching just fixed in 5b6e2f6e. The `RunProjectionListQuery` reader's free-text `query` shares this column but is not yet wired to any surface; when it is, ref lookups go through the typed `event_ref`/`run_ref` filters, not free-text.
- ↩️ **`frozen_clock` / `open_connection()` in the test fixture** — minor test-hygiene on `search_workspace`; deferring to keep this PR's test change minimal (the fixture's `datetime.now` seeding predates this PR and the lint passes). Noted for a follow-up.

`devtools verify --quick` green (mypy --strict); 363 query/materialization tests pass incl. the new sort regression.

PR #2571, opened 2026-07-08
fix(browser-capture): require an auto-minted bearer token by default

Sinity:
## Summary

The browser-capture receiver now requires a bearer token by default —
auto-minted into a 0600 file on first use — instead of defaulting to no
auth at all on plain loopback. Closes the remaining gaps in
polylogue-gnie and unblocks polylogue-kwsb.1's deferred token-mandatory
work.

## Problem

`polylogue/browser_capture/server.py`'s `_reject_token()` gate is shared
uniformly by every route (GET `/v1/status`, `/v1/archive-state`,
`/v1/post-commands`, and POST `/v1/browser-captures`,
`/v1/post-commands`) but only enforces a bearer token when one happens
to be configured. With `auth_token=None` (the previous default), any
local process — not just a browser page — could read
spool/archive-lifecycle state, drain queued post commands, or post
forged captures. kwsb.1's Host/Origin admission gate (#2559, merged)
closed the DNS-rebinding read hole but deliberately deferred this
default-auth work, since flipping the default with no pairing path
would 401 every operator who hadn't manually configured a token.

## Solution

- `polylogue/browser_capture/receiver.py`: `load_or_mint_receiver_token()`
  (atomic mkstemp + `fchmod(0o600)` + `os.replace`, a plain
  pairing-secret file rather than the keyring-backed OAuth store in
  `sources/token_store.py`) and `resolve_receiver_auth_token(explicit,
  allow_no_auth, token_path)`: an explicit `--browser-capture-auth-token`
  always wins, otherwise auto-mint/load, with
  `--browser-capture-allow-no-auth` /
  `POLYLOGUE_BROWSER_CAPTURE_ALLOW_NO_AUTH` as the loud, logged opt-out.
- Wired into both `polylogued run` (`daemon/cli.py`) and the standalone
  `browser-capture serve` CLI. `make_server()`'s own default is
  untouched, so existing unit tests that build a server directly with
  `auth_token=None` are unaffected.
- New `polylogue browser-capture token show [--rotate]` CLI prints the
  token to pair with the extension's existing manual token field
  (`browser-extension/src/background.js`, unchanged).
- `browser-extension/src/popup.js`: a 401 now renders as "Receiver
  requires a pairing token. Run `polylogued browser-capture token
  show`..." instead of a generic "Receiver offline" with no next step.
- `polylogue/config.py`: new `browser_capture_allow_no_auth` config key
  (TOML/env/CLI, inventory-registered).
- Docs updated: `docs/daemon.md`, `docs/configuration.md`,
  `docs/security.md`.

**Rejected alternative:** the bead's AC option (a) asked for the
extension to "auto-discover" the token file directly. Browser
extensions can't read arbitrary host files without a native-messaging
host — a much larger, separate capability — so this PR substitutes the
CLI-pairing + popup-UX equivalent above. Documented in the closing bead
notes (polylogue-gnie).

## Verification

- `devtools test tests/unit/browser_capture tests/unit/daemon/test_daemon_cli.py tests/unit/core/test_config_inventory.py tests/unit/core/test_config.py` → 251 passed, including a new end-to-end real-HTTP round trip (`test_receiver_token.py::test_default_resolved_token_gates_get_and_post_and_pairs_end_to_end`) proving the default-resolved token gates both GET and POST and that the same value pairs both.
- `devtools verify --quick` (format/lint/mypy/render/topology/layering/doc-commands/manifests/ci-workflows/test-clock-hygiene) → all green, including on the pre-push hook run above.
- `npx vitest run` in `browser-extension/` → 91 passed (incl. new unauthorized-state popup test); `npx eslint src/ tests/` clean.

## Follow-ups

- `polylogue-1rfj`: stale `polylogue browser-capture ...` doc references (should be `polylogued`) discovered in `browser-extension/README.md` and `docs/design/mk2/**` — pre-existing, not touched here.
- `polylogue-rzve`: `--api-auth-token` help text claims auto-generation that isn't implemented — unrelated daemon-API component, discovered in passing.

Ref polylogue-gnie polylogue-kwsb.1

## Summary by CodeRabbit

* **New Features**
  * Browser-capture now uses a pairing token by default, with a new token display/rotation command and updated startup options.
  * The browser extension popup now shows a clear pairing prompt when access is unauthorized.

* **Bug Fixes**
  * Unauthorized browser-capture access is now handled more clearly instead of appearing offline.
  * Added support for an explicit no-auth mode when needed.

* **Documentation**
  * Updated configuration, daemon, and security docs to reflect the new token-based browser-capture behavior.


Sinity:
Fixed: added `envvar=BROWSER_CAPTURE_ALLOW_NO_AUTH_ENV` to the `--allow-no-auth` option so `POLYLOGUE_BROWSER_CAPTURE_ALLOW_NO_AUTH=1` actually works on the standalone `browser-capture serve` command, matching the help text's claim. New test proves default/env-var/flag resolve identically (c85abec62).

PR #2572, opened 2026-07-08
fix(context): coerce non-user assertion authors to candidate at one chokepoint

Sinity:
## Summary

`upsert_assertion` — the sole assertion write path — now coerces any
non-`user` `author_kind` to `status=CANDIDATE` with a non-injected
context policy, closing a live self-promotion hole (polylogue-37t.15).

## Problem

`blackboard_post` (MCP tool default `author_kind="agent"`) let an agent
write an assertion that landed `status=ACTIVE` with whatever
`context_policy` the caller passed — an agent's claim could self-promote
to authoritative, injectable memory with no operator review. Enforcing
this per-writer (as `upsert_transform_candidate_assertions` /
`upsert_pathology_findings_as_assertions` already did by hand) provably
leaves holes: blackboard was the counterexample nobody had gated.

## Solution

The invariant now lives inside `upsert_assertion`
(`polylogue/storage/sqlite/archive_tiers/user_write.py`) itself — the
only assertion write path (confirmed live 2026-07-06: no async storage
twin to mirror). Any `author_kind != "user"` is coerced to
`status=CANDIDATE` + `context_policy={"inject": False,
"promotion_required": True}`, overriding whatever the caller requested.
If the `assertion_id` already carries a terminal judgment outcome
(accepted/rejected/deferred/superseded/deleted — set only via
`mark_assertion_status` from `judge_assertion_candidate`, never by this
function), that outcome is preserved instead of resurrected.

**Escape hatch:** a new `require_promotion: bool = True` parameter — the
design's own "allowlist argument, not `author_kind` sniffing." The one
caller that opts out is `upsert_session_tag_assertion`:
`AssertionKind.TAG` has no judgment-queue path (not in
`ASSERTION_CLAIM_KINDS`) and `ArchiveStore.add_user_tags`' existing-row
short-circuit would strand an agent-tagged session as a permanently
unreachable candidate with zero promotion path. Filed `polylogue-ldau`
to make that exemption a deliberate, tracked decision rather than a
silent side effect discovered mid-review. `upsert_mark` /
`upsert_suppression` / `upsert_session_metadata_assertion` already
hardcode `author_kind="user"` and are structurally unaffected either
way.

## Test fixture fallout

8 pre-existing test fixtures encoded the vulnerable pre-fix behavior
(`author_kind="agent"`/`"human"` landing active + injectable) across
`test_archive_tiers_assertions.py`, `test_blackboard_facade.py`,
`test_archive_tiers_assertion_write_through.py`,
`test_archive_tiers_archive.py` (tags — covered by the
`require_promotion` carve-out), `test_web_reader.py`,
`test_facade_contracts.py` (×2), `test_query_expression.py` (×2), and
`tests/visual/conftest.py`. Updated each to either assert the new
coerced-candidate outcome or use `author_kind="user"` where the
fixture's actual intent was "simulate an already-promoted/active row"
(matching production: only `_promote_candidate_assertion`, which always
uses `author_kind="user"`, produces a genuinely active row).

## Verification

- `devtools test tests/unit/storage/test_archive_tiers_assertions.py tests/unit/storage/test_blackboard_facade.py tests/unit/storage/test_archive_tiers_assertion_write_through.py tests/unit/storage/test_archive_tiers_archive.py tests/unit/operations/test_archive_debt.py tests/unit/mcp/test_blackboard_tools.py` → 85 passed.
- Targeted `-k assertion` runs on `test_web_reader.py` / `test_facade_contracts.py` / `test_query_expression.py` → 28 passed, plus a full run of `test_server_surfaces.py` + `test_tool_contracts.py` + `test_user_state_contracts.py` → 163 passed (confirming their `author_kind="agent"` fixtures are mocked/DTO-only reads, unaffected by the write-path change).
- `devtools verify --quick` → all green, including the pre-push hook run above.

## Follow-ups

- `polylogue-ldau`: agent-authored session tags have no judgment/review
  path — either build one or make the TAG exemption an explicit,
  documented risk decision.

Ref polylogue-37t.15

## Summary by CodeRabbit

* **Bug Fixes**
  * Non-user authored assertions are now consistently treated as pending review, with existing finalized outcomes preserved.
  * Session tag and related writes now follow the correct visibility/review behavior.
  * Stored assertion metadata now keeps the expected review-needed marker.

* **Documentation**
  * Updated guidance to reflect the new centralized review rule for non-user authored assertion writes.

* **Tests**
  * Refreshed test coverage to match the updated assertion lifecycle behavior.


Sinity:
Fixed: reworded the docs/search.md guarantee to state it as a default with the two named exceptions (terminal-judgment preservation, TAG's require_promotion=False opt-out) rather than an absolute (ed1c4824b).

PR #2575, opened 2026-07-08
fix(storage): stop usage_timeline silently dropping timeless-session usage

Sinity:
## Summary

`list_usage_timeline_insights` no longer unconditionally drops usage/cost
data from sessions with no reliable timestamp (polylogue-rvtu, split
from the polylogue-srjq sort_key_ms COALESCE audit).

## Problem

The base filters — `COALESCE(e.occurred_at_ms, s.sort_key_ms, 0) > 0`
(event scan) and `s.sort_key_ms > 0` (cost scan) — unconditionally
excluded any session with neither a reliable event timestamp nor a
session `sort_key_ms`, not just under a `since`/`until` window. Real
token/cost usage from a timeless session silently vanished from every
monthly bucket in `usage_timeline`/`cost_rollups` forever, understating
actual spend with no signal that data was dropped. This was the most
severe finding in
[`.agent/reports/sort-key-ms-coalesce-audit-2026-07-08.md`](https://github.com/Sinity/polylogue/blob/master/.agent/reports/sort-key-ms-coalesce-audit-2026-07-08.md)
(#2574) — unlike the windowing bugs elsewhere in the audit, this one
isn't gated by `since`/`until` at all.

## Solution

Removed both unconditional exclusion filters. The bucket computation in
each scan now uses a `CASE` expression: a row with a genuine timestamp
buckets normally (`YYYY-MM` via `strftime`), a row with none routes to an
explicit `"unknown"` bucket instead of disappearing. `since`/`until`
windowing (`s.sort_key_ms >= ?`/`<= ?`) is intentionally left unchanged
— that's the separate, less-severe windowing pattern the sibling
`polylogue-z29t`/`s5mm`/`2seq` beads address; this fix is scoped to the
more severe unconditional/unwindowed drop.

## Verification

- New `tests/unit/storage/test_usage_timeline.py`: seeds a session with
  `NULL updated_at_ms`/`created_at_ms` (so the generated `sort_key_ms`
  column is `NULL`) plus a usage event/cost row with `NULL
  occurred_at_ms` — both the event-count and cost-dollar paths now land
  in an `"unknown"` bucket instead of vanishing, plus a sanity check that
  ordinary timestamped sessions still bucket normally (unchanged
  behavior).
- `devtools test tests/unit/storage/test_usage_timeline.py tests/unit/api/test_facade_contracts.py -k usage_timeline tests/unit/cli/test_insights.py tests/unit/mcp/test_envelope_contracts.py tests/unit/mcp/test_tool_discovery.py -k usage` → all passed.
- `devtools verify --quick` → all green, including the pre-push hook run above.

Ref polylogue-rvtu

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed usage timeline insights so sessions without reliable timestamps are no longer dropped.
  * These records now appear in an explicit **unknown** bucket, with event counts and cost totals included.
  * Existing monthly bucketing for timestamped sessions remains unchanged.


Sinity:
Fixed: added a cheap existence check that bypasses `_usage_timeline_event_scan_cutoff_ms`'s pagination optimization whenever a genuinely timeless event exists, falling back to the always-correct full scan instead of risking the unconditional `occurred_at_ms IS NOT NULL` filter dropping "unknown"-bucket data (2eaf14937). Added a regression test forcing the cutoff heuristic to actually engage (3 real cost buckets + limit=3) alongside a timeless event, proving no crash/data loss.

Note: traced through the pagination/sort interaction and since the "unknown" bucket always sorts last (lexically after every `YYYY-MM` string), the specific output-visible failure mode described may not currently be reachable via the public API given today's sort semantics -- but the fix is correct, cheap, and closes the gap defensively regardless (e.g. against a future sort-order or group_by change).

PR #2587, opened 2026-07-08
docs(demos): attachment acquisition census by origin and byte volume

Sinity:
## Summary
Read-only census over the active archive grouped by (origin, acquisition_status), reconciled against `polylogue ops maintenance attachment-acquisition-debt`, committed under `.agent/demos/attachment-acquisition-census/` as the baseline ("before") evidence for the 83u attachment-acquisition program.

## Problem
polylogue-83u.6 asks: after the v13 schema split honestized attachment acquisition state, how much attachment evidence is actually backed by bytes, and where is the recoverable gap concentrated? Nothing answered this per-origin — the global attachment-acquisition-debt command (83u.4) reports totals only, not where the unfetched byte volume actually lives.

## Solution
- `regenerate.sh`: opens `index.db` read-only (`mode=ro`), joins `attachments` to `attachment_refs`/`sessions` (LEFT JOIN throughout so a zero-ref orphaned attachment still counts, keeping totals reconciled), groups by (origin, acquisition_status), and for acquired rows checks blob-store presence directly (same signal as `attachment-acquisition-debt`, broken down by origin instead of globally).
- Cross-checks its totals against the `attachment-acquisition-debt` CLI command's global output and records the match as `reconciliation.totals_match` in `census.json`.
- `README.md` documents the regeneration command and the baseline finding; `ANALYSIS.md` is the human-readable summary table.

## Finding (baseline, 2026-07-08)
7,390 attachments; 967 acquired (0 missing blob files); 6,423 unfetched. Unfetched byte volume is heavily concentrated in `chatgpt-export` (13.4GB declared, 0 acquired) — the largest single re-acquisition opportunity in the archive and the natural first target once polylogue-83u.2/83u.3 ship.

## Verification
- Ran `regenerate.sh` against the live active archive; `census.json`'s `reconciliation.totals_match` is `true` (attachment_count, acquired-on-disk + missing count, and missing-blob-ref count all agree with the `attachment-acquisition-debt` command's totals).
- `devtools render all --check` clean after `devtools workspace demo-shelf` regenerated the shelf indices (CURATED_CATALOG.md, MANIFEST.readable.json, README.md, SUMMARY_INDEX.json).

## AC honesty
This is the "before" half of the before/after pair the bead's AC requires; the "after" re-run happens once polylogue-83u.2/83u.3 land, at which point the delta gets written back into the parent 83u epic as closing evidence — not claimed here.

Ref polylogue-83u.6.

## Summary by CodeRabbit

* **New Features**
  * Added a new “Attachment Acquisition Census” demo entry to the catalog and overview pages.
  * Included new demo outputs showing attachment acquisition totals, unreadable/missing bytes, and reconciliation results.
  * Added regeneration instructions and a ready-to-run script for producing the census report.


Sinity:
Fixed in 3b7d3679e — tracks `acquired_null_blob_hash_count` separately and includes it in the reconciliation formula. Re-verified live: `totals_match` still `True`, the anomaly count is 0 (doesn't currently exist in the archive, but the formula is now correct in general).

PR #2590, opened 2026-07-09
docs(demos): D4 behavioral archaeology -- six DSL queries, rapid fire

Sinity:
## Summary
Real, executed Demo Finding Packet for polylogue-212.4: six DSL queries answering engineering-lead questions about AI coding sessions that no chat UI transcript view can answer, run against the seeded demo corpus and packaged per the 212.7 packet contract.

## Problem
polylogue-212.4 named six illustrative query shapes (SEQ thrash-loop hunt, tool call volume, tool failure rate, semantic probe, time-scoped population, query-piped-into-read) but none had actually been run and captured as a conforming packet.

## Solution
`.agent/demos/d4-behavioral-archaeology/` (registered in `.agent/demos/registry.json`, mode: public) with all 7 required files:
- Q1 `seq(action:shell -> action:shell)` (repeated shell calls): 2/11 sessions match, verified via `then select --json`.
- Q2 tool call volume: `actions where exit_code:>=0 | group by tool | count` — Bash 9, Read 8, Task 1, Write 1, exec_command 1.
- Q3 tool failure rate: `actions where is_error:true | group by tool | count` — Bash 4, exec_command 1.
- Q4 semantic probe: `near:"flaky async test"` — 0 results, honestly attributed to the fixture's sparse embedding coverage (2/43 messages), not claimed as a semantic-search failure.
- Q5 time-scoped population: `since:2y` — 9/11 sessions.
- Q6 piped into read: `find 'origin:codex-session' then read --first --view messages` — resolves to a real captured tool error and the agent's next-step response.
- `--explain` shown once on Q1 to prove the parsed AST matches intent.

## A real defect surfaced while authoring this, not hidden
Bare `polylogue find "sessions where seq(action:shell -> action:shell)"` (no `then` verb) returns the FULL unfiltered session list (total 11) instead of the 2 that actually match — reproducible with an ordinary non-SEQ predicate too (bare `find "sessions where origin:codex-session"` also returns 11, while the equivalent compact form `find "origin:codex-session"` correctly returns 5). Filed as polylogue-70qb with the exact three-command repro rather than worked around or hidden — this is the demo's own thesis in action: a DSL query surfaced a real product defect a chat transcript view never could.

## Verification
- `devtools lab policy demo-packet-registry` passes with both registry entries (the earlier contract-proof stub plus this real demo).
- Every query's raw output is captured verbatim in `run.log`; `report.md`'s findings/counterexamples/limits sections cite it directly.
- `devtools render all --check` clean (demo-shelf indices regenerated via `devtools workspace demo-shelf`).

Ref polylogue-212.4.

## Summary by CodeRabbit

* **New Features**
  * Added two new demo scenarios to the catalog, README, and registry so they appear in the available demo list.
  * Included new demo materials for running, reviewing, and reproducing the “D4 Behavioral Archaeology” experience.

* **Bug Fixes**
  * Documented a query behavior issue where one search form could return broader results than expected.
  * Updated validation notes to reflect successful demo checks and known coverage limits.


Sinity:
Both fixed in 972f9861f — filled the placeholder bead id (polylogue-70qb) and added the denominator query (43 total messages) to evidence.ndjson so the 2/43 ratio is fully cited.

PR #2597, opened 2026-07-09
fix(storage): pair actions view by transcript rank, not tool_id equality

Sinity:
## Summary
The `actions` view (index.py) is the sole read model 8+ downstream consumers (tool usage stats, session profiles, action filters) rely on for the tool_use/tool_result relation. It paired both sides with a plain equality join on tool_id, which fans out N*M when a provider re-emits the same tool_id on distinct messages.

## Problem
Verified live 2026-07-06 (bead xnkf): real Claude Code sessions contain provider-re-emitted messages with identical toolu_ ids on distinct native message ids (not variants — in-file re-emission). One live sample session (a9bb6d50-...) carried 491 real tool_use blocks but the buggy view reported 503 action rows — a 12-row over-count from duplicate-tool_id fan-out.

## Solution
Rewrite the view (`INDEX_SCHEMA_VERSION` 25→26, derived-tier canonical DDL edit): rank tool_use and tool_result blocks separately by transcript order (message position, then block position) within `(session_id, tool_id)`, then pair same-rank rows via a UNION of two branches — the ranked-pair branch for blocks with a real tool_id, and an always-unpaired branch for blocks with a NULL or empty-string tool_id (preserving the pre-fix behavior that a tool_id-less use still surfaces, just without a paired result; a plain equality join never matches NULL, so this isn't a behavior change for the NULL case — only the empty-string cross-product class is newly guarded).

A regression fixture caught this exact split mid-development: my first pass dropped tool_id-less uses from the view entirely, breaking an existing test (`test_actions_view_uses_blocks_without_session_payload_bloat`) before I added the UNION branch.

## Verification
- New regression tests in `tests/unit/storage/test_archive_tiers_ddl.py`: `test_actions_view_pairs_reemitted_tool_id_by_transcript_rank_not_cross_product` (2 uses + 2 results sharing one tool_id → exactly 2 paired rows, not 4) and `test_actions_view_never_cross_pairs_empty_string_tool_id` (two empty-string-tool_id uses plus an empty-string result never cross-pair).
- Live before/after (ad-hoc query against the real archive, no schema rebuild performed): the cited sample session drops from 503 to 491 action rows, exactly matching its raw tool_use block count.
- `devtools test` across `test_archive_tiers_ddl.py`, `test_store_ops.py`, `test_tool_usage.py`, `test_archive_search_contracts.py`, `test_archive_tiers_archive.py`, `test_archive_query.py`: 241 passed.
- `mypy --strict` on changed files: clean. `devtools render all --check`: clean.

## Migration
Derived-tier schema bump (25→26): existing `index.db` files raise on next daemon startup until rebuilt — `polylogue ops reset --index && polylogued run`.

Ref polylogue-xnkf.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how action entries are matched so repeated tool IDs no longer create duplicate pairings.
  * Prevented empty or missing tool IDs from incorrectly matching with unrelated results.

* **Tests**
  * Added coverage for repeated tool ID pairing behavior.
  * Added coverage for empty-string tool ID handling and unpaired actions.


Sinity:
Fixed in 7a3471fdc — added `variant_index` to both ORDER BY clauses as the deterministic tie-breaker, confirmed a new test (`test_actions_view_ranks_variant_messages_deterministically`) that deliberately inserts uses/results in mismatched order genuinely fails without the fix and passes with it.

PR #2603, opened 2026-07-09
feat(lineage): surface a completeness signal on composed reads

Sinity:
## Summary
Lineage composition (`read_archive_session_envelope`, `get_messages`) can silently return an INCOMPLETE transcript in two cases: a chain deeper than `_MAX_LINEAGE_DEPTH` (ancestors beyond the cutoff are dropped), and a dangling branch point (the parent message was hard-deleted, so only the child's own divergent tail is returned starting mid-conversation). Neither case surfaced any signal — a partial transcript was served as if whole.

## Problem
For a system-of-record, serving a truncated transcript without marking it as such is a construct-validity hole: a consumer (reader, MCP get_messages, context-image builder) has no way to distinguish "this is the complete logical transcript" from "this stops mid-conversation because something was dropped."

## Solution
- **Sync**: `ArchiveSessionEnvelope` gains `lineage_complete: bool = True` and `lineage_truncation_reason: str | None = None`. `read_archive_session_envelope` tracks both conditions during its recursive composition and propagates a parent's own incompleteness up (a parent's own reason wins over a locally re-derived one — see the code comment on why checking the parent's completeness FIRST matters: a depth-limited parent's composed messages don't include its own inherited prefix, so the next level's branch-point search would otherwise fail and get mis-attributed as "dangling_branch_point" instead of the real "depth_limit" cause).
- **Async**: new `get_messages_with_lineage_completeness` returns `tuple[list[MessageRecord], LineageCompleteness]` (new small model in `storage/runtime/archive/records.py`). `get_messages` becomes a thin wrapper that discards the signal — single source of truth, no logic duplication, and every existing caller (5+ files) is unaffected since its return contract is unchanged.
- Depth-limit hits are logged (`logger.warning`) in both paths.

## Verification
- New tests in `tests/unit/storage/test_lineage_normalization.py`: dangling-branch-point (both paths), depth-limit for the sync path (a real 65-level chain against its 64 recursion limit) and for the async path (patched `_MAX_LINEAGE_DEPTH` down to 3 for a fast, small chain, since the iterative async path has its own much larger 1024 backstop), plus a sanity check that a normal shallow fork reports `complete=True`.
- Caught and fixed a real masking bug during development: without prioritizing the parent's own completeness signal, a depth-limited ancestor's curtailed composition made the very next level's branch-point search fail too, reporting "dangling_branch_point" instead of the true "depth_limit" root cause — reverted the fix and confirmed the new test fails without it, then restored the fix and confirmed it passes.
- `devtools test` across `test_lineage_normalization.py`, `test_archive_tiers_write.py`, `test_store_ops.py`: 147 passed.
- `mypy --strict` on all changed files: clean. `devtools render all --check`: clean.

Ref polylogue-4ts.6.

## Summary by CodeRabbit

* **New Features**
  * Added lineage completeness details to transcript/message retrieval, so composed histories can now report whether they are complete or truncated.

* **Bug Fixes**
  * Improved transcript assembly to avoid silent partial results when lineage depth is exceeded or a branch point is missing.
  * Preserved the most accurate available transcript while clearly flagging incomplete lineage with a reason.


Sinity:
Addressed in 4acbc7f2d:
- **P2 (carry the flag into read payloads)**: fixed for the two MCP-facing surfaces the AC explicitly names (`MCPMessagesListPayload`/get_messages tool, `MCPArchiveSessionPayload`), with a new test proving both project a truncated envelope's signal correctly. The CLI reader payload (`_session_payload`) and Python API `Session` model (`_archive_session_to_session`) still don't carry it — filed as polylogue-vv2b rather than silently left or scope-crept into this PR.
- **Literal type + shared constants**: done — `LineageTruncationReason` Literal type plus `LINEAGE_TRUNCATION_DEPTH_LIMIT`/`LINEAGE_TRUNCATION_DANGLING_BRANCH_POINT` constants shared between the sync and async composition paths.
- **Test naming**: both renamed to match what they actually exercise.

PR #2630, opened 2026-07-10
fix: word-boundary anchor the work-event text-signal keyword matcher

Sinity:
## Summary

Fixes substring false-positives in the work-event text-signal keyword classifier and inventories the activity-type classifier as permanently heuristic-tier per the bead's second acceptance clause.

## Problem

`polylogue/archive/session/extraction.py`'s `_TEXT_SIGNAL_TABLE` matched keyword patterns against user text via naive `pattern in text` substring checks, silently mislabeling work events on unrelated words that happen to contain a pattern: `fix` inside `prefix`/`suffix`, `test` inside `latest`/`contest`, `plan` inside `explanation`/`airplane`, `data` inside `update`/`validate`/`metadata`, `spec` inside `respect`/`inspect`, `move` inside `remove`, `config` inside `reconfigured`. The noise was invisible in the hardcoded confidence float on `WorkEvent` rows.

## Solution

- `_word_boundary_pattern()` compiles each pattern group into a single `\b(?:...)\b` alternation regex. Word boundaries work correctly for both single words and multi-word phrases (`stack trace`, `should we`) since `\b` only requires a word/non-word transition at the pattern's own start/end, not internally.
- `_text_signal_from_lowered_text` now does `pattern.search(text)` instead of the substring check.
- No change to confidence literals or signal-table ordering — this is the correctness fix only (calibration is a separate bead, 9e5.9).
- Inventoried the activity-type classifier in the rigor contract matrix (`polylogue/insights/rigor.py` + mirrored `docs/insights-rigor-matrix.md`): a `notes` entry on the `session_work_events` contract explicitly states it stays heuristic-tier permanently (no structural ground-truth signal exists for "what category of work is this," unlike `tool_result_is_error`/`exit_code`) plus a per-origin coverage caveat.

## Verification

- New `tests/unit/archive/test_work_event_text_signals.py` — 11 parametrized false-positive cases (previously-colliding words no longer match) + 10 genuine single-word/phrase signal cases (still match correctly), 21 passed.
- `devtools test tests/unit/insights/` — 401 passed (rigor contract/audit coverage unaffected).
- `devtools verify --quick` — exit 0.

Ref polylogue-b0b.1


Sinity:
Follow-up commit corrects an overclaim in the rigor-matrix note this PR adds: it originally said the activity-type classifier "stays heuristic-tier permanently" without qualifying whether it predicts anything real. Checked polylogue-9e5.9's own evidence — it measured a sibling keyword heuristic in the same file (`_terminal_state`'s `_ERROR_MARKERS` fallback) against structural ground truth on 14,377 real runs and found only 50.5% agreement (coin-flip level). This classifier's own accuracy has never been measured.

Reworded the note to say so honestly and filed `polylogue-ve9z` (decision bead) for the actual product-scope question (invest in real classification vs. measure-first vs. delete vs. status-quo). The word-boundary fix itself is unaffected — it corrects the matching mechanism, not a claim about predictive value.

PR #2641, opened 2026-07-10
fix: restore deterministic archive contracts

Sinity:
## Summary

Restore the deterministic archive baseline by repairing the production
contracts exposed by the original 12-node failure cluster and updating stale
expectations where behavior had intentionally changed.

## Problem

The baseline mixed stale fixtures with real defects: browser-capture ownership
depended on arrival order and retained stale provenance/freshness state; direct
writes reprocessed identical content; the SQL audit used source-line identity;
the token-output guard missed ordinary dataflow and response sinks; and blob
repair fixtures implicitly depended on wall-clock age.

A later cold review found that the direct hash skip was not yet evidence-honest:
same-size changed inline attachment bytes were absent from canonical content
identity, missing FTS rows survived identical direct re-ingest, and the token
guard did not classify the daemon's real `_send_json`/`_send_error` helpers as
output sinks.

The first seeded baseline run at `9edb87062` passed 13,235 tests and exposed
five further failures. Three installed-script smoke tests were resolving
ambient wrappers from the canonical checkout instead of the worktree. Two demo
tests revealed an order-dependent storage defect: a later full replacement
deleted archive-generated `capture_gap` evidence, so concurrent parser
completion order could change verification results.

The investigation also proved a separate blob-publication/GC race. That durable
source-tier repair is tracked under `polylogue-8jg9.2` and remains isolated from
this branch because it requires a source-v4 migration and backup gate.

## Solution

- Centralize browser-capture precedence for daemon and direct writers, require
  structural native-payload evidence, and replace accepted ownership and
  freshness state coherently.
- Hash-skip identical direct writes while still refreshing raw provenance.
  Inline attachment bytes now contribute their SHA-256 identity, and identical
  repeats repair missing session FTS rows before returning a skip.
- Make SQL audit finding keys stable under line movement while distinguishing
  duplicate sites.
- Expand token-output analysis across aliases, transforms, control flow,
  helpers, stdout/stderr, direct response streams, `_send_json`, `_send_error`,
  and journald-style sends.
- Resolve installed entry points from the checkout virtualenv first, fail
  rather than skip when they are absent, and include stderr in help failures.
- Preserve archive-generated `capture_gap` events across parser-owned session
  projection replacement, with a DOM-to-native-to-fuller-export regression.
- Freeze blob-repair fixture time and align CLI/demo expectations with current
  archive policy.
- Record the cause matrix and five independent adversarial review rounds in
  Beads (`polylogue-w9wt`, `polylogue-qkmd`, `polylogue-8jg9.2`). The review
  reached its five-round cap with findings in every round; every demonstrated
  reproduction is repaired, but this PR does not claim reviewer convergence.

## Verification

- Original baseline cluster: `13 passed in 20.94s`.
- Strengthened browser/direct-write/security/SQL selection: `25 passed in
  66.04s`.
- Cold-review regressions plus auth route matrix: `51 passed in 8.21s`.
- Seeded baseline at `9edb87062`: `13,235 passed, 1 skipped, 5 failed`; all five
  failures classified into the checkout-resolution and capture-gap clusters
  repaired by `f1f4e4452` and `56dd2ddfb`.
- Installed-script, demo-verification, and three-arrival storage selection:
  `13 passed in 28.87s` single-process, run
  `20260710T152213Z-focused-test-1228750-1c8a7f5f`.
- The same selection with `POLYLOGUE_PYTEST_WORKERS=4`: `13 passed in 16.86s`,
  run `20260710T152249Z-focused-test-1229069-ff2ec83f`.
- `devtools verify --quick`: all 13 steps passed in 15.80s; run
  `20260710T152506Z-quick-1230864-34851fbb` at `56dd2ddfb`.
- `devtools verify --seed-testmon --skip-slow` at `56dd2ddfb`: all static
  gates passed; pytest `13,241 passed, 1 skipped` in 290.19s, run
  `20260710T152627Z-seed-testmon-1231847-be1f2ab1`.

Ref `polylogue-w9wt`, `polylogue-qkmd`, and `polylogue-8jg9.2`.


Sinity:
@coderabbitai review

PR #2692, opened 2026-07-11
fix(storage): preserve semantic replay frontiers

Sinity:
## Summary

Preserve semantic authority when a later full snapshot updates a session bootstrapped through membership replay.

## Problem

Production recovery acquired six complete browser snapshots, but full replay always emitted byte frontiers. Their existing heads were semantic, so CAS rejected every update as incomparable even when the new semantic projection was strictly later.

## Solution

- inspect the existing logical head inside the replay transaction
- derive the accepted session projection frontier when that head is semantic
- retain byte frontiers for byte-governed full/append chains
- prove later semantic replacement advances while older and same-frontier conflicting replacements roll back session, messages, blocks, FTS, head, and receipts

Ref polylogue-yla8.4.

## Verification

- `devtools test tests/unit/storage/test_revision_replay.py` -> 7 passed
- `devtools verify --quick` -> 13/13 steps passed

### DIFF
--- a/polylogue/storage/sqlite/archive_tiers/archive.py
+++ b/polylogue/storage/sqlite/archive_tiers/archive.py
@@ -1930,6 +1930,22 @@ class ArchiveStore:
             write_source_blob_refs(self._ensure_source_conn(), raw_id, refs)
         session_ids: set[str] = set()
         with self._conn:
+            existing_head = self._conn.execute(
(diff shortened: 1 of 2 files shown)


Sinity:
Review triage: the interrupted implementation was taken over and independently diff-reviewed. The real replay regression is 7/7 green and proves both semantic advancement and transactional rollback of older/conflicting replacements; quick verification is 13/13. Automated findings will be triaged before merge.

PR #2713, opened 2026-07-11
perf(storage): bound provider usage full diagnostics (#2713)

Sinity:
## Summary

Bound provider-usage Python retention plus stale/cumulative reconstruction by session/model cardinality. The required provider-event totals remain one set-based event scan. Indexed frontier seeks, exact overflow fallbacks, durable raw-membership classifications, and one explicit model-whitespace contract make the active Codex report interactive while preserving the established zero-rollup comparison behavior.

## Problem

The first rewrite bounded returned stale samples but still scanned and sorted the 2.21M-row event table twice. It took 112.19 seconds and 1,098,080 KiB RSS on the active archive, while cumulative reconstruction fetched every matching event and raw-debt classification reopened already-classified blobs.

Final review also found two edge-contract gaps: SQLite's default space-only `TRIM` disagreed with the writer's Python `str.strip()` model normalization, and the optimized stale query treated event-only origins with no model rollup rows as newly stale.

## Solution

- Resolve the latest cumulative event through the existing `(session_id, position)` index and return at most one row per session.
- Stream only sessions without cumulative telemetry through Python arbitrary-precision counters, preserving accepted values that overflow SQLite `SUM`.
- Keep provider-event aggregation on its set-based event-scan path, with an exact streaming fallback for legal provider-counter aggregate overflow.
- Bind one explicit Python-compatible model-whitespace set into SQLite and reuse the same normalizer in streaming code.
- Preserve the established empty-rollup contract: without model rows, the stale audit has no comparison basis.
- Use durable `raw_membership_census` states before probing legacy raw blobs.
- Preserve session IDs losslessly in stale samples and remove the obsolete interpolated-SQL audit exception.

## Verification

- `devtools test tests/unit/storage/test_provider_usage_report.py tests/unit/storage/test_no_string_interpolated_sql.py::test_audited_sites_are_real_violations` - 17 passed at `f437ca4bb`.
- `devtools verify --quick` - all 13 checks passed; pre-push receipt `20260711T184110Z-quick-581184-e2469c26` at `f437ca4bb`.
- `.agent/scripts/bd-graph-lint` - no dependency cycles or violations.
- Earlier active-archive performance receipt for the unchanged frontier algorithm: 2,213,101 Codex events and 2,774 sessions improved from warmed 58.78 s / 1,632,480 KiB RSS to 4.70 s / 71,452 KiB RSS.
- That active report retained the same 13,709-byte SHA-256 `a75bea72927b866446faba09ed621cfd097757099beacdaf5ef18b1fac5aafe9` before and after the performance rewrite.
- Active `EXPLAIN QUERY PLAN` uses correlated scalar subqueries plus `idx_session_provider_usage_events_session`; it does not scan the provider-event table for cumulative/stale frontiers.

Anti-vacuity: default SQLite `TRIM` hides the multi-model whitespace-tail stale fixture; changing the overflow fallback's normalizer changes its missing-model count; removing the empty-rollup early return changes coverage state; restoring SQLite `SUM` breaks the accepted `2**62` counter case; removing `LIMIT 1` or the session-position predicate breaks plan assertions; removing census short-circuits triggers a forced blob-probe failure; delimiter packing breaks the control-character identity case.

Scope note: provider-event and cumulative counter overflow are covered here. Pre-existing SQL `SUM` limits in derived model/pricing rollups are separate from this PR.

Ref polylogue-xy95.

### DIFF
--- a/polylogue/storage/usage.py
+++ b/polylogue/storage/usage.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import json
 import sqlite3
-from collections import defaultdict
+from collections import Counter, defaultdict
 from collections.abc import Iterable
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -733,9 +733,11 @@ def _acquired_not_materialized_raw_rows(
                r.raw_id AS raw_id,
                r.source_path AS source_path,
                r.blob_hash AS blob_hash,
-               r.parsed_at_ms AS parsed_at_ms
+               r.parsed_at_ms AS parsed_at_ms,
+               c.status AS membership_census_status
         FROM {alias_sql}.raw_sessions AS r
         LEFT JOIN sessions AS s ON s.raw_id = r.raw_id
+        LEFT JOIN {alias_sql}.raw_membership_census AS c ON c.raw_id = r.raw_id
         {_where_origin(origin, table_alias="r")}
           {"AND" if origin is not None else "WHERE"} (r.parse_error IS NULL OR TRIM(r.parse_error) = '')
           AND s.session_id IS NULL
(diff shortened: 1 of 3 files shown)


Sinity:
CI classification for `f437ca4bb`: GitHub-hosted jobs did not execute any steps (`runner_id=0`, empty step/log sets). Their check annotations report: "The job was not started because your account is locked due to a billing issue." CodeRabbit and GitGuardian completed successfully. This is external to the patch; the final-head local gates remain 17 focused tests and all 13 quick checks (`20260711T184110Z-quick-581184-e2469c26`).

PR #2716, opened 2026-07-11
fix(storage): unify cross-route revision authority

Sinity:
## Summary

Unify single-session full captures and multi-session export members under one typed semantic revision authority when they describe the same logical session.

## Problem

Production post-repair catch-up proved the existing split was arrival-order dependent. An older Gemini full snapshot retried forever against a semantic head, while a ChatGPT browser capture for an account-export session collided at the same frontier. More seriously, a numerically larger but divergent single capture could pass cardinality-only CAS and overwrite the membership-governed session.

Ref polylogue-yla8.

## Solution

- Move only self-contained, append-independent full cohorts from byte revision fields into durable membership census rows.
- Handle both bundle-first and single-first arrival, including raws already bound by a previous failed deployed attempt.
- Transition a related byte head to semantic authority inside the same index transaction as the replacement and receipts.
- Keep divergent or insufficiently ordered metadata variants ambiguous and retry-visible.
- Accept metadata-only revisions only with pairwise-unique direct provider update timestamps.
- Stop browser capture from substituting provenance capture time for missing provider `updated_at`.
- Reuse one membership application helper and report the accepted cohort member's message count.

## Verification

- `devtools test -k 'single_session_full or bundle_promotes_prior_single_full or live_multi_session_divergence or metadata_revision or browser_capture_does_not_launder' tests/unit/sources/test_live_batch_support.py tests/unit/archive/test_session_revision_membership.py tests/unit/sources/test_browser_capture.py` — 7 passed.
- `devtools test tests/unit/archive/test_session_revision_membership.py` — 5 passed.
- `devtools test tests/unit/storage/test_revision_replay.py::test_full_replay_preserves_semantic_head_and_rolls_back_regressions` — 1 passed.
- `devtools verify --quick` — run `20260711T220233Z-quick-1445786-3516ef36`, 13/13 steps passed.
- Full `test_live_batch_support.py` result was 47 passed / 6 failed; all six exact failures reproduced unchanged on a clean detached `origin/master` checkout and are unrelated baseline failures.
- Three independent adversarial review rounds found and then verified closure of dual-governance, reverse-arrival, rebuild-stability, byte-head rollback, and timestamp-laundering blockers; final review found no release blocker.

[...]

PR #2717, opened 2026-07-11
fix(storage): apply authorized semantic head transitions

Sinity:
## Summary

Allow the membership classifier's proven metadata-only successor to replace an equal-frontier semantic head atomically.

## Problem

The first packaged retry after PR #2716 still rejected the production ChatGPT capture. Membership classification selected the later metadata revision using direct provider timestamps, but generic equal-frontier CAS rejected the changed session hash before updating the head.

Ref polylogue-yla8.

## Solution

Retire an existing accepted head, byte or semantic, only when its accepted raw belongs to the classifier's accepted/equivalent cohort and its reparsed projection hash exactly matches the stored accepted hash. Head retirement, authoritative replacement, FTS proof, immutable receipts, and the new semantic head remain one index transaction. Ambiguous/divergent classifications never reach this path.

## Verification

- Metadata-only live-route advancement plus larger-divergence containment: 2 passed.
- `devtools verify --quick`: final-head run `20260711T221806Z-quick-1498581-d007ebfe`, 13/13 passed.
- Independent adversarial review found no authority widening, rollback, or anti-vacuity blocker.
- Packaged production evidence before this patch: classifier reached equal-frontier CAS and failed with `raw revision CAS rejected a conflicting accepted head`; the new test reproduces that exact pre-patch failure.

### DIFF
--- a/polylogue/storage/sqlite/archive_tiers/archive.py
+++ b/polylogue/storage/sqlite/archive_tiers/archive.py
@@ -2121,7 +2121,7 @@ class ArchiveStore:
                     """,
                     (logical_source_key,),
                 ).fetchone()
-                if existing_head is not None and str(existing_head[2]) == "byte":
+                if existing_head is not None:
                     existing_raw_id = str(existing_head[0])
                     existing_projection = projections_by_raw_id.get(existing_raw_id)
                     classified_raw_ids = {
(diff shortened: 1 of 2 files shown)


Sinity:
GitHub-hosted jobs again failed before runner allocation under the known account billing lock. Final-head local quick run 20260711T221806Z-quick-1498581-d007ebfe passed 13/13 and the exact production-shaped metadata transition plus divergent containment passed 2/2. GitGuardian passed. CodeRabbit completed as a rate-limit/tool result with no review findings; an independent adversarial review found no authority-widening or rollback blocker.

PR #2751, opened 2026-07-12
feat(search): fold Polish l-stroke and remove_diacritics for FTS recall

Sinity:
## Summary

Closes the specific Polish-search recall hole named in polylogue-9jsi: `unicode61` FTS5 tokenization alone cannot fold `ł`/`Ł` (Latin L with stroke) because Unicode assigns it no canonical decomposition, so a plain-ASCII query like `latwo` never found seeded `łatwo` content. This PR adds a narrow, deterministic `pl_fold` layer around FTS5, expressed once and used identically on the write and query sides, plus enables `unicode61 remove_diacritics 2` (which handles the *other* Polish diacritics — `ó`, `ż`, `ą`, ... — on its own).

## Problem

Verified empirically before implementing (see `tests/unit/storage/test_pl_fold.py` and the PR discussion below): with `tokenize='unicode61 remove_diacritics 2'`, indexing `różowy` and querying `rozowy` already matches — the tokenizer folds ordinary combining-mark diacritics correctly and symmetrically on both sides. But `żółty` tokenizes to `zołty` (ż and ó folded, ł left untouched), because `ł`/`Ł` has no NFD/NFKD decomposition for `remove_diacritics` or Unicode normalization to reach. So even with the tokenizer bump alone, `latwo` would still miss `łatwo`, and `remove_diacritics` was not even enabled to begin with — both problems compound.

## Solution

- `polylogue/storage/fts/pl_fold.py`: a single `PL_FOLD_TABLE` (`ł`→`l`, `Ł`→`L`) expressed three ways so write-side and query-side folding cannot drift apart:
  - `pl_fold()` — Python implementation, used for query-side normalization.
  - `pl_fold_sql_expr()` — an inline SQL `REPLACE`-chain, embedded directly into trigger/insert DDL text. This is deliberately **not** a call to a registered custom function, because SQLite trigger bodies execute on whatever connection performs the write — including raw test/tooling connections that reuse the shared trigger DDL (`fts_lifecycle.py`'s `_BLOCKS_FTS_TRIGGER_DDL` etc.) without ever registering a UDF. An inline `REPLACE` chain has no such dependency.
  - `register_pl_fold()` — registers the identical transform as a SQL scalar function `pl_fold(x)`, wired into both sync (`storage/sqlite/connection.py`) and async (`storage/sqlite/async_sqlite.py`) read/write connection setup, for ad-hoc queries, diagnostics, and the Python/SQL agreement test.
- All three `unicode61` contentless FTS surfaces — `messages_fts`, `threads_fts`, `session_work_events_fts` — move to `tokenize='unicode61 remove_diacritics 2'` and fold `search_text` through `pl_fold_sql_expr` before insertion, at every DDL/trigger/bulk-insert site (`archive_tiers/index.py`, `fts/fts_lifecycle.py`, `fts/sql.py`). `threads_fts`/`session_work_events_fts` were included alongside `messages_fts` (beyond the bead's literal example) because query-side folding via `escape_fts5_query` is already global across all FTS surfaces — leaving those two write-side unfolded would have reintroduced exactly the "silent recall loss" drift the bead calls out as the top risk.
- `escape_fts5_query` — the single chokepoint every production `MATCH` builder in the archive routes through (8+ call sites across `archive_tiers/archive.py`, `search/query_builders.py`, `search_providers/fts5.py`, `sqlite/queries/session_insight_*_queries.py`, `insights/resume.py`) — applies the identical Python `pl_fold` to query text before quoting.
- `INDEX_SCHEMA_VERSION` 34 → 35 (derived-tier, no migration chain — existing archives rebuild via `polylogue ops reset --index && polylogued run`). `docs/internals.md` and `docs/search.md` updated.

**Deliberately out of scope** (per the bead's own AC, which requires trigram to stay off "until benchmarked with size+precision report" and never silently default): no trigram fallback lane is added here. Tracked as follow-up polylogue-xul7.

**Also discovered, filed, left unfixed** (confirmed pre-existing and unrelated via `git log`): `tests/unit/storage/test_archive_tiers_assertions.py::test_fresh_user_tier_has_no_legacy_overlay_tables` fails on any fresh checkout since #2703 added `context_deliveries` to `user.db` without updating this assertion. Filed as polylogue-gxly.

## Verification

- `devtools test tests/unit/storage/test_pl_fold.py` — 12 passed. Covers: `ł`/`Ł`-only folding, idempotency, Python/registered-UDF/inline-REPLACE-chain 3-way agreement, the two-canonical-DDL-site tokenizer drift lock, and end-to-end recall (`latwo`/`zrobilem` finding seeded `łatwo`/`zrobiłem`, plus a literal `ł` query, plus `remove_diacritics` alone finding `różowy`/`zrobię`) against the real trigger-backed schema via the `test_conn` fixture.
- `devtools test tests/unit/storage/test_fts5.py tests/unit/storage/test_fts_repair_sql.py tests/unit/storage/test_fts_bloat_invariants.py tests/unit/storage/test_dangling_fts_derived_surfaces.py tests/unit/storage/test_index.py tests/unit/storage/test_fts5_query_correctness.py tests/unit/storage/test_fts_escape_in_insights.py tests/unit/storage/test_blocks_command_trigram.py` — 211 passed.
- `devtools test tests/unit/storage/test_archive_tiers_assertions.py tests/unit/storage/test_schema_policy_contracts.py tests/unit/cli/test_status_diagnostics.py tests/unit/cli/test_tutorial.py` — 63/64 passed (1 pre-existing unrelated failure, polylogue-gxly).
- `devtools test -k "test_polylogued_status_json_reports_schema_mismatch_not_ready or test_polylogued_status_plain_reports_schema_mismatch" tests/unit/daemon/test_daemon_cli.py` — 2 passed (the schema-version-sensitive daemon status tests; the full `test_daemon_cli.py` file was not run in full, it contains an unrelated slow periodic-loop test).
- `mypy --strict` on every touched file — clean.
- `devtools render all --check` — clean (topology projection regenerated for the new module).
- Pre-push `devtools verify --quick` (format + lint + mypy + render-all-check) — passed.

Ref polylogue-9jsi


Sinity:
Superseded by merged PR #2754. The replacement includes the original Polish fold plus the audited ArchiveStore rebuild and dangling-repair coverage.

PR #2753, opened 2026-07-12
feat(search): fold Polish diacritics across every FTS write path

Sinity:
## Summary

Add deterministic Polish lexical folding for every FTS5 surface and every write, rebuild, and repair route. Plain-ASCII queries such as `latwo` and `zrobilem` now find `łatwo` and `zrobiłem` content without introducing an unmeasured trigram fallback. This replacement is based on current `master` and supersedes conflicted PR #2751.

## Problem

SQLite FTS5 `unicode61` does not fold `ł`/`Ł` because those code points have no Unicode decomposition. `remove_diacritics 2` handles ordinary Polish diacritics, but not l-stroke.

The first implementation correctly folded fresh-write triggers and normal queries, but audit found three production maintenance paths that still inserted raw `search_text`: the public `ArchiveStore.rebuild_index()` path and the dangling-row repair paths for threads and session work events. Because query text was always folded, running one of those maintenance routes silently removed the new recall guarantee for the repaired rows.

## Solution

- Add one `PL_FOLD_TABLE` for `ł` -> `l` and `Ł` -> `L`, with Python, registered SQLite-function, and inline SQL-expression projections derived from it.
- Move `messages_fts`, `threads_fts`, and `session_work_events_fts` to `unicode61 remove_diacritics 2`.
- Apply the inline fold at fresh-write triggers, bulk/missing-row writers, full rebuilds, the public message-index rebuild, and both keyed dangling-derived repair paths.
- Fold query text at the existing `escape_fts5_query` chokepoint used by production `MATCH` builders.
- Keep `INDEX_SCHEMA_VERSION` at 35. This is a rebuildable derived-tier change; existing archives use `polylogue ops reset --index && polylogued run`, with no in-place migration helper.
- Regenerate topology from current master and document the complete rebuild/repair contract.

## Acceptance criteria

| Criterion | Status | Evidence |
| --- | --- | --- |
| `latwo` / `zrobilem` find seeded `łatwo` / `zrobiłem` | Satisfied | Real trigger-backed tests in `test_pl_fold.py` |
| Fold is idempotent and Python/SQL projections agree | Satisfied | Python, registered-UDF, and inline-REPLACE agreement tests |
| Canonical DDL tokenizer sites cannot drift | Satisfied | DDL tokenizer contract tests for all three surfaces |
| Production `MATCH` builders normalize query text | Satisfied | Shared `escape_fts5_query` / `normalize_fts5_query` routing and call-site audit |
| Rebuild and repair writers preserve folding | Satisfied | Public `ArchiveStore.rebuild_index()` test plus actual dangling repair test for both derived surfaces |
| Trigram remains off until size and precision are measured | Satisfied | No trigram fallback added; follow-up `polylogue-xul7` carries the measured opt-in lane |
| Derived schema change has an explicit rebuild plan | Satisfied | Index version 35 plus documented reset/reingest command |

Anti-vacuity: removing the fold from `rebuild_archive_messages_fts` makes the public rebuild test lose both ASCII and literal Polish hits. Removing either fold from `dangling_repair.py` makes the corresponding thread or work-event assertions fail after `repair_missing_fts_rows`; the tests seed source rows before repair and do not rely on fresh-write triggers.

## Verification

- `devtools test tests/unit/storage/test_pl_fold.py` — 13 passed.
- `devtools test tests/unit/storage/test_fts_repair_sql.py` — 8 passed.
- `devtools test tests/unit/storage/test_dangling_fts_derived_surfaces.py` — 11 passed.
- `devtools test tests/unit/storage/test_schema_policy_contracts.py` — 14 passed.
- `devtools test -k schema_mismatch tests/unit/daemon/test_daemon_cli.py` — 2 passed, 73 deselected.
- `devtools render topology-projection` — wrote 912 rows, 6 existing TBD cells.
- `devtools render topology-status` — clean.
- `devtools render all --check` — exit 0; all generated surfaces in sync.
- `.agent/scripts/bd-graph-lint` — reports two inherited master missing-AC records (`polylogue-2ilz`, `polylogue-nu2h`); no dependency cycles, duplicate labels, or inversions, and this branch does not modify either record.
- `git diff --check` — exit 0.
- `jq -e -c . .beads/issues.jsonl >/dev/null` — exit 0.
- `devtools verify --quick` — exit 0; repeated successfully by the pre-push hook at commit `31eb97de6`.

Ref polylogue-9jsi


Sinity:
Superseded by merged PR #2754, rebuilt from current master with the schema-version documentation fix and equivalent focused verification.

PR #2790, opened 2026-07-12
feat(mcp): bound agent response payloads

Sinity:
## Summary

Bounds MCP responses for agent use, makes search affordances opt-in, adds message body caps/excerpts, turns get_session_summary into a bounded orientation payload, and completes the response-envelope/pagination ergonomics with actionable query errors, distinct list rows, and zero-hit diagnostics.

## Problem

Live MCP searches and message reads could return response bodies far beyond an agent context budget. Search carried the full affordance catalog on every response, session summaries contained metadata only, invalid filters gave agents no recovery set, list search could repeat a session for each matching block, and an AND-style search miss could not explain its terms.

## Solution

- Add a shared 25 KB serializer budget that replaces oversized payloads with metadata plus a replayable narrower MCP call descriptor.
- Make `search(include_affordances=true)` the explicit opt-in for its affordance catalog.
- Add `max_chars_per_message` and head/tail excerpt handling to `get_messages`.
- Add counts, phase count, top tools, and first/last authored-user excerpts to `get_session_summary`.
- Map `QuerySpecError` into an `invalid_query` envelope with field, tool, and valid-values metadata.
- Coalesce block-level `list_sessions` results into distinct session rows with `match_count`, using the existing distinct-session count and only the raw hits needed for the requested page.
- Include filtered per-term counts only when a multi-term `search` has zero total matches; exhausted result pages remain ordinary empty pages.

## Verification

- `nix develop --command devtools test tests/unit/mcp/test_tool_contracts.py -k 'multi_term or coalesces or invalid_sort'` — 3 passed.
- `nix develop --command devtools test tests/unit/mcp/test_query_tool_schema_derivation.py tests/unit/mcp/test_tool_contracts.py -k 'supported_request_fields or stops_after_requested or exhausted_page or multi_term or coalesces or invalid_sort'` — 6 passed.
- `nix develop --command devtools test tests/unit/mcp -k 'envelope or pagination or summary'` — 178 passed, 473 deselected (1:55).
- Each pushed commit passed the pre-push quick gate: format, lint, mypy, render, topology, layering, schema roundtrip, manifests, and MCP test-infrastructure checks.

## Acceptance status

Ref polylogue-rsad.

- Satisfied: opt-in affordances; shared response-budget envelopes across the covered read and mutable-list paths; per-message cap/match-aware excerpt behavior; content-bearing session summaries; direct and search-DSL invalid-value envelopes; distinct session-list rows with `match_count`; and zero-hit per-term diagnostics.
- Remaining polylogue-rsad scope: `query_units` must use the shared valid-values error shape and preserve replay arguments; `archive_search_sessions` must preserve replay arguments. These are documented in adversarial iteration 5; the bead remains open.
- Excluded: recall/context tool changes remain owned by the hermes-wedge lane.

## Test evidence

The tests invoke registered MCP handlers against real seeded archive rows, not test-only validators:

- Removing the `QuerySpecError` mapper or its valid-values mapping breaks the invalid-sort test.
- Removing the archive search-result coalescing or `match_count` projection breaks the repeated-block session-list test.
- Removing the distinct-page bound makes the coalescing test fetch an unnecessary second raw FTS batch.
- Emitting diagnostics based on an empty page rather than zero total breaks the exhausted-page test.
- Existing lane tests similarly exercise the shared budget serializer/context, affordance opt-in pass-through, message projection cap, and summary enrichment.

### DIFF
--- a/polylogue/mcp/archive_support.py
+++ b/polylogue/mcp/archive_support.py
@@ -336,14 +336,15 @@ def archive_session_list_payload(
         # ``list_summaries`` already applied offset and limit above.
         page = summaries
     else:
+        total = archive.count_search_sessions(text_query, **filters)
         summaries, match_counts = _coalesced_search_summaries(
             archive,
             query=text_query,
             filters=filters,
             sort=_sort_value(spec.sort),
             reverse=spec.reverse,
+            unique_limit=min(total, offset + limit),
         )
-        total = len(summaries)
         page = summaries[offset : offset + limit]
     next_offset = offset + len(page) if offset + len(page) < total else None
     return MCPPaginatedQueryResultPayload(
@@ -364,13 +365,14 @@ def _coalesced_search_summaries(
     filters: ArchiveQueryFilters,
     sort: str | None,
     reverse: bool,
+    unique_limit: int,
 ) -> tuple[tuple[ArchiveSessionSummary, ...], dict[str, int]]:
-    """Coalesce block-level FTS rows into stable, unique session list rows."""
+    """Coalesce only the raw hits required to construct one session page."""
     chunk_size = 250
     raw_offset = 0
     summaries_by_id: dict[str, ArchiveSessionSummary] = {}
     match_counts: dict[str, int] = {}
-    while True:
+    while len(summaries_by_id) < unique_limit:
         hits = archive.search_summaries(
             query,
             limit=chunk_size,
@@ -384,6 +386,10 @@ def _coalesced_search_summaries(
         for hit in hits:
             summaries_by_id.setdefault(hit.session_id, archive.read_summary(hit.session_id))
             match_counts[hit.session_id] = match_counts.get(hit.session_id, 0) + 1
+            if len(summaries_by_id) >= unique_limit:
+                break
+        if len(summaries_by_id) >= unique_limit:
+            break
         raw_offset += len(hits)
         if len(hits) < chunk_size:
             break
@@ -439,7 +445,7 @@ def archive_search_payload(
         **filters,
     )
     total = archive.count_search_sessions(query, **filters)
-    diagnostics = _search_term_diagnostics(archive, query=query, filters=filters, spec=spec) if not hits else None
+    diagnostics = _search_term_diagnostics(archive, query=query, filters=filters, spec=spec) if total == 0 else None
     return build_search_envelope(
         tuple(archive_search_hit_payload(hit, archive=archive) for hit in hits),
         total=total,
--- a/polylogue/mcp/archive_support.py
+++ b/polylogue/mcp/archive_support.py
@@ -171,13 +171,15 @@ def archive_matched_summary_payload(
     summary: ArchiveSessionSummary,
     *,
     match_count: int,
+    match_count_is_exact: bool,
 ) -> MCPMatchedSessionSummaryPayload:
-    """Add exact raw-match cardinality to one coalesced session row."""
+    """Add observed raw-match cardinality to one coalesced session row."""
     from polylogue.mcp.payloads import MCPMatchedSessionSummaryPayload
 
     return MCPMatchedSessionSummaryPayload(
         **archive_summary_payload(summary).model_dump(mode="python"),
         match_count=match_count,
+        match_count_is_exact=match_count_is_exact,
     )
 
 
@@ -310,7 +312,11 @@ def archive_session_list_payload(
         total = offset + len(page) + (1 if next_offset is not None else 0)
         return MCPPaginatedQueryResultPayload(
             items=tuple(
-                archive_matched_summary_payload(summary, match_count=match_counts[summary.session_id])
+                archive_matched_summary_payload(
+                    summary,
+                    match_count=match_counts[summary.session_id],
+                    match_count_is_exact=False,
+                )
                 for summary in page
             ),
             total=total,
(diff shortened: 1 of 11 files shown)


Sinity:
## Adversarial review — iteration 1

Findings:

- Real gap: oversized session-scoped non-query reads could fall back without replay arguments.
- Real gap: a bounded lexical coalescing page could report a partial raw-match count as if it were exact.
- Real gap: recursively nested fallback metadata could itself exceed the 25 KB budget.
- Borderline: the ratified affordance shape asked for a one-time MCP resource; the existing tool met the original AC but not that exact shape.
- Borderline: this implementation fixes the shared response budget at 25 KB. The bead’s “configured budget” wording is interpreted here as that shared server policy; adding a user-configurable setting would require a config-surface decision outside this lane.

Fixes in `0a42f249b`:

- Safe-call wrappers now provide session-scoped replay identifiers; an oversized `archive_get_session` response returns a resolvable continuation.
- `match_count_is_exact` distinguishes fully scanned counts from page-bounded observed counts.
- Budget metadata compaction now has a depth limit, preserving the hard response bound for nested models.
- Added `polylogue://capabilities/action-affordances`, reusing the existing catalog without reintroducing search boilerplate.

What survived scrutiny:

- Per-message cap/excerpt behavior, sort valid-values errors, zero-total diagnostics, normal distinct-page traversal, and the 25 KB replay envelope all remain covered by production-route tests.
- Focused verification: 5 passed for the iteration-specific regressions; pre-push quick gate passed.
- No Bead state changed; Ref polylogue-rsad.

Sinity:
## Adversarial review — iteration 2

Findings:

- Real gap: the `archive_get_session` continuation repeated the same unbounded call.
- Real gap: `list_corrections` bypassed the shared serializer and had no bounded retrieval path.
- Real gap: the messages resource had no callable continuation when its fixed resource page exceeded the budget.
- Borderline retained: 25 KB is a fixed shared server policy in this implementation, not an operator-configurable setting.

Fixes in `d55ad14fa`:

- `archive_get_session` now falls back to capped `get_messages`.
- `list_corrections` now uses the shared serializer, exposes limit/offset, and caps correction text on fallback replay.
- `polylogue://messages/{conv_id}` now falls back to capped `get_messages`.

What survived scrutiny:

- Three end-to-end tests invoke each advertised continuation and verify it returns a non-fallback bounded response.
- Focused verification: 3 passed, with `POLYLOGUE_PYTEST_BASETEMP_ROOT=/realm/tmp/polylogue-pytest`; pre-push quick gate passed.
- No Bead state changed; Ref polylogue-rsad.

Sinity:
## Adversarial review — iteration 3

Findings:

- Real gap: `list_tags` could still exceed the response budget without a size-limited, replayable call.
- Real gap: excerpt mode retained only head/tail text, so it could omit the matched evidence that prompted a read.
- Real gap: unknown `origin` query filters silently normalized to `unknown-export` instead of returning the established valid-values error.

Fixes in `252d182d8`:

- `list_tags` now has deterministic `limit`/`offset` paging and records its arguments for a bounded replay continuation.
- `get_messages(match_query=...)` centers a bounded excerpt on the first requested match span and keeps that query in replay arguments.
- MCP query construction rejects unknown `origin` and `exclude_origin` tokens through the typed `invalid_query` envelope.

What survived scrutiny:

- New production-route tests invoke the registered MCP handlers and actual response serializer. Removing tag slicing/context, match-span selection, or origin validation makes the corresponding test fail.
- Focused regressions: 3 passed. Required MCP selection: 181 passed (`tests/unit/mcp -k 'envelope or pagination or summary'`), with pytest temp rooted at `/realm/tmp/polylogue-pytest`.
- Push quick gate passed (format, lint, mypy, generated/layering checks). No Bead state changed.

Ref polylogue-rsad.

Sinity:
## Adversarial review — iteration 4

Findings:

- Real gap: mutable list tools and `polylogue://tags` could still yield an unchanged or absent continuation after an oversized response.
- Real gap: invalid closed values inside the query DSL bypassed the same valid-values envelope used by direct query parameters.

Fixes in `9415cd2ec` and `8635af3f8`:

- Mutable list responses now carry `limit`/`offset`/`next_offset`, retain replay context, and use compact per-item output on bounded replay; the tags resource replays through paged `list_tags`.
- `ExpressionCompileError` now maps to `invalid_query`, including the closed-vocabulary recovery values for DSL `origin` and `action` errors.

What survived scrutiny:

- Registered-handler tests force both a large annotations list and a large tags resource through their advertised continuations (2 passed).
- Registered `search` calls with bad DSL origins/actions return the typed recovery envelope (4 passed including direct-origin/sort coverage).
- Both pushed commits passed the pre-push quick gate. No Bead state changed.

Ref polylogue-rsad.

Sinity:
## Adversarial review — iteration 5 (loop cap reached)

Remaining real gaps:

- `query_units` locally catches DSL compilation errors, so unknown closed values there bypass the shared `invalid_query` field/valid-values envelope. Correction-kind errors likewise remain an `unknown_kind` payload without the closed-vocabulary recovery set.
- `query_units` and `archive_search_sessions` can overflow without preserving their required `expression`/`query` arguments in response context, producing a continuation that cannot be invoked.

No changes were made in this iteration. This is the fifth independent cold review, the adversarial-loop cap; I am stopping rather than claiming convergence.

Remaining `polylogue-rsad` work is therefore not complete. The current branch stays pushed and clean; no Bead state changed.

Ref polylogue-rsad.

PR #2796, opened 2026-07-12
fix(embeddings): make embedding lifecycle actionable

Sinity:
## Summary

Makes embedding lifecycle state explicit and safe for the current v35 index-generation model. Orphan cleanup requires active, source-snapshotted generation evidence; terminal failures retain inspectable lifecycle history and operator resolution actions.

## Problem

A current-schema index without rebuild readiness evidence was still eligible to authorize orphan deletion. Archive file-set deployments also expose index.db through a public symlink while keeping generation metadata beside the database-tier pointer anchor. Terminal provider failures could remain aggregate-only debt, while repeated attempts inflated active debt and acknowledged exclusions could be misreported as either backlog or complete coverage.

## Solution

- Require an active generation pointer, active metadata, and a non-empty source snapshot for every orphan-cleanup apply; retain schema and inode checks as additional guards.
- Resolve generation metadata beside the validated pointer anchor, so a public index.db symlink uses the authoritative external tier without relaxing deletion authority.
- Persist all status-error writes in the lifecycle ledger, retaining session/message refs, provider/model, error class, timestamps, and durable resolution evidence.
- Supersede an earlier active event when a later failure occurs for the same session; successful embedding resolves current events.
- Expose bounded actionable identities and a shell-valid resolution command through shared CLI/MCP status. Distinguish acknowledged or superseded terminal exclusions from automatic retry backlog and coverage.

## Current live v35 readiness

Read-only inspection found the public pointer anchor at /realm/db/polylogue/index.db resolving to gen-v35-fastforward-1783887475997-88c34860. The active index is schema v35, matching the packaged schema, and exactly one matching metadata record has a non-empty source snapshot. That record is state=inactive, so it is not authoritative deletion truth. No reconciliation apply or other live archive mutation was performed. The v35 activation owner must mark the matching generation active under the recorded cutover protocol before a fresh inspect and any bounded apply can be considered.

## Acceptance criteria

| Bead | Status | Evidence |
| --- | --- | --- |
| polylogue-1dk1 | Satisfied in code; live apply deferred | Bounded identity cleanup is idempotent, revalidates generation identity before commit, and requires active source-snapshotted v35-style authority. The external-tier regression uses real index and vec/meta databases; it deletes only after finding active metadata beside the pointer anchor. Live apply remains gated by the inactive metadata record. |
| polylogue-egm8 | Satisfied in code | Current retryable and terminal rows are bounded and inspectable with source refs. Explicit acknowledgement, supersession, and requeue preserve audit history; only retryable work returns to automatic catch-up. |

## Verification

- nix develop --command devtools test tests/unit/storage/test_embedding_orphan_reconcile.py — 16 passed.
- nix develop --command devtools test tests/unit/storage -k embedding — 89 passed, 1735 deselected.
- Pre-push quick gate passed: Ruff format/lint, strict mypy, rendering, topology/layering, schema roundtrip, manifests, and repository verification.
- Live readiness inspection was read-only: pointer/schema/metadata only; no apply flag or archive write was issued.

### Anti-vacuity

- test_apply_accepts_generation_metadata_beside_external_pointer_anchor uses a real public index symlink, pointer anchor, active generation metadata, index database, and vec/meta database. Reverting metadata lookup to the symlink directory makes its authorized deletion fail.
- test_apply_refuses_inactive_generation_even_when_schema_matches and test_apply_refuses_generationless_index_even_when_schema_matches use real index and vec/meta databases. Removing either generation-authority guard permits the asserted orphan deletion.
- test_embedding_failure_lifecycle_preserves_audit_and_requeues and test_new_failure_supersedes_prior_active_failure_for_same_session exercise the production lifecycle writer against fresh embeddings DDL. Removing audit retention, requeue status mutation, or supersession leaves their asserted state wrong.
- test_status_detail_exposes_bounded_terminal_failure_resolution and test_status_excludes_acknowledged_terminal_failure_from_retry_backlog exercise the real archive status payload and renderer. Removing lifecycle detail, blocked-session accounting, or the pending-message exclusion makes the status assertions fail.
- test_resolve_failure_cli_requeues_terminal_failure invokes the Click command against an index and embeddings file pair. Removing its call to the production resolver leaves the failure active and retry status unset.

### Direct adversarial review

Seven direct passes found and fixed: non-detail terminal misclassification; bounded-detail aggregates; missing plaintext identities; shell-invalid resolution guidance and no command-path test; duplicate active attempts; acknowledged rows misclassified as retry backlog or complete; generation-less cleanup authority; public writer bypass; timeout wording; and finally the live external pointer-anchor lookup. The current live check found an inactive generation record, so the destructive step remains deliberately blocked rather than being treated as a code-path failure.

Ref polylogue-1dk1

Ref polylogue-egm8


Sinity:
Review triage (all CodeRabbit findings):

- **Route resolution through the daemon writer (Major)** — accepted as valid architecture debt, not fixed here: `polylogue ops embed resolve-failure` is an operator break-glass command and the daemon has no exposed mutation path for embedding lifecycle yet. Follow-up bead filed (see polylogue-20d.1 lane: once the daemon write path exists, this CLI mutation should dispatch through it). Merge-acceptable per single-writer threat model: the command takes the same sqlite connection discipline as other ops break-glass commands.
- **'timed out' classification (Minor)** — fixed in b05707911.
- **Success paths skip failure resolution (Major)** — valid; fixed in 16ce55fd5 by moving resolution into `_record_archive_embedding_success` so all three terminal success outcomes clear open failures. Regression test: `test_archive_success_outcomes_resolve_open_failures`.
- **Local faults ledgered as Voyage provider failures (Major)** — valid; fixed in 16ce55fd5 with `_ProviderRequestError` wrapping: only exceptions from the provider call carry `provider=voyage`; local faults record `provider=local` / `error_class=internal_error`. Tests: `test_archive_local_fault_is_not_ledgered_as_provider_failure`, `test_archive_provider_fault_keeps_provider_attribution`.
- **needs_reindex left set on acknowledge/supersede (Major)** — valid; fixed in 16ce55fd5: non-requeue resolutions clear `needs_reindex` but keep `error_message`, moving the session from the pending backlog into the visible blocked count (matches this branch's blocked-session contract). Test: `test_acknowledging_retryable_failure_stops_auto_requeue_but_stays_blocked`.
- **ACTION placeholder (Trivial)** — fixed in 16ce55fd5: renders `--action <acknowledge|requeue|supersede>`.

Verification: `devtools test tests/unit/storage/test_archive_tiers_embedding_write.py tests/unit/storage/test_embedding_contracts.py tests/unit/cli/test_embed_status_fast.py` → 79 passed. `devtools verify --quick` green at 16ce55fd5. The one broader-selection failure (`test_periodic_embedding_backlog_waits_for_catch_up_complete` timeout) reproduces on pristine master — pre-existing, bug bead being filed.

PR #2798, opened 2026-07-12
test(storage): model write-path lineage interleavings

Sinity:
## Summary

Add a Hypothesis stateful property test for archive write-path interleavings.

## Problem

The existing lifecycle state machine does not model full replacement, append windows, prefix-sharing lineage, stale writes, variants, or dangling branch points—the combinations most likely to hide write-path regressions.

## Solution

Add `tests/property/test_write_path_state_machine.py`. It writes `ParsedSession` data through `write_parsed_session_to_archive` and reads through `read_archive_session_envelope`; it also checks actual `session_links` and FTS docsize tables. The model covers parent/child replay, edited full replacement, merge append with variants, stale replacement, and one hard-deleted branch point.

For the known 4ts.6 behavior, an incomplete lineage is required to be a non-overextending suffix of the modeled transcript.

## Acceptance criteria

- [x] Six operation classes and five requested invariants are exercised.
- [x] Default and CI Hypothesis profiles finish in seconds.
- [x] No production divergence was found; no follow-up bead is proposed.

## Verification

- `devtools test tests/property/test_write_path_state_machine.py` — `1 passed in 6.32s`
- `HYPOTHESIS_PROFILE=ci devtools test tests/property/test_write_path_state_machine.py` — `1 passed in 5.32s`
- `ruff format --check tests/property/test_write_path_state_machine.py && ruff check tests/property/test_write_path_state_machine.py && mypy --strict tests/property/test_write_path_state_machine.py` — formatted, lint-clean, and `Success: no issues found in 1 source file`

Anti-vacuity: every rule exercises the production SQLite writer and composed archive reader, then queries production `session_links` and `messages_fts_docsize`; no mock or test-only write model is used.

Ref polylogue-g9f2.

### DIFF
--- a/tests/property/test_write_path_state_machine.py
+++ b/tests/property/test_write_path_state_machine.py
@@ -7,8 +7,11 @@ not duplicate prefix extraction, FTS maintenance, or session-link resolution.
 
 from __future__ import annotations
 
+import asyncio
 import sqlite3
+import tempfile
 from dataclasses import dataclass
+from pathlib import Path
 
 from hypothesis import HealthCheck, settings
 from hypothesis.stateful import RuleBasedStateMachine, initialize, rule
@@ -18,6 +21,8 @@ from polylogue.archive.session.branch_type import BranchType
 from polylogue.core.enums import Provider, TopologyEdgeStatus
 from polylogue.pipeline.ids import session_content_hash
 from polylogue.sources.parsers.base import ParsedMessage, ParsedSession
+from polylogue.storage.repository import SessionRepository
+from polylogue.storage.sqlite.archive_tiers.bootstrap import initialize_active_archive_root
 from polylogue.storage.sqlite.archive_tiers.write import (
     ArchiveSessionEnvelope,
     read_archive_session_envelope,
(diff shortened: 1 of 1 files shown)


Sinity:
## Adversarial review — iteration 1

Findings and fixes:

- Real gap: the model tested the sync envelope but not the public `SessionRepository.get_messages` route. Fixed in `b08afa9b5` with a real split-file scratch archive under `/realm/tmp`.
- Real gap: only prefix-sharing children were generated. Added a known-parent, zero-shared-prefix rule and exact `spawned-fresh`/null-branch-point assertions.
- Real gap: link checking was a permissive status-set check and never exercised deferred resolution. Added child-before-parent plus late-parent rules, asserting the exact unresolved-to-prefix-sharing transition and physical tail extraction.

What survived scrutiny: FTS docsize parity still compares the production docsize table with text-bearing stored blocks; dangling-branch behavior remains constrained to a non-overextending suffix.

Borderline recorded: this bounded state machine covers the dangling-branch 4ts.6 path; existing focused lineage tests own the depth-limit path.

Sinity:
## Adversarial review — iteration 2

Findings and fixes:

- Real gap: variants were append-only. Added a full-replace same-position sibling-variant rule with exact `(position, variant_index)` assertions.
- Real gap: quarantine was unreachable. Added a real async resolver cycle test that requires `status=quarantined`, no resolved destination, and cycle-rejection evidence.
- Borderline fixed: FTS parity now compares exact docid sets (`messages_fts_docsize.id` versus indexable `blocks.rowid`), not only cardinality.

What survived scrutiny: spawned-fresh and deferred parent resolution remain exact-link assertions; the public repository composed-read test remains real-route coverage.

Verification: the quarantine selector passed (`1 passed`; 6.96s test call). The pre-push quick gate passed format, lint, and mypy; GitHub Actions remains the known account-init failure pattern.

PR #2828, opened 2026-07-13
fix(storage): prove quarantined browser repair witness (#2828)

Sinity:
## Summary

Make the v35 browser-origin repair accept only the real quarantined singleton witness shape.

## Problem

The three live lkrc raws have complete canonical membership censuses with `decision = NULL`, which is correct for quarantined raws. The prior hardening incorrectly required `applied`, making the legitimate repair path unreachable.

## Solution

Require an undecided quarantined membership plus exactly one immutable old unknown-key `selected_baseline` receipt matching the session, raw, byte digest, frontier, generation, and baseline. Keep non-null membership decisions and missing old receipts fail-closed.

## Verification

- `devtools test tests/unit/storage/test_browser_capture_origin_repair.py` — 20 passed.
- `ruff check polylogue/storage/repair.py tests/unit/storage/test_browser_capture_origin_repair.py` — all checks passed.
- `mypy polylogue/storage/repair.py` — success, no issues.
- Pre-push `devtools verify --quick` completed before push.

Ref polylogue-lkrc.

### DIFF
--- a/polylogue/storage/repair.py
+++ b/polylogue/storage/repair.py
@@ -1175,7 +1175,7 @@ def _verify_browser_origin_copy_forward_source_stage(
             len(projection.message_hashes),
             0,
             RawRevisionAuthority.QUARANTINED.value,
-            "applied",
+            None,
         )
         or census is None
         or tuple(census) != ("complete", 1)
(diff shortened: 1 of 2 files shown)


Sinity:
@codex review

PR #2836, opened 2026-07-13
fix(storage): reconcile proven raw authority

Sinity:
## Summary

Replace the fixed source-v7 repair cohort with evidence-bound authority proof,
and let browser-origin repair produce a new byte-proven canonical raw when the
only canonical head is semantic evidence.

## Problem

A legacy source tier has no durable `capture_mode`, so a raw-ID allowlist made
otherwise proven injective origins impossible to reconcile. One ChatGPT
accepted raw also lacks membership/census rows, while its same-path cohort is
available for exact parser proof. Separately, a semantic canonical browser
head cannot be treated as a byte-proven replacement, but should not block a
new byte-proven canonical copy.

## Solution

- Require an injective `Origin` provider fiber on source-v7 rows, retaining the
  capture-mode requirement for non-injective origins.
- Prove and stage a same-path, quarantined, singleton census cohort inside the
  source transaction, then CAS only explicitly requested accepted raws.
- Treat an exact semantic canonical browser head as historical evidence; CAS
  its derived head to the new byte-proven copy-forward raw while retaining the
  old raw and application evidence.
- Add success, refusal, and idempotency-oriented regression coverage for both
  repair lanes.

Ref polylogue-lkrc.

## Verification

- `POLYLOGUE_PYTEST_WORKERS=0 devtools test ...source-v7 and semantic repair nodes...`
  - `7 passed in 22.67s`
- `devtools verify --quick`
  - format, lint, strict mypy, render, topology/layering, schema roundtrip,
    and all quick verification gates passed.

The earlier two-file focused run was stopped after 90 seconds without progress;
the exact replacement node selection above completed normally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repairs for quarantined accepted records when membership and census evidence is incomplete.
  * Added safeguards to ensure repairs use consistent, same-source evidence and reject mixed or insufficiently proven data.
  * Improved browser-capture repairs by recognizing valid equivalent canonical records and preserving the correct semantic history.
  * Added post-repair validation to ensure repaired records converge to the expected final state.

### DIFF
--- a/polylogue/storage/repair.py
+++ b/polylogue/storage/repair.py
@@ -909,7 +909,7 @@ def _inspect_quarantined_accepted_raw(
         ),
         parallel_session_head_count=len(parallel_session_heads),
         quarantined_sibling_raw_count=len(competing_revision_rows),
-        membership_row_count=len(membership_rows),
+        membership_row_count=0 if census_stage_raw_ids else len(membership_rows),
         census_stage_raw_ids=census_stage_raw_ids,
         parsed_message_count=len(parsed.messages),
     )
--- a/polylogue/storage/repair.py
+++ b/polylogue/storage/repair.py
@@ -451,6 +451,18 @@ def _stage_quarantined_census_cohort(
         row = conn.execute("SELECT raw_id FROM raw_sessions WHERE raw_id = ?", (staged_raw_id,)).fetchone()
         if row is None:
             raise RuntimeError(f"census staging witness changed for {staged_raw_id}")
+        membership_count = int(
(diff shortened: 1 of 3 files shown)


Sinity:
Automated-review triage: the cohort budget, staged-row reinsertion, semantic parser-failure, and semantic source-revision findings are addressed in 014de3948 and d788ad23f, with focused regressions. The staged proof-count finding is addressed in 9435f0778. The row-count suggestion is not applicable to current code: it compares `fetchone()` with the corresponding one-element tuple, so both operands have the same shape.

PR #2838, opened 2026-07-13
fix(protocol): split mutable summary head from immutable transcript segments

Sinity:
## Summary

Fixes the material-protocol append semantic bug (external protocol review, 2026-07-13) by redesigning revision layout into a per-revision mutable **head** (session/lineage/usage summary, re-encoded every revision, never byte-reused) and immutable **transcript** segments (message/block/attachment/session_event, the sole append-reuse surface, gated on canonical bytes). SEMANTICS_VERSION 1 → 2.

## Problem

`encode_appended_revision` decided prior-segment reuse by comparing record-id sequences only, then reused prior bytes wholesale. Stable record ids carry revision-mutable data (session `message_count`/`updated_at`/`title`/`tags`, usage aggregates, lineage status), so an appended revision **verified while contradicting itself** — reproduced against the pre-fix implementation:

```
verify_passed: true
encoded_session_message_count: 2
manifest_message_record_count: 4
session_segment_reused_byte_for_byte: true
```

Existing tests passed because they asserted byte reuse and anchor stability; the verifier checked digests/counts/anchors but no cross-record semantic closure. This was a protocol-design conflict: mutable revision summaries do not belong in segments whose contract is byte-for-byte reuse.

## Solution

- **Head/transcript split** (the reviewer's recommended design, chosen over reject-on-any-byte-change, which would have made append reuse never apply since the session record legitimately changes every revision): head = `head.ndjson` at reserved index -1 with its own seq space, always fresh; transcript keeps its own seq space from 0 and stays byte-stable. A side benefit: head growth (a new model's usage row, a new lineage edge) no longer breaks transcript appendability — under the old global-seq layout it shifted every transcript seq.
- **Byte-gated append**: every prior transcript record must re-hash to its manifest anchor `sha256`; a changed record with a stable id raises `NotAnAppendError` ("an edit is not an append").
- **Semantic-closure verifier laws** (`SemanticClosureError`): exactly one session record matching manifest `session_id`; `session.message_count` == actual message records; `message.block_count` == actual blocks per message; record kinds confined to their space.
- Manifest gains `head_segment`; content digest covers head + transcript; checked-in cross-repo fixture regenerated per its own policy; `docs/material-protocol-v1.md` updated.

## Verification

- `devtools test tests/unit/material_protocol/` → **42 passed**, including three new regressions: `test_append_keeps_mutable_summary_current_in_the_head` (pins the exact reported contradiction), `test_append_rejects_edited_prefix_even_with_stable_record_ids`, and `test_verifier_rejects_contradictory_session_message_count` (digests/anchors/counts all self-consistent; only the closure law catches it).
- Pre-fix reproduction confirmed the bug verbatim before any change was made.
- `devtools verify --quick` green (format, lint, strict mypy, render, topology, layering, schema roundtrip).

## Follow-ups

- Cross-language canonical-bytes parity (Python/Rust golden fixtures over floats, Unicode normalization, arbitrary keys) remains unproven — tracked as a Beads follow-up before content hashes become a cross-language authority contract.

Ref polylogue-303r.1


Sinity:
Codex P2 triage: **valid** — with the session record moved to the head (which the append prefix check deliberately never compares), an empty-transcript prior accepted a different session's material as an append. Fixed in the latest commit: `encode_appended_revision` now checks `session_id`/`origin`/`native_id` against the prior manifest before the transcript prefix check, raising `NotAnAppendError("session identity changed…")`. Regression: `test_append_rejects_different_session_identity_even_with_empty_transcript` (empty-transcript prior + different session id; removing the guard re-accepts it). Suite: 43 passed.

PR #2873, opened 2026-07-14
feat(sinex): durable publication-obligation ledger and transport contract

Sinity:
## Summary

Implements the Polylogue-side half of polylogue-303r.2 ("Publish Sinex
materials and anchored observations with durable retry"): a durable
`source.db` publication-obligation ledger, a transport contract modeled on
Sinex's documented `DurableEmissionReceipt`/`RawEnvelopeSettlement`
primitives, an orchestration service, and a real adapter from live archive
`Session` reads into the already-shipped `material_protocol.v1` encoder
(polylogue-303r.1). Defaults to `off` (zero transport work, zero obligation
writes) — standalone SQLite remains the permanent, canonical mode.

## Problem

Before this PR, no Python code implementing any Sinex producer existed in
Polylogue — `docs/sinex-interop.md`'s "low-volume bridge event" was aspirational
doc text with no implementation, confirmed by `grep -rli sinex polylogue/
--include='*.py'` turning up only `material_protocol` and doc references.
303r.2 needs the Polylogue side of a durable publish path: stage material
bytes, wait for a real receipt, and only then let a local projection treat
a revision as confirmed — with off/mirror/primary mode gating, crash-safety
between the local commit and the transport attempt, and idempotent retry.

## Solution

**`polylogue/sinex/`** (new package, 6 modules):
- `models.py` — `PublicationMode` (off/mirror/primary), `ObligationStatus`,
  and `ReceiptState`. `ReceiptState.unlocks_progress()` only returns true for
  `PersistedConfirmed`/`DurableDebt`/`SpoolAcceptedLossless`, mirroring
  sinex-r6d.11's documented rule that a bare in-memory/NATS-publish accept
  must never be mistaken for a durable commit.
- `obligations.py` — CRUD over `sinex_publication_obligations`, operating on
  a caller-supplied `sqlite3.Connection` so obligation creation composes
  into the same transaction as the revision it covers (design requirement).
- `transport.py` — the `SinexTransport` protocol, `NullTransport` (off mode:
  any call is a loud `TransportUsedInOffModeError`, never a silent no-op),
  and `LocalReferenceTransport` — a contract-faithful in-process double with
  injectable fault points for crash/rejection/debt scenarios.
- `service.py` — `PublicationService`: stage → attempt → mark, `lag()`/
  `pending()` for mirror-mode lag reporting, and an `on_confirmed` callback
  that fires only when a receipt actually unlocks progress.
- `material_adapter.py` — builds a real `SessionMaterial` from a live
  `Session` domain read using the real `Origin`/`Role`/`BlockType` enums and
  `native_id_from_session_id`; lineage/usage/session-events are a declared
  gap (a `FidelityGapInput` names the omission), not silently dropped.

**`polylogue/storage/sqlite/`**: additive migration 010 adds
`sinex_publication_obligations` (source.db v9 → v10), keyed by
`(object_id, protocol_version, revision_id, manifest_digest)` — that 4-tuple
is both the SQL primary key and the transport idempotency key.

**`polylogue/config.py`**: new `sinex_mode` config key (`[sinex] mode` in
`polylogue.toml`, env `POLYLOGUE_SINEX_MODE`), default `off`.

**Drive-by fix** (`polylogue/cli/commands/config.py`): `polylogue config
--format json` corrupted its own output whenever a config-inventory
description was long enough to exceed Rich's console width — `console.print`
without `soft_wrap=True` inserts a literal newline mid-line, turning a JSON
string into invalid JSON. `sinex_mode`'s description was the first entry
long enough to trigger it (confirmed via a minimal, isolated
`rich.console.Console` reproduction before and after the fix), but the bug
predates this PR and would have hit the next long description regardless.

### Rejected / deferred

Not implemented, per explicit upstream blocker (documented in
`docs/sinex-interop.md` and the package docstring): live Sinex JetStream
transport. As of this PR, `sinex-4j2.1.1` (the Sinex-side consumer for this
exact contract) has not merged, and `sinex-r6d.11` itself (the
`DurableEmissionReceipt` primitive the contract targets) is still open
upstream — there is no real endpoint to integrate against yet. This ships
wired to `LocalReferenceTransport` by default; pointing a real deployment at
live Sinex transport is cross-repo follow-up once those land. Full
end-to-end wiring into the live daemon ingest hot path (materializing every
changed session automatically) is also deferred — this PR delivers the
obligation/transport contract plus a real, callable staging path over live
archive data, not automatic background publication.

## Verification

```
devtools test tests/unit/sinex -> 24 passed
devtools test tests/unit/storage/test_durable_migrations.py -> 33 passed
devtools test tests/unit/cli/test_config_command.py -> 10 passed
devtools test tests/unit/core/test_config.py tests/unit/core/test_config_inventory.py -> 100 passed
mypy polylogue tests/unit/sinex -> Success: no issues found in 947 source files
ruff format --check / ruff check (touched files) -> clean
devtools verify --quick -> 15/15 steps ok, exit_code 0 (also re-ran clean via the pre-push hook)
devtools render topology-projection && devtools render topology-status -> new polylogue/sinex/* modules owner=stable
```

Not run: `devtools verify --all` / the heavy CI test suite (skipped per-PR
by design; runs post-merge on master per this repo's CI policy).

Ref polylogue-303r.2.


Sinity:
Fix round addressing the reviewer's major finding (unwired `sinex_mode` config key producing zero observable effect):

- `config.py`: `config_diagnostics()` now emits a `sinex_mode_not_yet_wired` warning when `sinex_mode` is `mirror`/`primary` (unconsumed by any code path today) and a `sinex_mode_unrecognized` error for typo'd values, surfaced through the already-reachable `polylogue config --format json` diagnostics array. `off` mode stays silent. Corrected the inventory entry's `reload_behavior` from the unverifiable `"daemon-loop"` to `"unwired"` and expanded its description to state the gap explicitly.
- `docs/sinex-interop.md`, `docs/architecture.md`, `polylogue/sinex/__init__.py`: corrected wording that implied `config.sinex_mode` is already consumed by a real (even reference-only) call site -- now state plainly that no production call site exists yet and point at the new diagnostic.
- `tests/unit/core/test_config_inventory.py`: 4 new tests proving the diagnostic fires for `mirror`, fires for an unrecognized value, and stays silent for `off`/default.

Deliberately **not** done: wiring an actual `PublicationService` construction into ingest/daemon/CLI. That's real production write-path work (staging obligations on every material revision), already scoped as follow-up under polylogue-303r.2's own notes ("full automatic wiring into the live daemon ingest hot path" listed as "Not attempted"), not something to improvise inside a fix round. Live Sinex transport remains separately blocked on unmerged upstream sinex-4j2.1.1 / sinex-r6d.11.

Verification:
- `devtools test tests/unit/core/test_config_inventory.py` -> 15 passed
- `devtools test tests/unit/sinex tests/unit/cli/test_config_command.py` -> 34 passed
- `devtools verify --quick` -> 15/15 steps, exit 0

Ref polylogue-303r.2

PR #2875, opened 2026-07-14
feat(security): local excision + secret-candidate detection (polylogue-27m)

Sinity:
## Summary

Implements the local-excision and secret-hygiene mechanics for polylogue-27m
("the archive can forget on purpose"): a candidate-only secret scanner, an
authoritative standalone/off-mode excision engine, and the mirror/primary
durable lifecycle-request/outbox mechanism exercised against a
fault-injecting versioned Sinex contract fake.

## Problem

Before this PR there was **no excision mechanism at all** (no schema, no
code) and **no candidate-only secret detector** for captured content.
`docs/plans/security-privacy-coverage.yaml` tracked this as an explicit
`major` severity gap (`security.captured-content-secret-detection`, owner
`security-privacy`, `bead: polylogue-27m`, `next_evidence: devtools test -k
secret_candidate`). `reset --session` tombstones a session in `user.db` and
deletes its `index.db` rows, but nothing prevented an ordinary re-ingest of
the *same unmodified source file* from silently resurrecting the deleted
content once `index.db` is rebuilt from `source.db` — there was no durable
removed-hash marker anywhere.

## Solution

**`polylogue/security/secret_scan.py`** — candidate-only detector.
Credential-shaped regex rules (AWS/GitHub/Slack/OpenAI/Anthropic key
shapes, PEM private-key headers, JWTs, an entropy-filtered generic
`key=value` rule) find spans in text. `scan_text_for_secret_candidates`
never returns, stores, or logs the matched literal — only a SHA-256
fingerprint, byte length, pattern id, and span offsets survive.
`record_secret_candidates` writes `AssertionKind.SECRET_CANDIDATE` rows via
the shared `upsert_assertion` chokepoint with `author_kind="detector"`,
which unconditionally coerces the row to a non-injectable `CANDIDATE`
status (the same invariant `PATHOLOGY`/`TRANSFORM_CANDIDATE` findings
already use — a detector can never self-promote a finding to
authoritative/injectable context).

**`polylogue/security/excision.py`** — standalone/off-mode excision,
authoritative. `apply_session_excision` removes a session's rows from
`embeddings.db` (message vectors), `index.db` (`sessions` cascades to
`messages`/`blocks`/`session_links`/FTS via `ON DELETE CASCADE` + triggers),
and `source.db` (`blob_refs` + `raw_sessions`, cascading to
`raw_session_memberships`/`raw_membership_census`), then records a durable
removed-hash marker per blob in the new `excised_content` table
(migration `011_excised_content.sql`, `SOURCE_SCHEMA_VERSION` 10→11 --
renumbered from the original 9→10 during a master-sync merge that resolved
a genuine collision with another PR's independently-landed `010_*`
migration; both additive migrations are kept, `sinex_publication_obligations`
at 010 and `excised_content` at 011) and
writes one `AssertionKind.EXCISION_RECORD` audit receipt to `user.db`
(reason, actor, removed hashes, per-tier counts). Content-bearing
assertions targeting the excised refs are removed too, so no dangling
annotation can quote deleted content. Blob *bytes* are never force-unlinked:
removing the referencing rows un-references the blob and the existing
reference-counted blob GC (polylogue-83u) reclaims it on its own next run
under its own lease discipline.

**Anti-resurrection gate**: `write_source_raw_session`
(`polylogue/storage/sqlite/archive_tiers/source_write.py`) is the single
acquire-time write choke point shared by the CLI `polylogue import` path
and the daemon `polylogued run` watch path (both funnel through
`pipeline.services.archive_ingest.parse_sources_archive` →
`ArchiveStore.write_raw_and_parsed_result` → `write_source_raw_session`). It
now checks `is_blob_hash_excised` before inserting and raises
`ContentExcisedError` on a match. `write_pair` in `archive_ingest.py`
catches this specifically and skips just that one file
(`ParseResult.excised_skips`) — a batch import is never aborted by an
excised file. No change was needed to the daemon's reprocess/materialize
path: it only ever operates on an *already-stored* `raw_sessions` row, so
once acquire refuses re-storage there is nothing left to reprocess.

**`polylogue/security/lifecycle.py`** — mirror/primary durable
lifecycle-request/outbox mechanics. A durable `AssertionKind.EXCISION_REQUEST`
row lives in `user.db` (never `ops.db`, so an `ops` reset cannot erase a
pending request). `drive_lifecycle_request` advances a `pending →
acknowledged → confirmed` (or `pending → rejected`, terminal) state
machine against any `ExcisionLifecycleContract`. `SinexContractFake` is a
**test-only** fault-injecting implementation (can drop N requests to
simulate network loss, or force a rejection). `primary_may_invalidate_locally`
/ `apply_primary_invalidation_if_confirmed` guarantee primary-mode local
invalidation happens **only** after a `confirmed` state — a rejected,
still-pending, or unknown request always returns `success=False` with an
explicit reason and never touches the archive. A "process restart" is
modeled in tests by constructing a **new** `SinexContractFake` instance and
re-driving the same durable request id: the outcome is identical because
the durable row, not client memory, is authoritative.

**`polylogue ops excise`** CLI command
(`--session ID --reason "..." [--dry-run] [--yes] [--mode
standalone|mirror|primary] [--json]`), following `reset`'s
dry-run/confirm/`--json` conventions (`MutationResultPayload`).

## Explicit non-goal (per the bead's own scope)

This PR does **not** claim a real Sinex purge, a clean Sinex rebuild,
disconnected-replica closure, or a backup-restore proof.
`polylogue ops excise --mode mirror|primary` only records the durable
local request today; nothing in this repository drives it against a real
Sinex confirmation yet. `SinexContractFake` is explicitly test-only.
Binding this mechanism to Sinex's real `privacy_invalidation_scope`, purge
authority, and non-resurrection proof across backups/replicas is tracked
separately as polylogue-303r.6, and a contract-fake-only test run must
never be read as satisfying that scope.

## Verification

```
devtools verify --quick
# exit_code: 0 (ruff format/check, mypy --strict over 1968 files,
# render all --check with zero "out of sync" lines, verify
# layering/topology/manifests/closure-matrix)

devtools test tests/unit/security/test_secret_scan.py \
  tests/unit/security/test_excision.py \
  tests/unit/security/test_excision_lifecycle.py \
  tests/unit/cli/test_excise.py
# 41 passed in 77.97s

devtools test tests/unit/storage/test_durable_migrations.py
# 33 passed in 272.14s (confirms migration 010 applies cleanly from
# v1/v2/v3/v7 fixtures; 4 pre-existing tests needed their hardcoded
# applied_versions tuples/fixture DDL updated for the new version)

devtools test tests/unit/security/test_no_secret_leak_in_logs.py
# 2 passed (secret-log-hygiene AST guard -- confirms the new modules
# never interpolate a secret-shaped variable into a log/print call)

python3 -m devtools lab policy schema-versioning
# "Schema evolution policy intact."
```

Anti-vacuity notes:
- `test_reingest_does_not_resurrect_excised_content` and
  `test_reingest_batch_skips_excised_file_without_aborting` exercise the
  real `write_source_raw_session`/`ContentExcisedError` gate through a
  real `parse_sources_archive` round trip built from the shared
  synthetic-corpus fixture generator (`SyntheticCorpus`/
  `build_default_corpus_specs`, the same machinery
  `tests/unit/pipeline/test_archive_ingest_commit_batching.py` uses) —
  not a hand-rolled JSONL literal. Removing the gate check, or reverting
  the `write_pair` skip-not-abort handling, makes these fail.
- `test_apply_removes_rows_from_every_tier` asserts zero rows remain in
  each of `sessions`/`messages`/`blocks`/`raw_sessions`/`blob_refs`/
  `message_embeddings`/`embedding_status` against real archive-tier
  schemas; commenting out any one tier's delete in `apply_session_excision`
  makes the corresponding assertion fail.
- `test_rejection_cannot_report_success` and
  `test_pending_and_acknowledged_cannot_report_success` assert
  `LifecycleInvalidationOutcome.success is False` with an explicit
  `reason` for every non-`confirmed` state; only
  `test_confirmed_request_actually_excises_the_local_replica` reaches
  `success=True`, and it verifies the session row is actually gone from
  `index.db` afterward.

Not run: the full non-integration suite (`devtools verify --all`) —
reserved for a final pre-merge diagnostic per repo convention; the
testmon-affected slice above plus `verify --quick` is the standard inner
loop.

## Fix round (2026-07-14): adversarial review findings, Ref polylogue-layg

An independent adversarial review found four real issues, held this PR off
merge. All four are addressed on this branch:

- **BLOCKER — second write chokepoint bypassed the excision gate.**
  `write_source_raw_session` checked `is_blob_hash_excised` before insert,
  but `write_source_raw_session_blob_ref` (the daemon's memory-bounded
  streaming write route for multi-GiB files, used via
  `sources.live.batch.LiveBatchProcessor._ingest_full_records_archive`) did
  not — a sibling real production write path could silently resurrect
  excised content. Fixed: the same gate now runs in both functions, and
  `_ingest_full_records_archive` catches `ContentExcisedError` and skips
  just that record (`_ArchiveFullWriteResult.excised_skips`), mirroring the
  CLI path's `write_pair` skip-not-abort handling. New regression coverage
  at both layers: `test_blob_ref_reingest_does_not_resurrect_excised_content`
  (direct function call) and
  `test_full_ingest_skips_durably_excised_content_without_aborting_batch`
  (full `LiveBatchProcessor` orchestration, anti-vacuity-checked by
  temporarily reverting the gate and confirming the test fails).

- **MAJOR — excision resolved only session_id-keyed rows, missing related
  rows.** Two concrete gaps, both closed:
  - *Lineage*: a prefix-sharing lineage parent (fork/resume/auto-compaction)
    could be excised without touching its dependents, silently breaking
    every dependent child's composed transcript read (a dangling
    `branch_point_message_id` with no bytes behind it).
    `apply_session_excision` now resolves the full transitive
    prefix-sharing lineage (`find_lineage_dependents`) and refuses by
    default (`LineageDependentsError`) unless `--cascade-lineage` is
    passed, in which case the whole lineage is excised together.
    `plan_session_excision`/`--dry-run` surfaces dependents up front.
  - *Attachments*: `blob_refs` groups every blob published under one raw
    ingestion by a shared `ref_id` (`ref_type IN ('raw_payload',
    'attachment', 'sidecar')`). The excised-content marker was previously
    recorded only for the raw payload's own hash, not an attachment's
    (which can differ) — so an attachment blob's `blob_refs` row was
    un-referenced and deleted, but not durably marked excised. Fixed:
    every distinct blob hash under the ref_id now gets its own marker.
    Verified separately (no code change needed, already correct):
    `attachment_refs.session_id`/`message_id` carry `ON DELETE CASCADE` to
    `sessions`/`messages`, so a content-hash-deduplicated attachment still
    referenced by a *different*, non-excised session is correctly left
    alone — only the excised session's own reference is removed.

- **MAJOR — `security-privacy-coverage.yaml` claimed
  `captured_content_secret_detection: implemented: true` without a
  production caller.** True: `secret_scan.py`'s regex/entropy rules and
  `record_secret_candidates` write path existed, but nothing in a running
  archive ever invoked them — only tests called the functions directly.
  Fixed by adding the missing production caller rather than downgrading the
  claim: `scan_session_for_secret_candidates` reads a session's real block
  text/tool-input from `index.db` and records real candidates, wired to a
  new `polylogue ops scan-secrets --session <id>` CLI command. The coverage
  doc now also states explicitly what `implemented: true` does *not* mean
  here: there is no bulk/archive-wide scan yet (an operator must invoke it
  per session id), tracked as a separate follow-up, polylogue-layg.1.

- **MINOR — one-shot `sqlite3.connect()` calls in `excision.py` omitted
  `PRAGMA busy_timeout`.** Fixed via `_connect_ro`/`_connect_rw` helpers
  using the shared `READ_DB_TIMEOUT`/`DB_TIMEOUT` constants (same one-shot
  pattern already used in `storage/blob_publication.py`); applied to the
  new `secret_scan.py` connections too.

Verification for this round:
```
devtools test tests/unit/security/test_secret_scan.py \
  tests/unit/security/test_excision.py \
  tests/unit/security/test_excision_lifecycle.py \
  tests/unit/cli/test_excise.py tests/unit/cli/test_scan_secrets.py
# 61 passed in 77.49s

devtools test tests/unit/sources/test_live_batch_support.py \
  -k test_full_ingest_skips_durably_excised_content_without_aborting_batch
# 1 passed; anti-vacuity-checked (temporarily reverted the write gate,
# confirmed the test fails: assert 0 == 1)

devtools test tests/unit/sources/test_live_batch_support.py  # full file
# 17 pre-existing failures, byte-identical test names/assertions reproduced
# against the unmodified baseline (git stash of only this file) -- order/
# parallelism-sensitive flakiness unrelated to this change, not introduced
# by it

devtools verify --quick
# exit_code: 0 (ruff format/check, mypy --strict, render all --check with
# zero "out of sync" lines, topology/layering/closure-matrix/manifests/
# ci-workflows/doc-commands/test-infra-currency/test-clock-hygiene/
# pytest-timeout-overrides/degrade-loudly all ok)
```

**Master-sync merge note.** Master had moved 9 commits since this branch
diverged, including another PR's own `010_*` source-tier migration
(`sinex_publication_obligations`) landing independently at
`SOURCE_SCHEMA_VERSION` 10 -- the exact version this PR also claimed for
`excised_content`. Resolved by renumbering this PR's migration to
`011_excised_content.sql` / `SOURCE_SCHEMA_VERSION` 11 (both tables kept,
`sinex_publication_obligations` first, `excised_content` second) and
updating the hardcoded `applied_versions` tuples in
`tests/unit/storage/test_durable_migrations.py`. Re-verified after the
merge: `devtools test tests/unit/storage/test_durable_migrations.py` (36
passed), `devtools verify --quick` (exit 0), `devtools render all --check`
(zero "out of sync" lines).

GitGuardian flags a "Generic High Entropy Secret" at
`tests/unit/security/test_secret_scan.py` (`"token: ghp_" + "a" * 40`, a
synthetic fixture matching the GitHub-token regex shape under test, not a
real credential; pre-existing from the base commit, out of this fix
round's scope). Triaged in a PR comment; not a required status check on
this repo.

## Follow-ups

- polylogue-303r.6 (already tracked, dependent of this bead): bind the
  mirror/primary local mechanics proven here to a real Sinex confirmation,
  purge, residual, rebuild, and backup non-resurrection proof.
- polylogue-layg.1 (new, filed this round): add a bulk/archive-wide
  secret-candidate scan (CLI `--all` or a daemon convergence stage) so an
  operator does not have to already know which session id to scan.

Ref polylogue-27m. Ref polylogue-layg.


Sinity:
GitGuardian false positive triage: the flagged "Generic High Entropy Secret" at `tests/unit/security/test_secret_scan.py` (originally line 51, commit 0665472a5) is `"token: ghp_" + "a" * 40` — a synthetic all-repeated-character string built specifically to match the GitHub-token *shape* the regex detector under test looks for. It is not, and cannot be, a real credential. This is a pre-existing test fixture from the base commit, not introduced by the fix-round commits on this branch; not modifying it as part of this fix round since it's out of the four findings' scope, but flagging the triage here so it isn't mistaken for an unaddressed real leak.

GitGuardian is not a required status check on this repo (no `required_status_checks` configured in branch protection), so it does not block merge.

PR #2884, opened 2026-07-14
refactor(paths): extract sibling_index_db helper, sweep 17 sites

Sinity:
## Summary

Extract the duplicated path derivation pattern for finding index.db siblings across daemon, CLI, storage, and source layers into a single reusable helper function. Consolidates divergent existence-check behaviors.

## Problem

Seven status/insight surfaces and ten other internal sites independently re-derive "the index.db next to this anchor database" with different fallback behaviors:
- convergence_stages.py:868 checks file existence + sessions table (strictest)
- fts_status.py:156 checks only file existence  
- provenance.py:202 doesn't check existence at all
- Plus 14 more variations

This duplication was the correctness trap: any status surface could disagree about whether the archive exists.

## Solution

New `polylogue.paths.sibling_index_db(anchor, *, require_exists: bool)` helper:
- Takes any database anchor path (source.db, embeddings.db, index.db itself, etc.)
- Returns the path to its sibling index.db in the same directory
- Optional `require_exists` parameter checks file existence

Sweep 17 sites across all layers to use the shared helper. All status surfaces now use identical derivation logic and will yield consistent verdicts.

## Verification

- `devtools verify --quick` (format, lint, mypy, render all) passes
- `devtools test -k "status or paths"` targeted test suite passes
- Specific test_paths_root_exports_only_directory_layout_symbols updated to include new export

Ref polylogue-a7xr.8.


Sinity:
Duplicate of #2894 (identical diff targeting the same bead, polylogue-a7xr.8; #2894 additionally includes the bd bookkeeping commit). Closing without merge.

PR #2885, opened 2026-07-14
feat(devtools): add read-only real-archive slice screening harness

Sinity:
## Summary

Adds `devtools demo real-slice-screen`, a read-only extraction + privacy
screening harness for candidate real-archive session slices, plus the
harness's own test coverage. This is the "harness/fixture machinery" half of
polylogue-212.11's real-archive-data extension; the actual candidate slice
and its privacy write-up are a separate, deliberately-unmerged artifact (see
Verification below) awaiting operator spot-check before any real content is
folded into `polylogue/scenarios/`.

## Problem

polylogue-212.11 ("one shared deterministic proof world for all flagship
demos") has already landed 37 declared constructs on a fully-synthetic
Incident 14:32 corpus (PRs #2662, #2674). Per explicit operator direction,
the shared proof world should also draw on a real, representative slice of
the operator's own archive (`/realm/db/polylogue`) rather than staying
synthetic-only — "the point is representative data, not full determinism."

That requires a privacy-vetting step. The bead's own instructions are
explicit that this step must **not** be autonomous: a candidate slice has to
be pulled, screened, and handed to the operator as a flagged, reviewable
artifact before it becomes a permanent shared fixture. This PR ships the
reusable machinery for that step; it does not itself promote any real
content into the product tree.

## Solution

- **`devtools/proof_world_real_slice.py`** — given one or more session ids
  (or a refs file) and a real archive root, opens the archive read-only via
  `Polylogue.get_session()` (the API's `ArchiveStore.open_existing(...,
  read_only=True)` default — never mutates the source archive), flattens
  each session's message text + structured block payloads to plain text, and
  runs a pattern screen:
  - **secret/credential shapes**: AWS access key ids, generic
    `key=value`/`key: value` credential assignments, bearer tokens,
    `BEGIN ... PRIVATE KEY` blocks, OpenAI-style `sk-...` keys, Slack
    tokens, SSH public keys, JWT-shaped strings.
  - **PII-adjacent shapes**: email addresses, `/home/<user>` paths, IPv4
    addresses — with a small allowlist for common test placeholders
    (`test@example.com`, `127.0.0.1`, `0.0.0.0`, ...) so the report isn't
    dominated by noise from ordinary test fixtures.
  - Any secret-pattern hit -> `flagged`; any non-allowlisted PII hit (with no
    secret hit) -> `review`; otherwise `clean`.
  - Writes `SCREENING_REPORT.md` (per-session verdict + truncated match
    samples), `manifest.json`, and per-session `transcripts/*.txt` to an
    arbitrary output directory. It never writes into `polylogue/scenarios/`
    or any other product fixture path — promoting a slice stays a distinct,
    explicit, human-gated step.
- Registered as `devtools demo real-slice-screen` in
  `devtools/command_catalog.py` (category `workspace`); `docs/devtools.md`
  regenerated via `devtools render devtools-reference`.
- `tests/unit/devtools/test_proof_world_real_slice.py`: pure-function
  coverage of the pattern scanner and verdict logic (including the
  allowlist), the text-flattening helper, report rendering, CLI argument
  handling, and an end-to-end read path against a `SessionBuilder`-seeded
  archive that plants a real AWS-key-shaped string and asserts it is both
  read back correctly and flagged.

### What this PR does *not* do

- It does not fold any real archive content into the shared demo corpus.
  I ran the harness against 5 candidate sessions from the operator's own
  archive (their own dev-work sessions on this very repository — flaky-test
  hunting, pipeline-idempotency test design, and two issue implementations)
  and wrote up the result (0 flagged, 3 "review" — all traced to test
  placeholders/loopback IPs/the operator's own home path in their own
  tool-output truncation messages, not third-party data — 2 fully clean) as
  a distinct, gitignored artifact under
  `.agent/scratch/real-slice-vetting-2026-07-14/` (`OPERATOR-SUMMARY.md`,
  `SCREENING_REPORT.md`, `manifest.json`, `transcripts/`). That directory is
  **not** part of this diff — `.agent/scratch/*` is gitignored — and is left
  in place for a quick operator spot-check per the task's explicit
  instruction that this step "should NOT be treated as fully autonomous."
- The remaining declared-but-unimplemented AC items from the bead notes
  (parser v1/v2 dual interpretation, typed cross-source 14:32 event hooks)
  are unaffected by this PR and remain open follow-up scope on
  polylogue-212.11 — see the bead's own notes history (PRs #2662/#2674/#2795).
- No raw provider JSON/JSONL bytes were extracted from the archive's blob
  store; the harness reads normalized message text via the `Polylogue` API.
  If/when the slice is approved, a follow-up should decide whether real
  parser-replay fidelity (raw bytes) or the normalized text is the right
  fixture-generator input — that decision is out of scope here.

## Verification

```
devtools test tests/unit/devtools/test_proof_world_real_slice.py tests/unit/devtools/test_command_catalog.py
  -> 19 passed in 37.41s

devtools verify --quick
  -> exit_code 0 (ruff format/check, mypy --strict, render all --check,
     topology/layering/closure-matrix/manifests/ci-workflows/doc-commands/
     test-infra-currency/test-clock-hygiene/pytest-timeout-overrides/
     degrade-loudly all ok)
```

Anti-vacuity: `test_screen_session_reads_real_archive_and_flags_planted_secret`
drives the real production path (`Polylogue.get_session` against a
`SessionBuilder`-seeded archive, not a mock/stub) and fails if `scan_text`'s
secret-pattern set is emptied, or if `screen_session` stops calling it.

Ref polylogue-212.11.

## Summary by CodeRabbit

- **New Features**
  - Added a read-only command to extract and privacy-screen candidate archive sessions.
  - Supports selecting sessions directly or from a references file.
  - Generates screening reports, structured results, and transcript files.
  - Detects credential-like content and personally identifying information, with redacted secret samples.

- **Documentation**
  - Added the new workspace command to the developer tools catalog.

- **Tests**
  - Added coverage for detection, redaction, verdicts, file handling, reporting, and command-line validation.

### DIFF
--- a/devtools/command_catalog.py
+++ b/devtools/command_catalog.py
@@ -456,6 +456,25 @@ COMMAND_SPECS: tuple[CommandSpec, ...] = (
         ),
     ),
     CommandSpec(
+        "demo real-slice-screen",
+        "workspace",
+        "Read-only extraction + privacy screening of a candidate real-archive session slice.",
+        "devtools.proof_world_real_slice",
+        use_when=(
+            "Assembling a candidate real-archive slice for the shared demo proof world "
+            "(polylogue-212.11): pulls sessions read-only via the Polylogue API, flattens them "
+            "to text, and screens for secret/credential and PII-adjacent patterns before any "
+            "operator decides to fold the slice into a shared fixture. Never mutates the source "
+            "archive and never writes into polylogue/scenarios/ on its own."
+        ),
+        examples=(
+            "devtools demo real-slice-screen --archive-root /realm/db/polylogue "
+            "--session claude-code-session:<id>:<agent> --out .agent/scratch/real-slice",
+            "devtools demo real-slice-screen --archive-root /realm/db/polylogue "
+            "--refs-file refs.txt --out .agent/scratch/real-slice",
+        ),
+    ),
+    CommandSpec(
         "workspace dev-loop",
         "workspace",
         "Preflight branch-local daemon, web-shell, and browser-capture development loops.",
--- a/devtools/proof_world_real_slice.py
+++ b/devtools/proof_world_real_slice.py
@@ -1,0 +1,317 @@
+"""Real-archive candidate-slice extraction and privacy screening.
+
+Support for polylogue-212.11 (shared deterministic proof world / Incident
+14:32): the deterministic demo corpus should eventually be extended with a
+*representative slice of real archive data*, not synthetic fixtures alone.
+That extension is deliberately a two-step, human-gated process:
+
+1. This tool runs **read-only** queries against a real Polylogue archive,
+   flattens each candidate session to plain text, and screens that text for
+   secrets/credentials and personal-information patterns. It writes a report
+   plus rendered transcripts to an arbitrary output directory. It never
+   mutates the source archive (``Polylogue.get_session`` opens the archive
+   tiers with ``read_only=True``) and never writes into the product fixture
+   tree (``polylogue/scenarios/``) on its own.
+2. An operator reviews the report and transcripts and decides, session by
+   session, whether the slice is safe to fold into the shared proof-world
+   corpus. Only after that explicit approval should the slice move into a
+   real fixture path.
+
+The screening pass is a best-effort heuristic layer, not a certification.
+"Clean" means "no configured pattern fired" — an operator still has to read
+the transcripts before promoting anything to a shared fixture.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import sys
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+# Patterns that indicate a live secret/credential shape. Kept intentionally
+# narrow (favor false negatives over drowning the report in noise) — this is
+# a triage aid, not a DLP product.
+_SECRET_PATTERNS: dict[str, re.Pattern[str]] = {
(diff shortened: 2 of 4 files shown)


Sinity:
Addressed both CodeRabbit findings from the a97b09da7 review pass in dfd019090:

1. **Repeated archive opens** — `screen_sessions()` now opens one `Polylogue(archive_root=...)` instance and reuses it across every session id via a new `_screen_session_with(poly, session_id)` helper. `screen_session()` keeps its existing two-arg signature (used directly by several tests and as the convenient single-session entry point) by opening its own scoped instance and delegating to the same helper. Added `test_screen_sessions_opens_the_archive_once_for_the_whole_batch`, which counts real `Polylogue.__aenter__` calls while driving the actual screening path — verified locally that it fails (2 opens instead of 1) if `screen_sessions` regresses to the old per-id-open loop.
2. **`Any` parameter on `_flatten_session_text`** — replaced with a `_SessionLike` Protocol (structural, read-only `@property` members). A plain-attribute Protocol was tried first but mypy's *invariant* attribute check rejected it (`Message.blocks: list[dict[str, object]]` vs. a bare `blocks: object` attribute); read-only properties make the check covariant, so both the real `Session`/`Message` domain models and the tests' duck-typed `_FakeSession`/`_FakeMessage` doubles satisfy it structurally without the tests needing to construct the real domain model.

Verification: `devtools test tests/unit/devtools/test_proof_world_real_slice.py` → 20 passed; `devtools verify --quick` → exit 0.

PR #2878, opened 2026-07-14
refactor: consolidate role synonym vocabulary (polylogue-a7xr.7)

Sinity:
## Summary

Consolidate role synonym vocabulary into a single source of truth to prevent divergence between Role.normalize and ROLE_SQL_VALUES. Also rename normalize_role in surfaces/payloads.py to role_label to avoid confusion with the canonicalizing version.

## Problem

Two separate places maintain role synonyms independently:
- `core/enums.py:127-137` — `Role.normalize` maps synonyms→canonical
- `archive/message/roles.py:18-24` — `ROLE_SQL_VALUES` holds the same sets inverted

When a synonym is added (e.g., "progress", "result"), both must be updated by hand or the other silently gets out of sync, causing --role filters to miss rows. Additionally, two unrelated functions named normalize_role create wrong-import failure modes.

## Solution

- Create `ROLE_SYNONYMS: dict[Role, frozenset[str]]` in `core/enums.py` as the single source of truth
- Update `Role.normalize` to iterate through `ROLE_SYNONYMS`
- Derive `ROLE_SQL_VALUES` in `archive/message/roles.py` from `ROLE_SYNONYMS` (~6 lines)
- Rename `normalize_role` in `surfaces/payloads.py` to `role_label` and update all call sites
- Add coupling test to verify every synonym round-trips through `Role.normalize` and `ROLE_SQL_VALUES` stays in sync

## Verification

- All 387 role-related tests pass (`devtools test -k role`)
- All 387 role-related tests pass (pytest-testmon affected selection)
- `devtools verify --quick` passes (format, lint, mypy, render all, layering)
- Coupling test verifies:
  - Every canonical role in `ROLE_SYNONYMS` has corresponding entry in `ROLE_SQL_VALUES`
  - Every synonym round-trips through `Role.normalize` to its canonical role
  - `ROLE_SQL_VALUES` values match `ROLE_SYNONYMS` as sorted tuples

Ref polylogue-a7xr.7.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Enhancements**
  * Improved recognition of equivalent role names from different providers.
  * Standardized role labels in session messages and archived message views.
  * Improved role-based message filtering by including recognized role alternatives.
* **Bug Fixes**
  * Unknown or empty roles now consistently display as “unknown.”
* **Tests**
  * Added coverage to verify role recognition and filtering remain synchronized.

PR #2886, opened 2026-07-14
refactor: consolidate five percentile implementations into core.stats (#2878)

Sinity:
## Summary
Consolidate duplicate percentile algorithms across daemon, insights, and archive subsystems into a single implementation in core/stats.py that supports both linear and nearest-rank methods.

## Problem
Five separate implementations existed (three using linear interpolation, two using nearest-rank) with different behaviors and scaling conventions (q vs p). This caused divergent p50/p95 results across operator-facing surfaces when displaying metrics side by side.

## Solution
Created `polylogue.core.stats.percentile()` with:
- Support for both 'linear' (default, numpy.percentile compatible) and 'nearest' methods
- Consistent quantile scale (q in [0, 1])
- Single source of truth for all percentile computations

Updated call sites to use the new function:
- daemon/status.py, cursor_lag_baseline.py, live_ingest_attempt_progress.py → linear method
- insights/portfolio.py → nearest method
- archive/semantic/timing.py → nearest method with p→q conversion
- insights/archive_rollups.py → nearest method

## Verification
- devtools verify --quick: all gates pass
- devtools test -k percentile: 1 passed
- No remaining duplicate _percentile implementations in codebase

Ref polylogue-a7xr.4


Sinity:
Duplicate of #2896 (same diff targeting the same bead, polylogue-a7xr.4, plus one extra test file — a resume-after-pause artifact that re-ran an already-completed Haiku item). Closing without merge.

PR #2887, opened 2026-07-14
refactor(run-projection): stop materializing cache tables

Sinity:
## Summary
Stop writing to session_runs, session_observed_events, and session_context_snapshots cache tables. These materialized rows are no longer needed since terminal read queries already synthesize the same data from sources via CTEs.

## Problem
Run-projection cache tables were designed as a materialization optimization when read patterns were expensive. After schema v16, archive_tiers/archive.py provided cheap source-derived CTEs for all three relations. The cache tables became redundant write overhead and complexity.

## Solution
- Remove cache table DDL from index.py canonical schema (v36)
- Stop calling replace_*_bulk_sync/async for cache tables in rebuild.py
- Update run_projection_reads.py to always use include_materialized=False
- Update demo constructs and corpus measurement to reflect source-derived-only model

Existing tests verify parity: run-projection reads use the same source data, just via CTEs instead of materialized rows.

## Verification
- devtools verify --quick passes (28.7s)
- Demo corpus seed roundtrip completes (construct count assertions)
- Mypy strict type checking passes

Ref polylogue-dab.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


Sinity:
Superseded by #2898 (same bead, polylogue-dab) — this branch's actual code diff against current master is now empty except bookkeeping (.beads/issues.jsonl, topology-target.yaml); #2898 carries the real 292-line refactor. Closing without merge.

PR #2915, opened 2026-07-15
fix(storage): make raw replay batches component-aware

Sinity:
## Summary

Makes bounded raw materialization schedule complete authority components rather than individual raw rows, classifies historical single-session snapshots into revision authority, isolates failures per selected component, and emits typed selected-attempt receipts. Ref `polylogue-hjpx`.

## Problem

A production-shaped reproduction with five revisions of one Codex session plus four independent raws produced nine candidates in five authority components. The old raw-row limit repeatedly selected several cheap rows from the same component; classification scanned them but yielded zero replayed logical sources, leaving the same executable backlog forever. The live July 15 preflight showed the same failure class at archive scale while the daemon processed two sources per pass and the candidate count grew.

## Solution

- Bind uncensused single-session historical raws as provisional full revisions, then use byte-prefix authority when possible and semantic membership authority for divergent full snapshots.
- Expand components transitively through source path, membership keys, and raw-revision logical keys, including newly discovered moved-path history.
- Apply the batch limit to complete components; blocked and executable work share the same budget.
- Rotate previously attempted retryable components behind never-attempted work using disposable daemon receipts.
- Execute independent components separately so a resource/runtime failure receives a retryable receipt without starving peers.
- Add stable selected-attempt IDs and typed outcomes to `RepairResult`, plus zero-work and nonzero daemon pass receipts.

This is the execution foundation, not closure of the full P0. Immutable pre-execution census and full before/after conservation remain `polylogue-hjpx.1`; the sanitized July-15-scale proof remains `polylogue-hjpx.2`. The umbrella stays open. No live archive mutation was authorized or performed.

## Acceptance matrix

- Reproduction and non-progress cause: satisfied.
- Complete-component batching and finite happy-path drain: satisfied for known components.
- Retry/resource/runtime isolation and fairness: satisfied for selected-attempt scheduling.
- Immutable full-census plan conservation, authority witnesses, durable rejected-stale blocker, and two-census fixed point: deferred to `polylogue-hjpx.1`.
- July-15-shape resource/convergence proof: deferred to `polylogue-hjpx.2`.
- Live apply gate: intentionally untouched; `polylogue-yla8` remains refused pending verified backup and explicit authorization.

## Verification

- `devtools test tests/unit/sources/test_revision_backfill.py -k 'incremental_target_expands_new_logical_key_across_source_paths or targeted_rebuild_expands_same_session_across_source_paths_only'` — 2 passed.
- `devtools test tests/unit/storage/test_repair.py -k 'isolates_failed_component or receipts_rotate_retryable_plan or batch_limit_counts_authority_components or blocks_aggregate_sub_limit_cohort'` — 4 passed.
- `devtools test -k raw_materialization` — 79 passed before the final per-component isolation refactor; the changed isolation path then passed the focused four-test command above.
- `devtools verify --quick` — all 16 steps passed; run `20260715T222101Z-quick-1942203-bab135b4`.
- Full `devtools verify` was not completed: a fresh-worktree testmon seed expanded to the full corpus, accumulated inherited failures, consumed ~17.5 GB peak memory plus ~9.3 GB swap and ~11 GB writes, then stalled at 94%. Tracked as `polylogue-b054.1.1`.
- `devtools test -k raw_authority` has one inherited clean-`origin/master` failure, tracked as `polylogue-lkrc.4`.

## Adversarial review notes

- Iteration 1 found cross-path component splitting, retry starvation, split blocked/executable budgets, and an invalid one-pass fixed-point claim. The first three were fixed with regressions; the fixed-point claim was removed and specified under `polylogue-hjpx.1`.
- Iteration 2 found that complete immutable census/conservation, durable rejection, authority witnesses, and scale proof still exceed this phase. Those are not claimed here and are captured in `polylogue-hjpx.1`/`.2`. It also found whole-batch exception starvation; this phase now isolates and receipts each selected component independently.

### DIFF
--- a/docs/plans/layering.yaml
+++ b/docs/plans/layering.yaml
@@ -31,7 +31,7 @@ writer_modules:
           interruption: atomic
       entrypoints:
         [apply_raw_membership_classification, apply_raw_revision_replay, classify_raw_revision_cohort,
-        delete_sessions, replace_raw_membership_census,
+        delete_sessions, release_provisional_full_revisions, replace_raw_membership_census,
         write_parsed_for_retained_raw, write_parsed_for_retained_raw_result, write_raw_and_parsed,
         write_raw_and_parsed_result, write_raw_blob_and_parsed, write_raw_blob_and_parsed_result]
       twin_write_contract: raw-membership-classification
--- a/polylogue/daemon/cli.py
+++ b/polylogue/daemon/cli.py
@@ -630,9 +630,29 @@ def _drain_raw_materialization_once(*, limit: int = _RAW_MATERIALIZATION_CONVERG
         result = repair_raw_materialization(config, dry_run=False, raw_artifact_limit=limit)
     finally:
         _close_raw_materialization_fts(config.archive_root / "index.db")
+    _emit_raw_materialization_pass(result)
     if not result.success:
         logger.warning("raw materialization: bounded convergence incomplete: %s", result.detail)
     return result.repaired_count
+
+
+def _emit_raw_materialization_pass(result: Any) -> None:
+    """Persist the conserved plan outcomes for one bounded daemon pass."""
+    outcomes = tuple(getattr(result, "plan_outcomes", ()))
+    from polylogue.daemon.events import emit_daemon_event
+
+    metrics = dict(getattr(result, "metrics", {}))
+    emit_daemon_event(
+        "raw_materialization_pass",
+        payload={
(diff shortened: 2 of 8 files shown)


Sinity:
Post-open final verification on commit ff6c04fc2: `devtools test -k raw_materialization` selected 80 tests and passed all 80 in 106.14s (15,989 deselected). This includes the final per-component fault-isolation refactor.

PR #2966, opened 2026-07-17
fix(storage): stream safe raw authority replay

Sinity:
## Summary

Streams admitted retained JSONL revisions through raw-authority census and replay while preserving byte-prefix authority without loading full blobs into memory.

## Problem

The replay path used `read_all()` for record streams despite blob/parser streaming support. It inflated peak RSS beneath the existing component resource envelope. Ref polylogue-hjpx.2.

## Solution

- Add chunked full-snapshot prefix proof and streamed full/append fold comparison.
- Add retained-raw descriptor/open APIs and route admitted record-stream parsing through them.
- Preserve the existing component-size admission gate: streaming bytes alone does not bound Codex parsed-session construction.
- Keep non-stream/bundle formats fail-closed behind the same envelope.
- Add anti-vacuity tests proving no eager blob read remains on the admitted stream route and streamed/eager classification agrees for both chain and ambiguous cohorts.

## Verification

- `devtools test tests/unit/storage/test_raw_revision_authority.py tests/unit/sources/test_revision_backfill.py tests/unit/storage/test_repair.py` — 82 passed.
- Pre-push `devtools verify --quick` — all 15 static/generated gates passed.

## Adversarial review notes

Iteration 1 found a real gap: the first revision admitted arbitrary oversized streams even though parser output is still materialized. Commit `8be68f9df` restores the hard admission limit; production-shape parser checkpointing remains explicitly owned by polylogue-hjpx.2.

## What I decided not to change

Non-stream exports/bundles and oversized record streams remain resource-blocked. They need parser-output checkpointing plus a measured production-shape envelope, not a weaker admission rule. Authority components remain atomic because partial execution would invalidate immutable plan and conservation receipts.

### DIFF
--- a/polylogue/archive/revision_authority.py
+++ b/polylogue/archive/revision_authority.py
@@ -2,10 +2,11 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from dataclasses import dataclass
 from enum import StrEnum
 from hashlib import sha256
-from typing import Literal
+from typing import BinaryIO, Literal
 
 
 class RawRevisionKind(StrEnum):
@@ -64,6 +65,15 @@ class RawRevisionEnvelope:
 class HistoricalRawRevision:
     raw_id: str
     payload: bytes
+
+
+@dataclass(frozen=True)
+class HistoricalRawRevisionStream:
+    """A retained full revision whose bytes can be compared without loading it."""
+
+    raw_id: str
+    payload_size: int
+    open_payload: Callable[[], BinaryIO]
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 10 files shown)


Sinity:
Review triage after the follow-up commits:

- The parser-output admission finding is valid and addressed by `8be68f9df`: the hard component-size gate remains for all raw shapes. Counting only non-stream bytes would reintroduce the unbounded ParsedSession/spill cohort path, so that recommendation is intentionally not adopted. Parser checkpointing plus a measured production-shape envelope remains `polylogue-hjpx.2` scope.
- The quadratic streamed-prefix scan is addressed by `484bcbc68`: it sorts by measured stream size, rejects duplicate-size cohorts, and checks only adjacent strict-prefix pairs, which is sufficient and necessary for a unique proper-prefix chain.
- The same follow-up centralizes retained stream parsing and covers historical backfill plus live append and full replay. Their anti-vacuity tests make `ArchiveBlobPublisher.read_all` raise, so a reintroduced eager route fails.

PR #2969, opened 2026-07-17
feat(devtools): guard and meter raw authority proofs

Sinity:
## Summary

Make the raw-authority scale scenario safe to schedule on a shared host and make its resource evidence usable for the July-shape fixed-point gate.

## Problem

The first cardinality run used the one-row archive writer for every synthetic raw. That writer correctly flushes and commits ordinary acquisitions, but at 15,264 rows it turned scenario generation into an I/O storm before authority replay could be measured. The original receipt also reported only `ru_maxrss`, which cannot distinguish process memory, swap, or the workload’s own I/O.

## Solution

The runner now uses the source writer’s documented bulk transaction contract: it writes each payload to the existing blob publisher, inserts source rows in one transaction with `manage_transaction=False`, then flushes once. It samples kernel-owned process RSS/PSS/swap, CPU, and read/write I/O counters around each replay pass and includes measured values in its receipt. It refuses to start when host I/O or memory full-stall pressure is above a conservative threshold; `--allow-contended-host` is an explicit operator override for intentionally contended experiments.

Replay semantics and the production archive writer remain unchanged.

## Verification

- `devtools test tests/unit/devtools/test_raw_authority_scale_proof.py` — `2 passed`
- `devtools verify --quick` — format, lint, strict mypy, generated-surface, topology, layering, schema, and policy gates passed.

## Remaining evidence

The full July-shape run remains deliberately pending until host I/O pressure falls below admission. This PR supplies the safe, measurable runner; it does not claim the full-scale result.

### DIFF
--- a/devtools/command_catalog.py
+++ b/devtools/command_catalog.py
@@ -557,6 +557,20 @@ COMMAND_SPECS: tuple[CommandSpec, ...] = (
         ),
     ),
     CommandSpec(
+        "workspace raw-authority-scale-proof",
+        "workspace",
+        "Run bounded raw-authority replay to a two-census fixed point.",
+        "devtools.raw_authority_scale_proof",
+        use_when=(
+            "Generate a receipt-backed raw-authority scenario before a live replay gate. "
+            "Use explicit July-15-shaped component/raw arguments for the contained scale run."
+        ),
+        examples=(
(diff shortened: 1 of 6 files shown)


Sinity:
Superseded by merged PR #2970; the runner changes are now on master.

PR #3039, opened 2026-07-17
feat(campaign): reconcile result projections (#3039)

Sinity:
## Summary

Make external-agent result indexes a receipt-derived projection rather than a second, manually editable outcome ledger.

## Problem

PR #3028 updated immutable analysis receipts but the index retained stale state values. A later manual patch was needed before state-oriented intake queries became truthful. The prior intake command also appended directly to the index.

## Solution

- add a generic reconciler for every workload, including explicit attempt and package-revision receipt forms;
- reject missing, duplicate, stale, or second-outcome projections;
- atomically replace each validated index only, never an immutable receipt;
- make package triage preserve raw custody and write an immutable attempt receipt before materializing the index;
- rebuild all current wave indexes to the single state field.

## Verification

- Focused campaign receipt reconciliation: 4 passed; exercises live projection, index-state mutation, receipt mutation, duplicate identity, and end-to-end triage publication.
- Quick verification gate: passed.
- Campaign reconciliation check: verified all workload projections.

Ref polylogue-yyvg.6.

### DIFF
--- a/.agent/handoffs/external-agent-campaigns/2026-07-16-gpt-pro-wave/README.md
+++ b/.agent/handoffs/external-agent-campaigns/2026-07-16-gpt-pro-wave/README.md
@@ -28,11 +28,16 @@ Launch gating and result intake are mechanical, not prose:
   foundation merges, write `foundation-receipt.json`
   (`{"merged_commit": "<sha>", "verified_at": "<iso8601>"}`) at this
   directory's root.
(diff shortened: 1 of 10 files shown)


Sinity:
Automated review triage: CodeRabbit reported a provider-side review quota limit and produced no actionable finding. Local verification remains: focused receipt reconciliation suite (4 passed), quick gate, and a live all-workload reconciliation check.

Issue #1252, opened 2026-05-17
refactor(schema): first-class attachment native identifiers (#864 slice B)


Sinity:
## Problem

Per #864: attachment indexes read `id`, `provider_id`, `fileId`, `driveId` out of `provider_meta`. JSON-expression indexes are fragile.

## Scope

- [ ] First-class identifier columns on `attachments` / `attachment_refs` OR `attachment_provider_ids` lookup table
- [ ] Remove JSON-expression indexes (see #817 slice A overlap)
- [ ] Backfill from provider_meta

## Acceptance

- [ ] Lookup queries use stored columns; no JSON-extract on hot path
- [ ] Test parity for chatgpt + claude-code + codex + gemini attachments

Ref #864 #817.

## Ambitious expansion

Beyond first-class identifier columns, **also index by `provider + upload_origin`** (drive / paste / URL / OAuth-discovered).

The attachment library UI (#1199) needs to group by origin without scanning `provider_meta`. The lookup table is the natural place to land this; adding the origin column now costs one column + one index, unblocks #1199 cleanly.

Sinity:
## Ambitious expansion

Beyond first-class identifier columns, **also index by `provider + upload_origin`** (drive vs paste vs URL vs OAuth-discovered).

The attachment library UI (#1199) needs to group attachments by origin without scanning `provider_meta`. The lookup table is the natural place to land this; adding the origin column now costs one column + one index, costs ~5% extra work, unblocks #1199 cleanly.

PR #3073, opened 2026-07-18
feat(sources): acquire live Drive attachment bytes into the blob store

Sinity:
## Summary
Implements the Drive sub-case of `polylogue-83u.2`: live Drive-hosted attachment references (`driveDocument(s)`/`driveImage`/`driveAudio`/`driveVideo`) are now fetched during ingest and land in the content-addressed blob store with a true SHA-256, instead of never being acquired.

## Problem
A 2026-07-08 investigation (recorded on the bead) found the only code that ever called `DriveSourceClient.download_bytes` for an attachment — `_apply_drive_attachments` + `iter_drive_sessions` — had **zero live callers**. The real ingest path (`iter_drive_raw_data` → generic dispatch parse) never wired attachment acquisition in at all. Even if that dead code had been revived, it wrote bytes to `attachment.path`/local disk rather than `ParsedAttachment.inline_bytes`, so it would not have reached the existing true-hash blob-publish path anyway (schema v13's `acquisition_status`).

## Solution
`iter_drive_raw_data` (`polylogue/sources/drive/__init__.py`) already owns the one live `DriveSourceAPI` client instance per file, inside the iterator scope, before the acquire and parse stages decouple for memory-bounded streaming. A new module, `polylogue/sources/drive/attachment_fetch.py`, walks the raw session JSON for Drive-hosted doc/media references **at that exact point**, downloads each via the same live client, and injects the fetched bytes as base64 under a sidecar key (`DRIVE_LIVE_FETCH_DATA_KEY`, defined in `polylogue/sources/parsers/drive_support_attachments.py`). The existing chunk parser (`attachment_from_doc`) decodes that sidecar into `ParsedAttachment.inline_bytes` — no other parser changes — so the already-shipped publish-then-write path (`ingest_batch/_core.py`) picks it up as a real content-addressed blob with `acquisition_status='acquired'`, with zero downstream changes needed.

Fetch failures and oversize files (50MB cap) leave the attachment honestly `upload_origin="drive"` with no `inline_bytes` — never a fabricated hash.

Removed `_apply_drive_attachments`/`iter_drive_sessions` (and the now-dead `sources/__init__.py` re-export) — the obsolete, zero-caller path this change replaces, per surgical-renewal doctrine.

## Non-goals (explicitly out of scope)
`polylogue-83u.2` also named export-zip-member and local-path attachment sub-cases. Both were re-confirmed to have **zero current producer** — no parser has ever emitted an attachment whose bytes live as a sibling zip member or a real local filesystem path. There is nothing live to un-bypass for either, so this PR does not touch them; noted on parent `polylogue-83u` rather than filing speculative follow-up beads with no defined producer.

## Verification
- `devtools test tests/unit/sources/test_drive_ops.py tests/unit/sources/test_drive_attachment_fetch.py tests/unit/storage/test_attachment_acquisition.py tests/unit/sources/test_source_laws.py` → **143 passed**
- `devtools verify --quick` → **exit 0** (ruff format/check, mypy --strict, render all --check, topology, layering, closure-matrix, manifests, ci-workflows, doc-commands, docs-coverage, test-infra-currency, test-clock-hygiene, pytest-timeout-overrides, degrade-loudly)
- `devtools test tests/unit/sources/` → **1705 passed**, 2 pre-existing failures in `test_chatgpt_normalization_survivors.py` confirmed unrelated (reproduce identically with this change stashed out)

Ref polylogue-83u.2


Sinity:
Addressed all 3 Codex findings in 74da2f354:

- **P1** (sidecar leaking into block metadata): `_drive_media_content_block` now strips `DRIVE_LIVE_FETCH_DATA_KEY` before building `ContentBlock.raw`, mirroring `_without_inline_bytes`'s existing treatment of `inlineData`/`inlineFile`/`inlineImage`. Covered by a new end-to-end test (`test_live_fetched_drive_attachment_bytes_reach_inline_bytes_not_block_metadata`).
- **P2** (fileId vs id precedence): `_doc_file_id` now tries `fileId` before `id` (matching `attachment_from_doc`'s own #1252 precedence) and never falls back to `driveId` (a shared-drive container id, not a file id). Covered by two new tests.
- **P2** (cached payloads bypass the injector): the injector now runs on every read (cache hit or fresh download), refreshing the cache file only when something new was actually resolved. Covered by a new test seeding a stale pre-injector cache file and proving the next run backfills it.

200 focused tests pass (`test_drive_ops.py`, `test_drive_attachment_fetch.py`, `test_parsers_drive.py`, `test_attachment_acquisition.py`, `test_source_laws.py`, `test_gemini_chunked_prompt_contract.py`), `devtools verify --quick` + `public-claims` green.

Re: `ci/circleci: quick-gate` failing — confirmed pre-existing/unrelated: it also fails on `4053787ab` (master itself, PR #3064, already merged), consistently ~18s in, well before any real lint/mypy/render step would complete. Not something this PR introduced.

PR #3074, opened 2026-07-18
feat: land WebUI v2 scaffold, design system, and generated client (3/8 verticals)

Sinity:
## Summary

Lands three of the eight planned WebUI v2 verticals (`polylogue-bby.11`) as one coherent, hand-reconciled stack: the TypeScript+Preact+Vite workspace scaffold with a daemon SSR strangler mount (`webui-01`), the generated design system + accessibility/verification harness (`webui-07`), and the generated OpenAPI client (`webui-08`). This is not "WebUI v2 complete" — it is foundation plus two adjacent verticals; list, reader, search, transcript, insights, and cost views are not part of this PR.

## Problem

`polylogue-bby.11` ratified TypeScript+Preact+Vite, SSR-first islands, generated design tokens, and a generated OpenAPI client as the WebUI v2 architecture. Three externally generated (GPT Pro) implementation packets existed for this cluster (`webui-01-workspace-scaffold-r01`, `webui-07-design-system-a11y-r01`, `webui-08-client-contracts-r01`), each authored independently against the same snapshot commit (`536a53e`, 13 commits behind current master) — unreviewed, and each unaware of the other two.

## Solution

**webui-01 (scaffold)** — `polylogue/daemon/webui.py` (manifest-governed Vite asset bundle, SHA-256 ETag/immutable caching, semantic SSR for a bounded `messages where words >= 0 | sort by time desc` overview); `GET /app` + `GET /app/assets/:asset` in `http.py` under the existing shell bootstrap admission policy; `/api/query-units` learns to replay an opaque `QueryTransaction` continuation. Committed, content-hashed `polylogue/daemon/static/dist` ships in the wheel/sdist. No existing legacy-shell route or module touched.

**webui-07 (design system)** — a generated Preact component kit (layout primitives, data table + continuation paging, honest empty/unknown/degraded/loading states, facets, transcript/code/diff blocks, timeline/sparkline, theme toggle) driven by Python-owned tokens (`polylogue/ui/theme.py` → `devtools render webui-design-system` → `tokens.css`/`contracts.ts`/`contrast-report.json`), a deterministic five-vertical SSR fixture, Vitest component tests, and Playwright keyboard/no-JS/axe/visual-regression coverage. No production route is mounted by this vertical.

**webui-08 (client contracts)** — `devtools render webui-client` derives a committed, deterministic TypeScript client (`webui/src/api/generated.ts` + `runtime.ts`) from `docs/openapi/search.yaml`, covering all 8 currently-declared daemon operations including `search()`/`query()` async page iterators. Zero new runtime dependencies.

**Reconciliation performed by hand** (all three packets collided on shared `webui/` workspace files since none was aware of the others):
- `webui/package.json`/`package-lock.json`: merged scripts additively; renamed each packet's own colliding `build`/`test`/`test:unit` script where it meant something different from the established name (webui-01's `build` must keep emitting the exact bytes the daemon serves; webui-07's fixture build became `build:design-system`; webui-08's Node-test-runner lane became `test:client-contracts`); unified devDependency versions to the newest proposal and regenerated the lockfile via `npm install` rather than hand-patching it.
- `webui/tsconfig.json`: unioned all three authors' compiler-strictness flags into one config (kept `exactOptionalPropertyTypes`, added `useUnknownInCatchVariables`/`moduleDetection`), which surfaced two real strictness violations the individual packets never saw against each other's flags — fixed in `playwright.config.ts` (conditional `launchOptions` spread instead of an `undefined` assignment) and `webui/src/api/runtime.ts` (`TransportError.cause` needed an explicit `override` modifier).
- `webui/tsconfig.client.json` (new): webui-08's own tsconfig used `rootDir: "src"`, incompatible with the merged config's broader `include`; added a narrow config scoped to `src/api/**/*.ts` so its `tsc`-emit-and-`node --test` lane still produces output at the path its own test file imports.
- `webui/src/test/setup.ts`: merged webui-01's jest-dom import (needed by its own test) with webui-07's theme/localStorage/matchMedia reset and Preact unmount cleanup.
- `webui/vitest.config.ts` (from webui-07) is now the sole Vitest config; removed the now-dead `test` block from webui-01's `vite.config.ts`.
- `webui/playwright.config.ts`: webui-07's patch silently deleted this operator's local Chromium-executable discovery fallback (`existsSync` search over `/etc/profiles/per-user/sinity/bin/google-chrome` etc.) while adding legitimate deterministic-viewport/locale/timezone settings — restored the fallback.
- Re-baselined the two Playwright visual-regression snapshots (`evidence-{light,dark}.png`) against this host's actual font rendering (the packet's committed baselines showed a ~2% pixel diff from a different container's font stack); reviewed both images before committing.
- `devtools/command_catalog.py`/`generated_surfaces.py`/`docs/devtools.md`: webui-07's and webui-08's new `render` command registrations collided at the same insertion point; both now registered in sequence, docs regenerated rather than hand-patched.

**Found and fixed a real correctness bug**, not just merged the packets verbatim: webui-01's already-integrated `/api/query-units` continuation handling didn't reject malformed base64/UTF-8 tokens cleanly (`QueryContinuation.decode()` was missing `binascii.Error`/`UnicodeDecodeError` from its caught-exception set, so a bad token from a browser or attacker raised an unhandled exception instead of a 400). webui-08's own patch contained a fix for this same class of issue, but its own validation had a **stale literal** (`projection != "default"`) that would have 400'd every real continuation, since production actually sets `projection="terminal-unit-envelope"` for this route as of #3051 (landed after both packets' shared snapshot). Adopted webui-08's more rigorous validation shape but corrected the literal against live source rather than either packet's own say-so.

## What's real vs. what's next

This is 3 of 8 declared verticals (`polylogue-bby.11` design field: scaffold, session list, search, transcript, insights/status, cost, design system, client contracts). Session list, search, transcript, insights/status, and cost views are **not** in this PR — they are separate, not-yet-started verticals per the bead's own design. The legacy `web_shell.py` reader remains fully in place and untouched; nothing is retired.

## Verification

- `cd webui && npm ci && npm run check` (generate:check + check:client + lint + typecheck + test [15 Vitest tests] + test:client-contracts [8 Node tests] + build + build:design-system) → all green
- `git diff --exit-code -- polylogue/daemon/static/dist` → clean (production build byte-identical to committed dist)
- `npx playwright test` (13 tests: 10 design-system + 3 first-party-auth) → 13/13 passed on rerun; one first-run failure in the pre-existing, untouched credential-TTL test was a timing race unrelated to this change
- `npm audit --audit-level=low` → 0 vulnerabilities
- `devtools test tests/unit/daemon/test_route_contracts.py tests/unit/daemon/test_web_reader.py tests/unit/archive/query/test_transaction.py tests/unit/devtools/test_render_webui_design_system.py tests/unit/devtools/test_render_webui_client.py tests/unit/devtools/test_render_openapi.py tests/unit/devtools/test_generated_surfaces.py tests/unit/devtools/test_command_catalog.py` → 281 passed, 1 failed (`TestReaderAssertionEndpoint::test_operational_web_payloads_redact_configured_archive_paths`, a symlinked-archive-root sqlite-open failure reproduced identically on unmodified master before any of these changes — pre-existing/environment-specific, unrelated file)
- `devtools render openapi / webui-design-system / webui-client / topology-projection / topology-status` (all `--check`), then `devtools render all --check` → all surfaces sync OK, zero drift
- `uv run ruff format --check` / `ruff check` / `mypy --follow-imports=skip` on all touched Python → clean (a handful of files needed reformatting under this repo's pinned ruff version vs. the packets' own; fixed, re-verified generators still byte-identical afterward)
- `devtools verify --quick` → exit_code 0, all 16 steps green (required fixing one real gap: `GET /app/assets/:asset` had no documentation footprint under `docs/`, added a real paragraph to `docs/architecture.md` rather than baseline-suppressing it)
- `devtools verify ci-workflows` → 17 workflow files checked, no errors

Not run: Nix build/eval (the flake.nix `postFixup` manifest assertion is source-reviewed only), full `devtools verify --all`, manual assistive-technology review beyond the automated axe scans.

Ref polylogue-bby.11


Sinity:
## CI status note: \`ci/circleci: quick-gate\` failure is pre-existing, unrelated to this PR

Investigated the CircleCI \`quick-gate\` failure before merging. It is **not** caused by this PR's diff:

- Master's own tip (before this PR's rebase target) fails the identical check with the identical 3 mypy findings:
  - \`polylogue/hooks/__init__.py:765\` (redundant-cast)
  - \`polylogue/archive/query/unit_results.py:297\` (redundant-cast)
  - \`polylogue/daemon/http.py\` (type-var, an unrelated socket-peek helper ~140 lines from anything this PR touches)
- None of these three files/lines are touched by webui-01/07/08.
- Root cause: this devshell's Nix-provided \`mypy\` (1.20.1, injected onto \`PYTHONPATH\` ahead of the venv) silently shadows the \`uv.lock\`-pinned \`mypy\` (2.3.0, from #2863) that CircleCI's bare \`uv sync --extra dev --frozen\` actually installs and runs. \`env -u PYTHONPATH .venv/bin/mypy polylogue/\` reproduces CircleCI's exact 3 findings locally; the devshell's shadowed \`mypy\` reports clean.
- Branch protection currently has no \`required_status_checks\` configured (verified via \`gh api repos/Sinity/polylogue/branches/master/protection\` — no such key present), so this is not a gating check.

Filed \`polylogue-i0ep\` to fix the 3 pre-existing mypy-2.3.0 findings and the devshell/lockfile mypy-version drift that's hiding them from local \`devtools verify\`. Not fixing them here to keep this PR's diff scoped to the three webui verticals it actually implements.

All of this PR's own verification (\`devtools verify --quick\` full 16-step gate, focused pytest, \`npm run check\`, Playwright e2e, generated-surface \`--check\` for every affected surface) is green — see the PR body's Verification section.

PR #2928, opened 2026-07-16
feat(browser): expose autonomous conversation transport (#2928)

Sinity:
## Summary

Replace the private GPT Pro launch/work-package implementation with a provider-neutral browser action conduit, and make canonical conversation capture converge automatically from provider evidence even after visible tabs close. The extension popup now treats ordinary pages as non-conversations and removes capture/status/open-tab/receiver hand cranks.

## Problem

The browser product had absorbed one operator-specific campaign: Sol prompts, work packages, cadence, handoff validation, and launch queues lived inside generic receiver/extension code. At the same time, ordinary capture could archive an early snapshot and never reacquire later assistant turns or files; status and recovery mechanics were exposed as user buttons. Ordinary webpages could also inherit stale conversation fidelity and appear as failed unsupported conversations.

## Solution

- Add a receiver-authoritative BrowserAction contract with versioned capabilities, idempotency, hash-pinned attachments, leases, durable submit-intent quarantine, typed failure receipts, exact provider receipts, and reconciliation.
- Execute ChatGPT actions through one extension-owned inactive first-party transport; never borrow or activate operator tabs. Exact supported selection is Chat / GPT-5.6 Sol / Pro, with fail-closed capability checks.
- Remove BrowserLaunchJob, BrowserPostCommand, Sol prompt, work-package, handoff, cadence, and campaign UI/routes from product code. External orchestration consumes generic actions plus ordinary Polylogue capture.
- Add a durable freshness queue fed by ChatGPT native-detail, service-worker, and DOM hints plus a rotating 15-minute inventory delta sweep. Closed conversations are captured through the same owned inactive transport.
- Automatically observe open conversations on install/startup and focused conversations on activation/completion. Missing sessions enter automatic catch-up.
- Make ordinary webpages and provider landing/project pages neutral non-conversations. Hide conversation-only fidelity/turn/assets/timeline controls there and remove Capture page, Check status, Sync open tabs, and Check receiver buttons.

## Acceptance / residual matrix

| Bead | Delivered here | Explicit residual |
| --- | --- | --- |
| polylogue-ptx | Generic action store/routes/CLI, owned inactive adapter, create + attachment + outcome-unknown live proofs, independent canonical capture | Live reply proof and complete packaged project/failure matrix remain |
| polylogue-3v1 | Automatic missing capture, bounded retry spool, durable freshness convergence, live closed-tab recapture, shared popup/ambient status | Gemini live bridge and queryable ops.db capture-gap events remain |
[...]

PR #3075, opened 2026-07-18
feat(browser-extension): fetch chatgpt-dom-v1 attachment bytes via the page bridge

Sinity:
## Summary
Closes the last remaining gap in `polylogue-83u.3`: the `chatgpt-dom-v1` DOM-scrape capture fallback now fetches attachment bytes through the same authenticated page-bridge mechanism the native adapter already uses, instead of recording only the chip name with `byte_count=0`.

## Problem
General ChatGPT attachment fetching already ships: `chatgpt_bridge.js` (`fetchAssetBytes`) and `chatgpt.js` (`acquireAssets`/`collectAssetDescriptors`) live-fetch sandbox/file-service bytes for the normal `chatgpt-native-v1` capture path (landed PR #2669, hardened #2672/#2712/#2926). The only remaining gap is the older `chatgpt-dom-v1` fallback — used only when the authenticated backend-api conversation read fails — which had **no acquisition path at all**: `collectAttachments` recorded just the chip's name.

## Architecture constraint
`chatgpt-dom-v1` has no backend-api `mapping` to resolve a file/sandbox id from (unlike `chatgpt-native-v1`); the DOM chip's own `href`/`src` is the only evidence available, and it is sometimes already a concrete `https://` URL the page itself rendered (e.g. an inline ``).

## Solution
Added a `"url"` request kind to `chatgpt_bridge.js`'s `fetchAssetBytes` that skips the metadata round trip entirely and reuses the **exact same** fetch+budget+hash mechanism already proven for the `sandbox`/`file` kinds (refactored the shared tail into `fetchBytesFromResolvedUrl`), including the existing cross-origin credential-drop rule (never forward page cookies/bearer to a provider-issued or DOM-rendered cross-origin URL). `chatgpt.js`'s `collectTurns` now awaits a new `acquireDomAttachmentBytes` step — budget-bounded identically to the native path (same per-file/total-byte/time/consecutive-failure caps) — that requests those bytes and sets `inline_base64` on the DOM attachment. That is the exact field `polylogue/sources/parsers/browser_capture.py` already decodes into `ParsedAttachment.inline_bytes` for every adapter, so **no receiver/parser change was needed** — `inline_base64` is not a new payload field, just one `chatgpt-dom-v1` was never wired to populate.

Chips with no resolvable URL stay honestly `byte_count=0`.

## Consistency with polylogue-ptx
`polylogue-ptx` (PR #2928, merged) replaced the extension's browser-action/campaign machinery with a provider-neutral `BrowserAction` conduit. This change only extends the existing content-script/bridge byte-fetch mechanism at its one documented gap; it does not touch, duplicate, or compete with the `BrowserAction` conduit's submit/receipt machinery.

## Verification
- `cd browser-extension && npx vitest run` → **329 passed** (1 pre-existing unrelated failure in `build.test.js`'s packaged-service-worker backfill fixture — reproduces identically without this change)
- `npx eslint src/ tests/` → clean
- `node scripts/validate-manifest.mjs` → ok
- New coverage: 5 focused `fetchAssetBytes` `"url"`-kind tests (direct fetch, cross-origin credential drop, same-origin credential keep, non-https rejection, byte-cap enforcement) plus 2 full-capture DOM-fallback smoke tests (`chatgpt_bridge.test.js`) executing the real production sources through a simulated authenticated page, proving a `chatgpt-dom-v1` capture now yields `inline_base64` whose SHA-256 exactly matches the delivered bytes, and that a chip with no fetchable URL stays honestly `byte_count=0`.

Ref polylogue-83u.3


Sinity:
Addressed the Codex finding in 669c122df:

- **P2** (unbounded buffering before the size check): added `readBoundedBody`, which reads the response via `response.body.getReader()` and cancels the stream the moment accumulated bytes exceed the budget, instead of buffering the whole response with `arrayBuffer()` and checking size only afterward. `fetchBytesFromResolvedUrl` now uses it for every asset kind (sandbox/file/url). Two new tests prove the actual mid-stream bound with no declared `Content-Length`: an oversized undeclared-length body is rejected (`downloaded_bytes_over_limit`) without ever needing the header, and a within-budget undeclared-length body is still acquired correctly.

330 tests pass (`npx vitest run`, 1 pre-existing unrelated `build.test.js` failure), `npx eslint src/ tests/` clean.

Re: `ci/circleci: quick-gate` failing — confirmed pre-existing/unrelated: it also fails on `4053787ab` (master itself, PR #3064, already merged). Not something this PR introduced.

PR #3097, opened 2026-07-18
feat(webui): render transcript messages through the shared semantic-card registry

Sinity:
## Summary

Builds the fourth WebUI v2 vertical (mission webui-04): the session read page's message bodies now project `polylogue/rendering`'s semantic-card registry (shell/edit/read/search/web/task/mcp/lineage/attachment/fallback cards, structural outcomes, bounded previews) instead of the placeholder role+text projection webui-02 shipped as an explicit interim seam.

## Problem

Session→HTML rendering is fragmented across five implementations (the canonical `polylogue/rendering/` registry, the legacy web shell's JS-in-Python-strings, and a CLI markdown axis). webui-02's session read page deliberately rendered a "deliberately simple placeholder" for message bodies and named the exact interface webui-04 would replace it through, to avoid this vertical adding a sixth divergent renderer. The daemon already serializes the registry's card documents per message (`semantic_entries`/`semantic_cards`/`semantic_card_suppressed`, schema-validated against `docs/schemas/semantic-card-v1.schema.json`) for the legacy web shell's client-side JS to consume — nothing in WebUI v2 consumed them yet.

## Solution

- `polylogue/daemon/webui.py`: replace `_render_message_flow_item`'s plain role/text projection with a dispatcher over each message's `semantic_entries` (card/prose/notice), reusing the exact JSON the daemon already produces via `semantic_card_placement_for_messages` — no new web-local classification, no new backend endpoint. A message with no materialized entries still falls back to the honest plain placeholder (older un-reindexed rows never silently vanish). `semantic_card_suppressed` messages (pure tool-result rows fully absorbed into a paired card) render nothing, matching the legacy shell's suppression behavior.
- `_render_lineage_banner` now prefers the registry's evidence-bearing lineage card (relation, authority, availability, cycle detection) over the bare `parent_id`/`branch_type` pair webui-02 shipped, when one is materialized in the session-level `semantic_entries`.
- Card field values following the registry's `session:<id>`/`message:<id>` ref convention (used by the lineage and task cards) are hyperlinked generically — to `/app/sessions/:id` or `#msg-:id` — so any card family adopting that convention gets working navigation without a bespoke ref field. This is what "task/delegation: link to child session" resolves to today; the task card itself carries no dedicated child-session field yet.
- `webui/src/design-system/semantic-cards.tsx`: Preact components (`SemanticCardView`/`SemanticProseView`/`SemanticNoticeView`) mirroring the SSR renderer 1:1 by CSS class, so the "load more messages" island hydrates newly-fetched pages with the same card rendering instead of regressing to plain text.
- `webui/src/contracts/semantic-cards.ts`: lenient per-entry parsing (a malformed entry degrades to nothing rather than blanking the whole message, matching the observability view's per-panel isolation precedent) of the semantic-card-v1/semantic-transcript-v1 schemas.

Progressive disclosure is native `<details>`/`<summary>` for bounded previews and evidence — readable and expandable without JS, consistent with every other WebUI v2 page's no-JS-required baseline. Virtualizing very long transcripts and wiring a dedicated task/delegation child-session ref are follow-ups, not done here.

## Anti-fragmentation parity harness

`tests/unit/rendering/test_semantic_card_web_cli_parity.py` is the regression test the mission calls out as missing: for one session mixing a failed shell result (exit code 2), a forked lineage relationship, and an unrecognized provider tool, it asserts the web-serving path (`semantic_card_placement_for_messages`, exactly what `daemon/http.py` calls for every archive-backed session read) and the CLI's own projection pair (`build_semantic_transcript` + `render_semantic_transcript_markdown`) agree on card count, kind set, and structural outcome. A regression here means one surface silently reclassified a tool or dropped a card the other still shows.

## Verification

- `cd webui && npm run build && npm test` — 27 Vitest tests passed.
- `node scripts/lint-design-system.mjs` — OK.
- `devtools test tests/unit/daemon/test_web_reader.py::TestWebUIV2 tests/unit/rendering/test_semantic_card_web_cli_parity.py tests/unit/daemon/test_route_contracts.py` — 120 passed.
- `devtools test tests/unit/daemon/test_web_reader.py tests/unit/rendering/test_semantic_cards.py` — 257 passed, 3 pre-existing failures unrelated to this change (reproduce identically on the unmodified checkout: a missing generated fixture file `tests/data/semantic_cards/cases/result-before-use.json`, and the already-known sqlite-open failure in `TestReaderAssertionEndpoint`).
- `devtools verify --quick` — format, lint, strict mypy, and all generated-surface/policy gates passed.
- Live smoke: `polylogue demo seed` + `polylogued run` against the demo archive — exercised a real failed shell card (exit code 4, bounded output preview) with its paired tool-result message correctly suppressed, and a lineage card with working session/message hyperlinks (session, root, parent, branch point, active leaf).

## Follow-ups

- Virtualized rendering for very long transcripts.
- A dedicated task/delegation child-session ref field on the TASK card (today's TASK card has no `session:` field to link).
- The parity harness compares the archive-backed reader against the CLI's shared projection pair; extending it to the DB-backed reader (`lineage_descriptor_from_session`, used by `polylogue/cli/messages.py` and the daemon's non-archive-root async path) is a natural next slice toward ap7's acceptance criterion that archive-backed and DB-backed readers agree on lineage/delegation cards.

Ref polylogue-bby.11, polylogue-ap7

## Summary by CodeRabbit

* **New Features**
  * Session transcripts now display rich semantic cards, including outcomes, evidence, linked references, notices, prose, and diff previews.
  * Lineage information is shown as a dedicated card when available.
  * Unknown tools receive a clear fallback card instead of a raw placeholder.
  * Suppressed semantic entries are hidden from the transcript.
  * Responsive styling improves card and small-screen layouts.

* **Bug Fixes**
  * Older or unmaterialized messages continue to display their existing previews and status flags.


Sinity:
Addressed in a381f1748:
- A message whose `semantic_entries` are all malformed/unsupported now falls back to the plain-text placeholder instead of rendering blank.
- Unrecognized `SemanticCardSource` fields now surface in the evidence `<details>` block (known fields first, then remaining scalar keys) in both the Python SSR renderer and the Preact hydration component.
- The parity test now also asserts structural outcome (state/exit_code) agreement per card kind, not just kind/count.
- Lowercased the new named CSS colors for Stylelint; left CSS system-color keywords (`Highlight`/`Canvas`/`CanvasText`) as-is per this file's existing convention.

PR #3125, opened 2026-07-18
perf(daemon): reuse one connection pair across catch-up cursor reconciliation

Sinity:
## Summary

Sixth daemon-drain fix from the 2026-07-18 restore: catch-up *planning* opened `source.db` and `index.db` fresh for every cursor-less candidate file — ~40,000 SQLite connection opens for a 20k-file plan — costing ~10 minutes of silent 98%-CPU startup before the first scan log line, paid again on every daemon restart.

## Problem

py-spy against the live daemon during the silent startup phase: MainThread pinned in `_reconcile_archived_cursor_outcome` ← `_needs_work_from_state` ← `_plan_catch_up`, with each call doing two `sqlite3.connect(...)` opens. The per-file query itself is indexed (`idx_raw_sessions_source_path`) — the cost is pure connection churn.

## Solution

- `_archived_cursor_reconciliation_scope()`: a context manager that opens the read-only pair once and shares it for one bulk pass; both bulk call sites (`_plan_catch_up`, the live changed-file batch filter) wrap their loops in it.
- The scope is deliberately bounded to a single planning pass — a long-lived cached connection would keep reading a replaced `index.db` inode across a blue-green generation swap, which the per-call opens implicitly avoided. Single-file event paths keep per-call opens.
- Per-file logic is unchanged, extracted to `_archived_cursor_row` used by both paths; open-failure still degrades to `UNAVAILABLE` exactly as before.

## Verification

`devtools test tests/unit/sources/test_live_watcher.py` — 87 passed (the existing reconciliation tests exercise both the scoped and per-call paths). `mypy` clean; pre-push quick gate green.

Ref polylogue-5jak

https://claude.ai/code/session_01QUsH3Rhq6oAZpYPWcsZqnZ

## Summary by CodeRabbit

* **Performance**
  * Improved efficiency when processing large catch-ups and live update batches.
  * Reduced repeated database access during archived change reconciliation.

* **Reliability**
  * Improved handling of archived cursor checks, including safer behavior when database information is temporarily unavailable.

### DIFF
--- a/polylogue/daemon/cli.py
+++ b/polylogue/daemon/cli.py
@@ -686,10 +686,13 @@ def _drain_raw_materialization_once(
         logger.warning("raw materialization: bounded convergence incomplete: %s", result.detail)
     metrics = dict(getattr(result, "metrics", {}))
     remaining = int(metrics.get("raw_materialization_remaining_candidate_count", 0))
+    if remaining == 0:
+        remaining = int(metrics.get("raw_materialization_census_incomplete_raw_count", 0))
     return raw_authority.RawMaterializationCounts(
         repaired_sessions=result.repaired_count,
         executed_plans=frontier_repaired,
         remaining_candidates=remaining,
+        censused_components=int(metrics.get("raw_materialization_census_components_attempted", 0)),
     )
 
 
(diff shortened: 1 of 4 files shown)


Sinity:
Added a second commit: census-paused conveyor passes (live: 'Raw replay planning paused until the persisted parser census completes for 22,142 relevant raw(s)') now count as burst progress — without it the census advances 16 components per 30s tick (~half a day for the current 22k-raw backlog) because the burst's no-progress guard ended the burst after every census pass.

PR #3155, opened 2026-07-19
perf(core): optionalize orjson via a three-tier JSON backend facade

Sinity:
## Summary

Makes `orjson` an optional accelerator instead of a hard dependency, so polylogue can eventually run on a free-threaded (3.14t) Python interpreter where orjson has no wheels and refuses to compile. `polylogue/core/json.py` becomes a real three-tier backend facade (orjson → msgspec → stdlib json), and every direct `import orjson` outside that facade is migrated to route through it.

## Problem

The 3.14t free-threading experiment gate (polylogue-7mtf) passed with a wide margin (3.9x-9.6x thread-parse speedup at 7.7% single-thread tax — see the bead for the full benchmark table), but surfaced a hard blocker: `orjson` (a hard pyproject dependency, imported directly in 8+ modules) ships zero `cp314t` wheels and its build script explicitly refuses to compile under free-threaded Python. Phase 2 of the free-threading adoption program (polylogue-xikl) opens with clearing that blocker before any of the thread-parallel parse work can land.

## Solution

- **`polylogue/core/json.py`** now selects a backend at import time, in priority order: `orjson` (fastest, no cp314t support) → `msgspec` (ships cp314t wheels since 0.20.0, Nov 2025 — confirmed via nixpkgs at the pinned rev too) → stdlib `json` (always available). Backend selection uses `importlib.import_module` rather than literal `import X as _x` statements specifically so mypy's `ignore_missing_imports` behavior stays consistent whether or not the optional packages are actually installed in the checking environment (a real cross-environment mypy failure surfaced and was fixed during this PR — see Verification).
- Every direct `import orjson` outside the facade is migrated: `sources/decoder_json.py`, `sources/live/batch_support.py`, `browser_capture/actions.py`, `browser_capture/receiver.py`, `browser_capture/server.py`, `daemon/events_http.py`, `daemon/http.py`, `archive/raw_payload/decode.py`, and `material_protocol/v1/canonical.py` (the content-hash-critical canonical serializer, which previously bypassed the facade entirely).
- `dumps_bytes`/`dumps` gained backend-neutral semantic flags (`sort_keys`, `indent`, `append_newline`) replacing the old raw orjson `option: int` bitmask passthrough. Output is byte-identical across all three backends for the compact-sorted and 2-space-indent shapes every call site needs — verified empirically, not just asserted — which is what keeps `material_protocol/v1/canonical.py`'s content-hash stability intact regardless of which backend produced the bytes.
- A single `JSONDecodeError(ValueError)` unifies `orjson.JSONDecodeError`, `msgspec.DecodeError`, and `json.JSONDecodeError` so callers never import or catch a backend-specific exception type.
- Two real cross-backend seams closed in the facade itself (would have been silent correctness bugs otherwise):
  - msgspec encodes `decimal.Decimal` natively as a JSON *string* (unlike orjson/stdlib, which raise and defer to the `default`/`enc_hook` callback) — the facade pre-normalizes `Decimal` to `float` on the msgspec path so all three backends agree it's a JSON number.
  - msgspec's `enc_hook` contract signals "unsupported type" via `NotImplementedError`, not `TypeError` like orjson/stdlib's `default` — adapted internally so one encoder function works across all backends.
  - The stdlib-only decode path now explicitly rejects non-finite tokens (`NaN`/`Infinity`) to match orjson/msgspec's strictness; a bug in the first draft of the backend dispatch would have silently accepted them when stdlib is the *primary* backend (no orjson/msgspec installed at all).
- **`pyproject.toml`**: `orjson` moves from a hard dependency to the `speed` optional extra; `msgspec` gets its own `speed-msgspec` extra (kept separate rather than combined, because a single combined extra would make the whole install fail on a 3.14t interpreter where orjson simply cannot be installed at all). Both are added to the `dev` extra so CI/tests exercise all three real backends via monkeypatching `core_json._BACKEND`, never by requiring orjson's actual absence.
- **`flake.nix`**: the nix package's `dependencies` list is independently curated (not derived from pyproject extras), so it deliberately keeps `orjson` on the standard (GIL) build for the speed win — a comment documents that a future `python314t` package variant should drop `orjson` and add `msgspec` instead, with zero facade code changes needed.

## Verification

- `mypy` (strict) on all 11 touched files — clean, in **both** the worktree's freshly-synced venv (orjson + msgspec installed) and the main checkout's venv (msgspec not installed at all) — this cross-environment check caught and fixed a real `no-redef` mypy bug in the first draft (explicit `ModuleType | None` pre-declarations conflicted with `import X as _x` differently depending on whether the package resolved).
- `devtools test tests/unit/core/test_json.py tests/unit/daemon/test_health_contract.py tests/unit/daemon/test_daemon_http_security.py tests/unit/material_protocol tests/unit/browser_capture tests/unit/sources/test_source_laws.py tests/unit/core/test_raw_payload_decode.py` — 979/980 pass. The one failure (`test_no_token_in_log_or_print_calls`) reproduces byte-identically on an unmodified checkout (verified via `git diff` → `git checkout --` → rerun → `git apply`), confirmed pre-existing and unrelated to this change.
- `devtools verify --quick` — exit 0 (format/lint/mypy/render/layering/manifests/ci-workflows/doc-commands/docs-coverage/clock-hygiene/degrade-loudly all green); this is also what the pre-push hook ran (from the main checkout's venv) before allowing the push.
- Decode-throughput benchmark (reusing `tests/infra/revision_backfill_benchmark.py`'s SMALL/LARGE/REVISION_CHAIN payload shapes, best-of-3, forced backend per run, same machine as the 7mtf gate): **msgspec is actually 1.8-7.1% *faster* than orjson** on decode for these shapes; stdlib-only is 34-53% slower than orjson. Both deltas are dwarfed by the 3.9x-9.6x thread-parallelism win the 3.14t gate already measured — recorded on the epic.

## Backend-ordering / packaging decisions (recorded on polylogue-xikl)

- Backend ordering: orjson → msgspec → stdlib. msgspec ships real `cp314t` wheels (confirmed on PyPI: 0.21.1 covers manylinux x86-64/ARM64 + macOS) and nixpkgs at the pinned rev already has a `msgspec` python-module derivation (built from source, so free-threaded compilation isn't blocked upstream the way orjson's is).
- Nix packaging: kept as a hard dependency on the standard build (independent of the pyproject extra split) since the flake's `dependencies` list is hand-curated, not extras-derived. A future 3.14t package variant swaps `orjson` for `msgspec` in that one list.

Ref polylogue-xikl.3

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -80,6 +80,16 @@
           click
           tenacity
           dateparser
+          # orjson became an OPTIONAL accelerator in pyproject.toml's `speed`
+          # extra (polylogue-xikl): polylogue.core.json falls back to
+          # msgspec/stdlib json when it's absent. This nix build's
+          # `dependencies` list is independently curated (not derived from
+          # pyproject extras), so it deliberately keeps orjson on the
+          # STANDARD (GIL) build for the speed win. A future python314t
+          # package variant should drop `orjson` from this list (it has no
+          # cp314t wheels and refuses to compile free-threaded) and add
+          # `msgspec` instead -- the facade's fallback ordering handles the
+          # rest with no code change.
           orjson
           structlog
           pydantic
--- a/polylogue/archive/raw_payload/decode.py
+++ b/polylogue/archive/raw_payload/decode.py
@@ -6,15 +6,13 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Literal, TypeAlias, cast
 
-import orjson
-
 from polylogue.archive.artifact_taxonomy import (
     ArtifactClassification,
     classify_artifact,
 )
 from polylogue.archive.raw_payload.streams import raw_line_stream
 from polylogue.core.enums import Provider
-from polylogue.core.json import JSONDocument, JSONValue, is_json_value, loads
+from polylogue.core.json import JSONDecodeError, JSONDocument, JSONValue, is_json_value, loads
 from polylogue.sources.dispatch import detect_provider
 
 _HERMES_STATE_DB_MARKER = "hermes_state_db"
@@ -28,7 +26,7 @@ def _decode_provider_utf8(raw: bytes) -> str:
 
     Some historical exports contain a lone UTF-16 surrogate encoded directly
     as its three-byte UTF-8 sequence. This is invalid Unicode scalar UTF-8,
-    so ``orjson`` correctly rejects it, but Python can preserve the original
+    so the active JSON backend correctly rejects it, but Python can preserve the original
     code unit with ``surrogatepass``. Arbitrary malformed byte sequences still
     raise and retain the ordinary malformed-JSONL behavior.
     """
@@ -44,9 +42,9 @@ def _decode_provider_utf8(raw: bytes) -> str:
 def _load_json_record(line: str) -> JSONValue:
     try:
         return loads(line)
-    except orjson.JSONDecodeError:
+    except JSONDecodeError:
         # Retry with stdlib json — tolerant of raw control characters
-        # (ANSI escape codes in bash output, etc.) that orjson rejects.
+        # (ANSI escape codes in bash output, etc.) that the active JSON backend rejects.
         import json
 
         return cast("JSONValue", json.loads(line))
@@ -56,12 +54,12 @@ def _load_raw_json(raw: bytes | str) -> JSONValue:
     """Parse a JSON document, retaining recoverable provider surrogates."""
     try:
         return loads(raw)
-    except (orjson.JSONDecodeError, ValueError) as error:
+    except (JSONDecodeError, ValueError) as error:
         if not isinstance(raw, bytes):
             raise
         try:
             return _load_json_record(_decode_provider_utf8(raw))
-        except (orjson.JSONDecodeError, ValueError, UnicodeDecodeError):
+        except (JSONDecodeError, ValueError, UnicodeDecodeError):
             raise error from None
 
 
@@ -111,7 +109,7 @@ def _decode_jsonl_payload(
                 continue
             try:
                 parsed = _load_json_record(line)
-            except (orjson.JSONDecodeError, ValueError) as exc:
+            except (JSONDecodeError, ValueError) as exc:
                 malformed_lines += 1
                 if malformed_detail is None:
                     malformed_detail = f"line {line_number}: {exc}"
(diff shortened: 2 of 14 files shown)


Sinity:
## Fixed: msgspec float-exponent divergence (coordinator review)

Repro confirmed exactly as reported: `dumps_bytes({"exp": 1e30, "big": 1.7976931348623157e308, "tiny": 5e-324}, sort_keys=True)` — orjson/stdlib gave `"1e+30"`/`"...e+308"`, msgspec gave `"1e30"`/`"...e308"` (missing `+`).

**Investigated further before implementing the suggested fix shape** (skip msgspec, fall back to stdlib for canonical dumps) — turns out stdlib is *not* actually byte-identical to orjson in general:

```
1e-05  -> orjson "0.00001"   stdlib "1e-05"     (different decimal/exponent threshold entirely)
2e-06  -> orjson "2e-6"      stdlib "2e-06"     (stdlib zero-pads short exponents)
```

The reviewer's 3-value repro corpus happened not to hit either of these (its exponents were all ≥2 digits, where stdlib and orjson coincidentally agree). So "orjson → stdlib, skip msgspec" would not have achieved real byte-for-byte stability, and would have thrown away msgspec's speed on exactly the hash-critical canonical path that matters most under 3.14t.

**Implemented instead:** `_normalize_msgspec_float_exponents` in `polylogue/core/json.py` post-processes msgspec's raw encoded bytes to insert the one thing it's actually missing — the `+` sign on positive exponents. Verified empirically that this is the *only* divergence between orjson and msgspec across a wide magnitude spread (1e-100 through 1e300, negative exponents, negative zero, non-ASCII). Applied unconditionally to every msgspec `dumps_bytes` call (not behind a canonical-only flag), so msgspec remains a fully valid, fast backend for hash dumps too — no speed sacrificed.

The normalizer is a two-alternative regex (string-literal-first, so `"batch e5"`-shaped string content is never mistaken for a number) — added a dedicated test proving that safety property.

stdlib's larger, unreconciled divergence is now explicit in the module docstring and pinned by `test_stdlib_diverges_from_orjson_for_small_exponents` so it reads as a documented, accepted last-resort-tier limitation rather than an oversight if someone goes looking later.

Commit aa7dec82d. Re-verified: `mypy` clean (both msgspec-present and msgspec-absent venvs), `devtools test` on the same file set = 983/984 (the 1 failure is the same pre-existing/unrelated one from the first review pass), `devtools verify --quick` exit 0.

PR #3166, opened 2026-07-19
perf(startup): cut CLI import tax + scrub nix wrapper env leaks

Sinity:
## Summary

Cuts import-time tax on three CLI paths via lazy parent-package inits (PEP
562 `__getattr__`, matching the existing pattern in `polylogue/storage/__init__.py`),
plus a nix wrapper fix for a free-threaded-build environment leak found while
verifying the epic's own follow-up note.

## Problem

- **polylogue-h1wt**: importing a single `ArchiveTier` enum
  (`polylogue.storage.sqlite.archive_tiers.types`) cost ~950ms-1.2s in a fresh
  process, because Python must run the parent packages'
  `__init__.py` first, and `polylogue/storage/sqlite/__init__.py` eagerly
  imported `SQLiteBackend` (the whole 10-mixin async backend) at module level.
- **polylogue-8s70**: `polylogue status`/`agents status --format json` paid
  ~1.5-2s of command-body import tax beyond `--help`'s ~0.4s budget, traced via
  cProfile to `storage.repair`'s own heavy import list (in particular a direct
  `sources.dispatch` import that pulled in the entire Google Drive download
  subsystem) plus eager `operations`/`insights` package inits reached through
  `readiness/capability.py`.
- **polylogue-xikl** (nix wrapper follow-up): the `polylogue-freethreaded`
  (python3.14t) package's own PR (#3162) already unset
  `PYTHONPATH`/`PYTHONHOME`/etc in its wrapper, but verifying it under this
  repo's own ambient 3.13 devshell environment still reproduced
  `ModuleNotFoundError: No module named '_sysconfigdata__linux_x86_64-linux-gnu'`
  on any command doing real work (`--version`/`--help` short-circuit before
  hitting the failing path, which is why the existing fix looked sufficient).

## Solution

**Import-tax fixes** (each measured individually, isolated `python -c` /
`-X importtime` / cProfile on this host):
- `polylogue/storage/sqlite/__init__.py`: defer `SQLiteBackend` via PEP 562
  `__getattr__`. `archive_tiers.types` import: ~1000ms → ~104ms.
- `polylogue/pipeline/ids.py`: `ParsedMessage`/`ParsedSession`/etc are
  type-only here; moved to `TYPE_CHECKING` so this pure hashing module stops
  forcing the whole `polylogue.sources` package (incl. Drive subsystem) on
  every caller.
- `polylogue/storage/repair.py`: deferred `sources.dispatch.detect_provider`/
  `is_stream_record_provider` into the 6 functions that actually call them
  (matching the file's own existing local-import convention for
  `sources.revision_backfill`). `import polylogue.storage.repair`:
  ~670ms → ~270ms.
- `polylogue/sources/__init__.py`: lazy re-exports so importing a submodule
  (e.g. `parsers.base`) no longer forces the Drive subsystem (`tenacity`,
  `drive.auth`, `drive.gateway`, ...) to load.
- `polylogue/operations/__init__.py`, `polylogue/insights/__init__.py`: same
  lazy-package-init pattern.
- `polylogue/operations/operation_status.py` (new): split `OperationStatus`
  (a plain `str` `Enum`) out of `operation_contract.py`, whose other export
  (`OperationFollowUp`) subclasses a pydantic `SurfacePayloadModel` that pulls
  in ~280ms of archive/semantic pricing machinery.
  `readiness/capability.py` only ever needed the enum.
- `devtools/help_latency_probe.py` + `_migrate_tier.py`: promoted
  `ops-maintenance-migrate-tier` from `informational` to `required`
  (~400ms, was ~1.04s against the 700ms budget).

Two additional micro-fixes were tried, measured as **zero benefit**, and
**reverted** rather than landed (extracting `date_from_iso` out of
`insights/archive.py`; `TYPE_CHECKING`-only `ArchiveCoverageInsight` in
`cli/shared/helper_summary.py`) — both callers need the heavy module anyway
via a separate real dependency (`archive_rollups.py`'s
`CostRollupInsight`/etc).

**Nix wrapper fix**: bisected the free-threaded-build failure with `env -i`
(adding exactly one inherited var back at a time) — `PYTHONPATH` alone does
NOT reproduce it. The actual trigger is `_PYTHON_SYSCONFIGDATA_NAME`, a
nixpkgs-set env var that Python's own `sysconfig._get_sysconfigdata_name()`
trusts over its own computation when present. The free-threaded build's real
sysconfigdata module carries a `t` abiflag segment
(`_sysconfigdata_t_linux_x86_64-linux-gnu`), but a 3.13-devshell-derived
value of this env var (`_sysconfigdata__linux_x86_64-linux-gnu`, no `t`)
doesn't match it. Added `--unset _PYTHON_SYSCONFIGDATA_NAME --unset
_PYTHON_HOST_PLATFORM` to all three wrapper sites in `flake.nix`.

## Verification

- `devtools verify --quick`: green (format/lint/mypy/render-all-check),
  including the repo's own pre-push hook run.
- `devtools test` on all touched files + their existing test coverage
  (`test_repair.py`, `test_archive_debt.py`, `test_specs.py`,
  `test_operation_contract.py`, `test_readiness_capability.py`,
  `test_drive_ops.py`, `pipeline/*`, `test_archive_query.py`,
  `test_insight_registry_runtime.py`): 702 passed, 6 pre-existing failures in
  `tests/unit/pipeline/test_parsing_service.py` (`Mock(spec=Config)` lacks
  `drive_config` — verified identical on the pre-change base commit via a
  disposable `git worktree`, unrelated: `config.py` is untouched here).
- Caught + fixed a real mypy regression before landing: the initial
  `sources/__init__.py` lazy-export commit missed `download_drive_files` in
  the `TYPE_CHECKING` block, which `devtools verify --quick` caught
  (`"object" not callable` in `test_drive_ops.py`) — fixed in a follow-up
  commit.
- Timing deltas (this host, empty archive root, cold subprocess, 3 runs):
  `polylogue status --format json` ~2.05s → ~1.47s; `agents status
  --format json` ~1.8s → ~1.3s; `ops maintenance migrate-tier --help` ~1.04s
  → ~0.40s.
- Nix: `nix build .#polylogue` and `.#polylogue-freethreaded` both succeed;
  with this session's actual ambient devshell env still exported, both
  packages' `polylogue demo seed` + `polylogue demo verify` complete
  end-to-end (16 sessions/62 messages/4 query hits) — the exact real-ingest
  path that previously raised `ModuleNotFoundError` on the freethreaded build.

## Acceptance-criteria honesty

- polylogue-h1wt: **satisfied** — `archive_tiers.types` import lands at the
  same order of magnitude as the bead's own reference baseline (~104ms vs
  ~90ms target); `ops-maintenance-migrate-tier` promoted to `required`; no
  regression found.
- polylogue-8s70: **not fully satisfied**. The bead's own AC (`status`/`agents
  status` under the 700ms cold-CLI budget) is not met — both commands are
  meaningfully faster (~28-30% cut) but still roughly 2x over budget. The
  remaining cost is `insights.archive`'s own weight (pydantic
  `CostRollupInsight`/`SessionCostInsight`/`ArchiveCoverageInsight` models
  plus `archive.semantic.pricing`/`content_projection`), pulled in for real
  by session-tag-rollup computation
  (`storage/insights/session/aggregates.py` → `archive_rollups.py`) and
  archive-coverage reporting — not an eager-import artifact fixable by moving
  an import statement. Closing the remaining gap needs either restructuring
  those pydantic model definitions or splitting session-tag-rollup
  computation out of the status command's default path, which is a bigger,
  riskier architectural change than this PR's scope. Recorded on the bead;
  recommend a follow-up scoped specifically to that remaining cost.
- polylogue-xikl (nix wrapper note): **satisfied** — reproduced, root-caused,
  fixed, and verified end-to-end for both packages.

Ref polylogue-h1wt, Ref polylogue-8s70, Ref polylogue-xikl


Sinity:
GitGuardian false positive triage: the flagged "Authentication Tuple" at `polylogue/sources/__init__.py` (the `"DriveAuthError": (".drive.types", "DriveAuthError")` dict entry in the lazy `__getattr__` re-export table) is a 2-tuple of `(module_path, attr_name)` strings, not a credential. The detector appears to be pattern-matching a 2-element tuple literal near the word "Auth" (from the real class name `DriveAuthError`). No secret present.

PR #3174, opened 2026-07-19
fix(insights): aggregate session_profiles tokens from canonical usage substrate

Sinity:
## Summary

`session_profiles` token/cost columns (`total_input_tokens`, `total_output_tokens`,
`total_cache_read_tokens`, `total_cache_write_tokens`, `total_cost_usd`, ...)
now aggregate from `session_model_usage` -- the same canonical usage substrate
the archive's own cost/usage rollups are built from -- instead of independently
recomputing an estimate from per-message token fields.

## Problem

Live-archive finding (polylogue-r7p6): for `codex-session`, `session_profiles`
token columns summed to 6.43M input tokens across 3,134 sessions (1,250 of
them exactly zero), while `session_model_usage` for the *same* sessions summed
to 6.74B input + 0.59B output tokens -- a three-orders-of-magnitude split-brain
between two read models that are both supposed to describe the same usage.
`claude-code-session` profiles were fine (27.7B cache-read tokens visible).

Root cause: `build_session_profile` -> `compute_session_cost`
(`polylogue/archive/semantic/cost_compute.py`) walked `session.messages`,
reading each message's `input_tokens`/`output_tokens`/`cache_read_tokens`/
`cache_write_tokens` fields. Those fields are populated by the Codex parser
(`sources/parsers/codex.py:_token_usage`) only when a message record embeds a
`usage` block directly, which is rare -- Codex's real usage arrives as
periodic *cumulative* `token_count` session events instead (`last_token_usage`
+ session-global `total_token_usage`). Those `token_count` events are
deliberately excluded from the `session_events` table as redundant with
`session_model_usage` (`_SESSION_EVENTS_REDUNDANT_TYPES` in
`storage/sqlite/archive_tiers/write.py`), and folded into
`session_model_usage` by `_aggregate_provider_usage_into_model_usage` (with
Codex's disjoint-lane mapping: input excludes cached, output includes
reasoning) instead. `session_profiles` never read that table at all, so for
sessions whose only usage evidence is the cumulative event stream, the profile
recomputation saw nothing but a crude word-count fallback -- explaining the
~1000x gap and the 1,250 exact zeros.

## Solution

- `compute_session_cost` (`polylogue/archive/semantic/cost_compute.py`) gains
  an optional `model_usage: Sequence[ModelUsageTotals]` parameter (new type in
  `archive/semantic/cost_records.py`). When supplied and non-empty, it is the
  **sole, provider-neutral source** of per-model token counts for every
  origin (not just Codex) -- the existing pricing/credit-cost loop underneath
  is unchanged, only the token-count *source* moved. The message-walk path
  (`_per_model_from_messages`) survives only as a fallback for callers that
  build an ad hoc profile with no materialized rollup to hand.
- `build_session_profile` (`archive/session/runtime.py`) and
  `build_session_insight_records` / `build_session_insight_record_bundles`
  (`storage/insights/session/rebuild.py`) thread `model_usage` through.
- New `polylogue/storage/sqlite/queries/model_usage.py`
  (`get_model_usage_batch` async / `sync_model_usage_batch` sync) batch-reads
  `session_model_usage` rows per session, mirroring the existing
  attachments/session-events batch-read pattern.
- `SessionInsightArchiveBatch` gains a `model_usage_by_session` field,
  populated in **both** `load_sync_batch` and `load_async_batch` -- wired into
  **both materializer twins**: `rebuild_session_insights_sync` and
  `rebuild_session_insights_async` (full rebuild), plus `refresh.py`'s
  single-session and bulk incremental-refresh paths (`storage/insights/session/refresh.py`).
- `docs/cost-model.md` gets a new "Single authority for per-session token
  totals" section naming `session_model_usage` as that authority.

**Known related gap, left out of scope:** the "large/degraded session" bounded
materialization path (`_large_session_profile_record[_from_row]` in
`rebuild.py`, sessions above `_SESSION_INSIGHT_DEGRADED_MESSAGE_THRESHOLD` =
10,000 messages) never populates token columns at all -- it explicitly writes
`cost_provenance="unknown"` and leaves totals at their 0 default, by design
("unknown, not fabricated"). That's a distinct, already-honest gap (it never
claims a wrong nonzero value), not the split-brain this bead reports, and
isn't what drives the reported 1000x/1,250-zero numbers (that threshold is
very high; the bug reproduces on ordinary small sessions). Not touched here.

**No schema change.** `session_profiles`' existing token/cost columns are
computed differently, not added to -- no migration, no index-tier bump.

**Existing archives self-heal.** `session_profiles` already carries a
materializer-version-stamped `insight_materialization` row and re-derives
automatically via the standard session-insight rebuild path (daemon's periodic
drain, hot-source convergence, convergence-debt retry) -- the same mechanism
that already repairs `session_model_usage` itself per `docs/cost-model.md`.
No `polylogue ops reset --index` is required, though a full rebuild also picks
up the fix immediately for anyone who wants it sooner.

## Verification

```
devtools test tests/unit/storage/test_session_profile_model_usage_consistency.py \
  tests/unit/storage/test_session_insight_profiles.py \
  tests/unit/storage/test_pricing_chain_roundtrip.py \
  tests/unit/storage/test_provider_usage_report.py \
  tests/unit/storage/test_incremental_rebuild_equivalence.py \
  tests/unit/storage/test_session_insight_rebuild_progress.py \
  tests/unit/storage/test_session_insight_parallel_fanout.py \
  tests/unit/storage/test_session_insight_refresh.py \
  tests/unit/cli/test_query_support_runtime.py
# -> 95 passed

devtools verify --quick
# -> ruff format/check, mypy --strict, render all --check, topology,
#    layering, closure-matrix, manifests, docs-coverage, degrade-loudly: all ok
```

New test file `tests/unit/storage/test_session_profile_model_usage_consistency.py`:
- seeds a Codex-shaped session (message with zero per-message token fields +
  a cumulative `token_count` event with realistic input/cached/output/reasoning
  numbers) through the real `write_parsed_session_to_archive` write path, then
  asserts `session_profiles` token columns equal `session_model_usage`'s
  aggregate **and** the concrete expected disjoint-lane numbers, after both
  `rebuild_session_insights_sync` and `rebuild_session_insights_async`.
- seeds an equivalent Claude Code (per-message-token) session proving no
  regression for the origin that was already correct.
- an explicit anti-vacuity test: building the profile from the same hydrated
  `Session` *without* `model_usage` reproduces the undercount (a few tokens
  from the word-count fallback vs the real ~4,000), and passing `model_usage`
  in corrects it -- so the fix, not a fixture artifact, is what makes the
  other tests pass.

Not run: the full non-focused test suite / integration tier (out of scope for
this change's blast radius; the affected-area focused set above covers every
call site touched).

Ref polylogue-r7p6


Sinity:
@coderabbitai review

Sinity:
Merging with coordinator (Fable) review standing in: both automated reviewers rate-limited without reviewing; diff reviewed manually — canonical-rollup-as-sole-source design, twin wiring, disjoint-lane fixture arithmetic, and empty-fallback semantics all verified. Ref polylogue-r7p6.

PR #3179, opened 2026-07-19
fix(demo): converge daemon import --demo path with direct seeder

Sinity:
## Summary

`polylogue import --demo --wait` (the live-daemon path) diverged from `polylogue demo seed` (the direct one-shot path) on the exact same fixture world: a different AI Studio session identity, missing provider-usage/embedding/repo demo enrichments, and a stale success banner / integration test asserting a 3-session/19-message world that predates the current 16-session demo corpus. This blocked publishing the README quickstart, which promises the two paths converge.

Ref polylogue-z1c6

## Problem

Reproduced both paths against isolated scratch archive roots (a real `polylogued run` subprocess for the daemon path, fully isolated `HOME`/`XDG_*`/`POLYLOGUE_*` env — no operator config, no live Drive/network access) and diffed `sessions`/`messages` row-for-row. Three independent, separately-verified divergences:

1. **Identity bug**: `aistudio-drive:demo-00` (direct seed) vs `aistudio-drive:demo-00-0` (daemon). The daemon's `revision_backfill._parse_one` decodes raw bytes through `list(_iter_json_stream(...))` — a list-shaped stream reader — even for a plain single-JSON-document file, while the direct seeder keeps the bare dict. That incidental list-wrapping trips `_lower_drive_like_payload`'s `_looks_like_chunked_session_list` branch, which (unlike its sibling fallthrough loop) always appends `-{index}` regardless of list length.
2. **Missing enrichment**: the daemon path never ran the four demo-only augmentation steps (`_inject_demo_session_usage`, `_materialize_session_insights`, `_inject_demo_session_repos`, `_seed_demo_embeddings`) that `seed_demo_archive` calls inline — so `provider_usage_messages`, `synthetic_message_embedding_rows`, `embedding_status_rows`, and the postmortem repo/cost facts were permanently absent from the daemon-produced archive.
3. **Stale banner/test**: the CLI success banner hardcoded `sessions=3 messages=19`, and `tests/integration/test_demo_daemon_convergence.py` asserted the same stale 3-session world — both predate the current 16-session demo corpus.

A fourth, deeper divergence was root-caused but **not fixed here**: one specific multi-material session (a direct ChatGPT export coalescing with its paired browser-capture variants) can nondeterministically lose 0-2 messages on the daemon path, because the daemon's incremental raw-materialization census (`classify_raw_revision_cohort`) can isolate-accept one competing raw as an "unambiguous singleton baseline" before its true siblings are discovered on a later tick, and a later membership-classification pass is then blocked from correcting it by an existing-head safety guard. This is a genuine architectural issue in the revision-authority subsystem, not a quick fix, and non-goal per this investigation's own findings (the "never choose between branches" conservatism for genuinely ambiguous data must not be weakened). Filed as `polylogue-52l2` with full root-cause detail, a suggested approach, and a related pre-existing (unrelated to this PR, confirmed by direct comparison against unmodified `master`) flake in the direct-seed path's own construct checks.

## Solution

- `polylogue/sources/dispatch.py`: apply the same `len(payloads) == 1` guard the sibling branch already uses in `_lower_drive_like_payload`, so a single-session wrapped payload keeps its bare fallback id.
- `polylogue/archive/session_revision_membership.py` + `polylogue/sources/revision_backfill.py`: wire `MembershipRevision.browser_snapshot_fidelity`/`provider_message_ids`/`provider_attachment_ids` from the parser's own ingest flags (previously never populated, making the whole dom/native precedence branch dead code), and add `_direct_export_precedence` so a genuine non-browser-capture revision always outranks browser-capture-only siblings.
- `polylogue/storage/sqlite/archive_tiers/ingest_precedence.py`: mirror the same direct-export-always-wins rule in the older/simpler `browser_capture_precedence` heuristic used by the non-revision-authority ingest path.
- `polylogue/demo/seed.py`: extract the four demo-only augmentation steps into `apply_demo_post_ingest_augmentation()` (idempotent), exported from `polylogue.demo`.
- `polylogue/cli/commands/import_command.py`: call the shared augmentation after daemon convergence (same place `--with-overlays` already runs client-side), derive the success banner's `sessions=N messages=M` from the real verifier result instead of a hardcoded placeholder, and split the wait-loop's semantic check (`verify_demo_archive(check_constructs=False)`) from the final full check so waiting on base ingest convergence doesn't deadlock against constructs only the augmentation stage can satisfy.
- `polylogue/demo/verify.py`: add the `check_constructs` parameter enabling the above.
- Tests: unit regressions for each fix (dispatch identity, membership-classification precedence, ingest-flag wiring, augmentation idempotency/parity), plus a rewritten `tests/integration/test_demo_daemon_convergence.py` against the current 16-session `DEMO_SESSION_IDS` world that asserts the fixed regressions and tolerates (with an explicit, bounded, documented assertion) the one still-open construct/message-count gap tracked in `polylogue-52l2`.

## Verification

- `mypy` on every touched file — clean (`Success: no issues found in 13 source files`).
- `ruff format --check` / `ruff check` — clean.
- `devtools test tests/unit/sources/test_dispatch_payloads.py tests/unit/archive/test_session_revision_membership.py tests/unit/sources/test_revision_backfill.py tests/unit/demo/test_demo_seed_verify.py` — 64 passed, 3 pre-existing failures classified below.
- `pytest tests/integration/test_demo_daemon_convergence.py -q -p no:randomly --timeout=280` — 1 passed (real `polylogued run` subprocess, real `import --demo --wait`).
- `devtools verify --quick` — exit 0 (format, lint, mypy, render-all-check, topology/layering/manifests/doc-commands/etc.).
- **Pre-existing failure classified, not caused by this PR**: `tests/unit/demo/test_demo_seed_verify.py::test_seed_demo_archive_creates_ready_queryable_archive` / `test_demo_verify_reports_missing_overlays` / `test_demo_verify_can_skip_daemon_source_path_leak_posture` flake (~40-60% failure rate across repeated `-p no:randomly` runs) on `source_outage_interval_events`/`capture_gap_events` construct checks in the **direct seeder** path. Confirmed by temporarily restoring the pre-change `ingest_precedence.py` from `git show HEAD` and re-running the same test 6x — identical flake rate with or without this PR's changes. Root cause not isolated (file-walk order is already deterministically sorted, ruled out); documented as an addendum to `polylogue-52l2`.

## Acceptance criteria (from polylogue-z1c6)

- Session/message identity converges for the AI Studio session and all provider-usage/embedding/repo enrichments: **satisfied**.
- Success banner and integration test assert the current canonical world: **satisfied**.
- Full 37-construct parity + `polylogue demo verify` passing against the daemon-produced archive: **not satisfied** for one multi-material session's 3 constructs, tracked in `polylogue-52l2` with the exact root cause and a suggested fix. Bead left open per the investigation's own scope decision (not silently closed).

### DIFF
--- a/polylogue/archive/ingest_flags.py
+++ b/polylogue/archive/ingest_flags.py
@@ -7,9 +7,19 @@ DOM_FALLBACK_INGEST_FLAG = "capture:dom-fallback"
 NATIVE_BROWSER_CAPTURE_INGEST_FLAG = "capture:browser-native-payload"
 COMPACT_BROWSER_CAPTURE_INGEST_FLAG = "capture:browser-native-compact"
 
+# Both flags mark "this session's content came from a native (non-DOM-fallback)
+# browser payload" -- `browser_capture.py` tags a session with one or the
+# other depending on whether the capture used the compact or full native
+# shape (see `_is_compact_native_capture`), never both. Any caller deriving
+# "does this session have native-payload precedence" must check both, or a
+# compact capture silently misclassifies as plain, non-browser-capture
+# content (polylogue-z1c6 review follow-up).
+NATIVE_BROWSER_CAPTURE_FLAGS = (NATIVE_BROWSER_CAPTURE_INGEST_FLAG, COMPACT_BROWSER_CAPTURE_INGEST_FLAG)
+
 __all__ = [
     "DOM_FALLBACK_INGEST_FLAG",
     "COMPACT_BROWSER_CAPTURE_INGEST_FLAG",
+    "NATIVE_BROWSER_CAPTURE_FLAGS",
     "NATIVE_BROWSER_CAPTURE_INGEST_FLAG",
     "TEMPORARY_CHAT_INGEST_FLAG",
 ]
--- a/polylogue/archive/session_revision_membership.py
+++ b/polylogue/archive/session_revision_membership.py
@@ -69,10 +69,18 @@ def classify_membership_revisions(revisions: list[MembershipRevision]) -> Member
     ):
         browser_order = _provider_ordered_browser_snapshots(representatives)
         if browser_order is None:
+            direct_export = _direct_export_precedence(representatives)
+            if direct_export is None:
+                return MembershipClassification(
+                    (),
+                    tuple(sorted(equivalents)),
+                    tuple(sorted(item.raw_id for item in representatives)),
+                )
+            accepted, browser_capture_raw_ids = direct_export
             return MembershipClassification(
+                (accepted.raw_id,),
+                tuple(sorted((*equivalents, *browser_capture_raw_ids))),
                 (),
-                tuple(sorted(equivalents)),
-                tuple(sorted(item.raw_id for item in representatives)),
             )
         representatives = browser_order
     return MembershipClassification(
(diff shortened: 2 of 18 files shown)


Sinity:
Addressed all three review findings:

**P1 #1 (order-independence)**: added the mirror rule to `browser_capture_precedence` (`incoming_is_browser_capture and not existing_is_browser_capture -> skip`), symmetric to the rule this PR already had. This changed the intentional policy encoded by the pre-existing `test_archive_tiers_archive_facade_native_browser_precedence_matrix` / `test_write_session_native_browser_precedence_matrix` suites (they asserted a "richer/tied content wins, ties favor native" policy for exactly the cases this fix targets), so I updated both matrices' expectations plus the three-arrival tests, and added a new parametrized test (`test_archive_tiers_archive_facade_export_vs_native_precedence_is_order_independent`) that runs both arrival orders back-to-back and asserts they converge to a byte-identical final transcript — proving order-independence directly instead of relying on eyeballing separate before/after rows.

**P1 #2 (compact flag)**: added `NATIVE_BROWSER_CAPTURE_FLAGS` (both `NATIVE_BROWSER_CAPTURE_INGEST_FLAG` and `COMPACT_BROWSER_CAPTURE_INGEST_FLAG`) as the one place enumerating "native browser-capture provenance"; `session_has_parser_ingest_flag`/`_incoming_has_ingest_flag` now accept a flag or sequence of flags. Both callers (`archive.py`, `ingest_batch/_core.py`) updated.

**P2 #3 (augmentation race)**: picked the "make it idempotently reasserted" option — `apply_demo_post_ingest_augmentation` now bounds a small self-heal loop (checks whether the injected usage survived materialization, re-applies if not, a few times a short beat apart). No CLI change needed since the self-heal lives in the shared function the CLI already calls once. Confirmed live (real `polylogued run` + `import --demo --wait`) that `total_cost_usd` reads back correctly now.

Verification: `devtools test tests/unit/storage/test_archive_tiers_archive.py tests/unit/pipeline/test_ingest_batch.py tests/unit/sources/test_dispatch_payloads.py tests/unit/archive/test_session_revision_membership.py tests/unit/sources/test_revision_backfill.py tests/unit/demo/test_demo_seed_verify.py` — 159 passed, 1 pre-existing/unrelated failure (`test_exact_session_action_count_bounds_pairing_before_global_ranking`, an environment-sensitive SQLite VM-step threshold check confirmed via `git diff HEAD` to be untouched by this change). `pytest tests/integration/test_demo_daemon_convergence.py -q -p no:randomly --timeout=280` — 1 passed (real daemon). `devtools verify --quick` — exit 0. mypy clean, ruff clean.

Rebased onto current `master` (picked up #3182's gemini-cli/antigravity/hermes demo origins along the way; updated the integration test's staged-directory assertion to derive from `demo_source_specs()` instead of a hardcoded list so it doesn't go stale again).

PR #3180, opened 2026-07-19
chore: excise entropic .agent content + clean root dir

Sinity:
## Summary

Untracks entropic agent-run output from `.agent/` and removes/consolidates
orphaned root-level directories, per operator directive. No content is
deleted from the current checkout except two genuinely dead example files
(see disposition table); everything else stays on disk, only `git rm --cached`.

## Problem

The public polylogue repo (GitHub-visible) had accumulated three kinds of
entropy: (1) `.agent/reports/`, `.agent/archive/`, and part of
`.agent/scratch/` were tracked via `!`-negations in `.gitignore` despite
being agent-run evidence/scratch, not reusable scaffold; (2) a pointer line
in a tracked handoff README named a private out-of-repo filesystem path;
(3) `contrib/` and `systemd/` each held one orphaned, unreferenced,
mutually-diverged systemd unit example alongside a genuinely-referenced
script.

## Solution — disposition table

| Path | Action | Reason |
| --- | --- | --- |
| `.agent/reports/` (10 files: verify logs, decision briefs, narration jsonl) | `git rm --cached` + gitignore | Agent-run evidence, not scaffold; was tracked via a `!.agent/reports/**` negation. Stays on disk. |
| `.agent/archive/` (298 files: retired devloop conductor packet, conductor-history, includes-2026-07, retired-demos, scratch-2026-07) | `git rm --cached` + gitignore | Retired scaffold/evidence explicitly marked "never resurrect" in CLAUDE.md; was tracked via a `!.agent/archive/**` negation. Stays on disk. |
| `.agent/scratch/archive/scratch-2026-07-16-loose-files/` (43 files) | `git rm --cached`, removed the `!.agent/scratch/archive/**` negation | Root cause of "wtf, why is this in the repo?": `.agent/scratch/*` is gitignored by design, but a later `!.agent/scratch/archive/**` negation explicitly re-included this one subtree, so plain `git add` picked it up. Negation removed so scratch stays gitignored end-to-end. Stays on disk. |
| `.agent/handoffs/README.md` pointer line ("For the Gemini/AI Studio equivalent, see `/realm/inbox/handoffs/polylogue-gemini-2026-07-16/`") | Line deleted | Named a private, unresolvable-by-others filesystem path outside the repo. Audited the rest of `.agent/handoffs/` for the same pattern (`grep` for `/realm/`, `/home/sinity`, "outside this repo", "tracked separately"); every other hit is an incidental evidence path inside research prose (e.g. "the archive lives at `/home/sinity/.local/share/polylogue`"), not a navigational pointer — left untouched per "do not restructure otherwise." |
| `contrib/polylogued.service` | Deleted | Diverged duplicate of `systemd/polylogued.service` (different `Description`, `After=`, resource limits, `ReadWritePaths` vs `ReadOnlyPaths`). Zero consumers. |
| `systemd/polylogued.service` (+ now-empty `systemd/` dir) | Deleted | NixOS/Home-Manager deployment generates its unit inline from Nix options (`nix/module.nix` → `systemd.services.polylogued`, `nix/hm-module.nix` → `systemd.user.services.polylogued`); `docs/installation.md` documents only that Nix-generated path. Neither static file is referenced by any install flow, packaging script, Containerfile, CI workflow, or doc — confirmed by repo-wide grep. |
| `contrib/polylogue-hook` | **Kept in place** | Not dead weight: documented as the canonical bash-form install path in `docs/hooks.md`, referenced from `packaging/polylogue-hooks/README.md` and its CLI docstring, and directly read + asserted against by `tests/unit/sources/test_hook_spool.py`. `contrib/` now holds only this file. |
| `scripts/cost_accounting_demo.py` | **Kept in place** | Has unique, real demo value (reproduces the real Codex double-billing cost bug end-to-end through the real writer, no mocks, hand-checkable math) and is the officially documented reproduction command in `docs/cost-model.md` ("Reproduce it" section, `uv run python scripts/cost_accounting_demo.py`). `scripts/` is documented in `docs/repo-layout.md` as exactly the right home for "standalone operator/demo scripts useful outside the devtools command registry" — this is that. A prior research pass (`.agent/handoffs/polylogue-deep-research-2026-07-09/2026-07-09-codebase-structure-audit.md`) independently confirmed it is "not orphaned." Deleting it would break the documented repro command with no replacement; folding it into a `devtools` `CommandSpec` or the curated `.agent/demos/` shelf (out of scope per this bead — "do not touch") would be a larger, separate change. This is a deliberate deviation from the bead's one-line "delete" suggestion; `scripts/` has no other content after this review. |
| `docs/repo-layout.md`, `.agent/README.md` | Updated | Kept in sync with the above: dropped the `systemd/` row, described `contrib/`'s actual remaining content, and corrected `.agent/README.md`'s "tracked shelves" claim (reports/archive are now gitignored evidence, not tracked). |

Root and `.agent` were swept for other tracked one-off scripts/reports/loose
files (`git ls-files` diffed against the disposition list above); nothing
else qualified. `.agent/tools/`, `.agent/scripts/` (bd-graph-lint,
bd-reimport-guard.py) are live, referenced tooling and were left alone, as
were `.beads/`, `.agent/handoffs/` (beyond the one line), `.agent/demos/`
(curated shelf), `tests/`, `polylogue/`, `docs/` per scope.

## Privacy pass

None of the untracked/deleted content contains credentials or third-party
personal data. `.agent/reports/`, `.agent/archive/`, and the scratch dump
are the operator's own research/verify-log/retired-scaffold notes; the
deleted systemd unit files are generic boilerplate. All of it remains
visible in prior git history regardless (no history rewrite performed, per
policy) — this PR only changes what's tracked going forward.

## Verification

- `devtools verify --quick` — exit 0, all 16 steps ok (ruff format/check,
  mypy --strict, `render all --check`, topology, layering, closure-matrix,
  schema roundtrip, manifests, ci-workflows, doc-commands, docs-coverage,
  test-infra-currency, test-clock-hygiene, pytest-timeout-overrides,
  degrade-loudly). No "out of sync" in the render-all output.
- `nix flake check --no-build` — `all checks passed!` (packages, devShells,
  checks, formatter, apps, nixosModules all evaluate clean; the pre-existing
  `apps.*.meta` warnings are unrelated to this change).
- Not run: full `devtools test`/`devtools verify --all` (no Python source
  changed, only docs/.gitignore/deletions of unreferenced files) and
  `nix build` of the package derivations (no build inputs changed; `nix
  flake check --no-build` already confirms the derivations still evaluate).

Ref polylogue-ocby


Sinity:
Re the P1 (bead-referenced reports untracked): this excision is an explicit operator directive — .agent/reports and .agent/archive are local agent workspaces, not product artifacts, and the public repo is not their custody layer. The files persist on the operator host; the two live-gate beads (polylogue-yla8, polylogue-hjpx.2) now carry notes stating where their packets live and what the actual closure gates are, so the durable task substrate is self-sufficient without the tracked copies. No further change planned for this finding.

Sinity:
Addressed the actionable P2 from review: CLAUDE.md and .agent/CONVENTIONS.md
still described the retired devloop packet as "archived at
.agent/archive/devloop-2026-07/ ... see its README" — both now-untracked
paths after the earlier commits in this branch. Pushed a follow-up commit
(7d07fd4cf) that:

- Rewords CLAUDE.md and .agent/CONVENTIONS.md to state the packet is retired
  outright, keeps the "do not resurrect devloop-* scripts" instruction, and
  drops the consult-its-README pointer.
- Fixes the same pattern everywhere else a tracked doc claimed
  `.agent/reports/` or `.agent/archive/` as tracked/consultable:
  `.agent/CONVENTIONS.md`'s directory-shape block, `.agent/README.md`
  (also caught `task-history/` mislabeled as a "tracked shelf" — it has zero
  commits in git history, always local-only), `.agent/scratch/README.md`,
  and `docs/design/README.md`.
- Left alone: tool/source strings that already correctly describe local-only
  output (e.g. `fanout_gen_prompts.py`'s "new file in your worktree only"),
  and test/source docstrings citing an audit doc as historical provenance
  (`claim_guard.py`, `tests/unit/{core,storage}/*`) — those cite evidence,
  they don't instruct a reader to navigate there.
- Not touching bead-note pointers to these reports, per the coordinator's
  split on the P1 item.

`devtools verify --quick` re-run clean: 16/16 steps, exit 0.

Ref polylogue-ocby

PR #3185, opened 2026-07-19
fix(test): repair lineage property oracle multi-hop cascade; exercise continuity cancellation

Sinity:
## Summary

Two independent trust-floor fixes from the 2026-07-19 AC-closure audit:

1. Repairs a multi-hop cascade bug in the write-path property state machine's own bookkeeping model (`tests/property/test_write_path_state_machine.py`) -- production was already correct.
2. Makes `devtools/continuity_replay.py` genuinely exercise and classify a real MCP mid-flight cancellation instead of hardcoding `cancellation_exercised: False`.

## Problem

**polylogue-866e**: `POLYLOGUE_HYPOTHESIS_REUSE_FAILURES=1 devtools test tests/property/test_write_path_state_machine.py` had passed repeatedly on master, but a fresh `--hypothesis-seed=577341254` run at the default 100-example profile produced a new falsifying sequence: `ingest_initial_parent`, `ingest_child_replaying_parent_prefix` x4 (which can chain into a grandchild via the shared `can_be_parent` selection), then `delete_parent_branch_point` -> `IndexError: tuple index out of range` inside the test's own `_assert_resolved_link`. Root-cause investigation (a direct-SQL + composed-read oracle replaying the exact call sequence outside pytest/Hypothesis) showed production's `read_archive_session_envelope` already recomposes the affected grandchild's transcript correctly; the bug was confined to the model's `delete_parent_branch_point` adjustment loop, which only compared each candidate's branch-point message id against the *directly deleted-from* session's positions and never cascaded the shift through a grandchild whose branch point lives inside an intermediate (already-shifted) session's own tail.

**polylogue-t8t**: `devtools/continuity_replay.py` declared `max_cancel_grace_ms` in `ContinuityBudget` and reported `cancellation_exercised` in every scenario's budget dict, but the value was a hardcoded `False` -- cancellation was scaffolded, never actually driven. `tests/integration/test_continuity_replay.py` asserted that `False` as the expected passing state, i.e. the test codified the gap rather than catching it.

## Solution

**866e** (`tests/property/test_write_path_state_machine.py`): replaced the direct SQL-lookup-based prefix adjustment with `WritePathStateMachine._cascaded_prefix_shift`, a recursive walk that resolves a branch point's shift against the cut root regardless of how many prefix-sharing hops separate them -- an index inside a session's borrowed prefix recurses into the parent's own composed transcript unchanged; an index inside the session's own tail shifts by whatever its borrowed segment shrank. Shifts are computed from a pre-mutation snapshot for every candidate before any `prefix_length` is mutated. Added `test_grandchild_transcript_recomposes_after_intermediate_ancestor_message_deletion` as a 4th named deterministic fixture (matching the file's existing two production-focused `test_` functions) that locks in the already-correct production behavior as a permanent regression guard.

**t8t** (`devtools/continuity_replay.py`): a naive single-call wall-clock race (create task, sleep briefly, send `notifications/cancelled`) was tried first and found genuinely flaky -- some scenarios' own real queries (a marker lookup with `limit=2`) complete in well under a millisecond end to end, so no fixed settle window reliably wins, while heavier queries reliably do. The shipped mechanism (`StdioMCPContinuityRoute.exercise_cancellation`) is deterministic instead: it issues `DEFAULT_CAPACITY + 4` concurrent copies of the scenario's own first real route step and sends a cancellation notification for every one. Regardless of how fast any individual copy's SQL work would finish, the copies that exceed the shared admission controller's ceiling are provably still queued (never touched SQLite) when the notifications arrive, and the admission wait loop checks `ctx.should_abort()` on its own poll cadence -- so at least one confirmed cancellation is guaranteed, not raced. Only tools independently confirmed to route through the interruptible archive-read path (`query`, `status`) are probed; `explain` (pure DSL grammar/capability introspection, no archive read to interrupt) is honestly reported `not_applicable` rather than forced -- probing it produced `completed_before_cancel` and occasional stdio connection instability, a genuine machinery gap, not a race to fix harder.

## Verification

- `devtools test tests/property/test_write_path_state_machine.py` -> 4 passed
- Direct non-Hypothesis reproduction script replaying the exact rule sequence confirmed `IndexError` before the fix and clean pass after, with the grandchild's shifted `prefix_length` landing at the expected value (2, not the stale 3)
- 400-example stateful stress run (`run_state_machine_as_test`, no pytest) and 11 distinct `--hypothesis-seed` runs at the default 100-example profile (including the original falsifying seed 577341254) all pass after the fix
- `devtools test tests/integration/test_continuity_replay.py tests/unit/product/test_continuity_scenarios.py tests/unit/mcp/test_prompt_query_parity.py` -> 22 passed (run twice for stability)
- `devtools test tests/unit/mcp/test_server_surfaces.py` -> 6 passed (no MCP surface regression)
- `mypy --strict` and `ruff format`/`check` clean on all touched files
- `devtools render all --check` -> exit 0, no "out of sync" surfaces
- `devtools verify --quick` -> exit 0 (all 16 steps green)

Ref polylogue-866e, polylogue-t8t

### DIFF
--- a/devtools/continuity_replay.py
+++ b/devtools/continuity_replay.py
@@ -18,7 +18,7 @@ import re
 import sys
 import time
 from collections.abc import Callable, Mapping, Sequence
-from contextlib import AsyncExitStack
+from contextlib import AsyncExitStack, suppress
 from dataclasses import dataclass, field
 from datetime import timedelta
 from pathlib import Path
@@ -27,6 +27,7 @@ from typing import TYPE_CHECKING, Literal, Protocol, TypeAlias, TypeGuard, cast
 if __package__ in {None, ""}:  # pragma: no cover - exercised by the script entry point
     sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
 
+from polylogue.archive.query.execution_control import DEFAULT_CAPACITY
 from polylogue.core.json import JSONDocument, JSONValue, require_json_document, require_json_value
 from polylogue.product.continuity_scenarios import (
     CONTINUITY_SCENARIOS,
@@ -50,6 +51,13 @@ RouteTransport: TypeAlias = Literal["stdio", "registered"]
 ArgumentMutator: TypeAlias = Callable[[str, RouteArguments, int], RouteArguments]
 ResponseMutator: TypeAlias = Callable[[str, RouteArguments, int, str], str]
 DiscoveryMutator: TypeAlias = Callable[[JSONDocument], JSONDocument]
+#: Concurrent copies of one real route step issued by the cancellation
+#: exercise (see StdioMCPContinuityRoute.exercise_cancellation). Comfortably
+#: above the production admission controller's default per-process capacity
+#: so at least one copy is provably still queued (never touched SQLite) when
+#: the cancellation notifications are sent, making the exercise deterministic
+#: rather than a race against how fast any one query happens to run.
+_CANCELLATION_PROBE_CONCURRENCY = DEFAULT_CAPACITY + 4
 _TEMPLATE_RE = re.compile(r"\{fixture:([A-Za-z0-9_.-]+)\}")
 _ATTEMPT_TOKEN_RE = re.compile(r"(?P<key>[a-z_]+):(?P<value>[A-Za-z0-9_-]+)")
 _ATTEMPT_GRADES: frozenset[str] = frozenset(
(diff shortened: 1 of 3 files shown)


Sinity:
CodeRabbit is rate-limited on this PR; merging on coordinator review instead (campaign precedent). Reviewed: 866e oracle fix is test-model-only with a deterministic repro + named regression fixture proving production recomposition correct; t8t cancellation exercise is deterministic by admission-ceiling saturation rather than wall-clock racing, with explain honestly not_applicable. 32 focused tests + quick gate green.

PR #3186, opened 2026-07-19
fix(storage): unify message-native-id normalization to stop FK failures

Sinity:
## Summary

Fixes a deterministic `sqlite3.IntegrityError` ("FOREIGN KEY constraint failed") in `_write_blocks` that killed the live v42 index rebuild (operation ab5bad1f) ~10 hours in, with zero context on which session caused it. Unifies message-native-id normalization into a single helper so the SQLite-generated `messages.message_id` column and the Python-side identity computation that feeds `blocks.message_id` can never disagree, and wraps the write path so a future divergence surfaces with session/logical-key context instead of a bare "FOREIGN KEY constraint failed".

## Problem

`_write_blocks` computes `blocks.message_id` via `_message_id()` (Python), which the SQLite `messages` table independently re-derives as a generated column (`session_id || ':' || COALESCE(native_id, position||'.'||variant_index)`). Before this change the two computations disagreed for pathological provider-native ids:

- **Whitespace-only ids** (e.g. `"  "`) stored truthy under `_write_messages`'s bare `_sqlite_text(...) or None`, but `identity_law.message_local_id`'s `native_id.strip()` check falls back to the position/variant component — two different message ids for the same message, so the `blocks` insert referenced a `message_id` the `messages` table never stored.
- **Non-empty but padded ids** (e.g. `"  msg-1  "`): `identity_law._required_text` strips the value, but the DB stored it raw/unstripped — a second, narrower divergence.
- **Lone-surrogate ids**: substituted to U+FFFD on the DB side (`_sqlite_text`) but not at all on the Python side, and duplicate-id detection counted *raw* ids, so two distinct raw ids that collapsed onto the same substituted form weren't flagged ambiguous — either an FK failure or a silent `INSERT OR REPLACE` row loss, depending on write order.

## Solution

- Added `_stored_message_native_id` (`polylogue/storage/sqlite/archive_tiers/write.py`) as the single source of truth: duplicate suppression → surrogate substitution (`_sqlite_text`) → the same strip/empty-to-`None` rule `identity_law` applies. Both `_write_messages`'s INSERT and `_message_id` (consumed by every message-id call site — blocks, web constructs, paste spans, parent links, session events, active-leaf resolution) now route through it.
- `_duplicate_message_native_ids` / `_effective_message_native_id` now operate on the same surrogate-substituted form, so two raw ids that collapse to the same normalized id are correctly treated as ambiguous (both fall back to position/variant) instead of silently colliding.
- **Twin-check**: grepped the async storage tree (`async_sqlite*.py`) for a second message/block writer — none exists; the daemon's "main process is the sole SQLite writer" invariant holds, so this single-writer fix is sufficient (no twin to patch).
- Wrapped `write_parsed_session_to_archive`'s write transaction so a `sqlite3.IntegrityError` is re-raised (chained via `from exc`, not swallowed) naming `session_id`/`origin`/`native_id`. Wrapped both `apply_raw_revision_replay` and `apply_raw_membership_classification` call sites in `backfill_historical_revision_evidence` (`polylogue/sources/revision_backfill.py`) with the `logical_key`, so this path's failures name what they were replaying.
- Did **not** change `core/identity_law.py` semantics — it remains the canonical law; the storage layer now conforms to it exactly.

## Verification

- `devtools test tests/unit/pipeline/test_archive_write.py tests/property/test_message_identity_normalization.py tests/unit/sources/test_revision_backfill.py` → 59 passed
- `devtools test tests/unit/storage/test_archive_tiers_write.py tests/property/test_write_path_state_machine.py tests/unit/storage/test_lineage_normalization.py tests/unit/storage/test_bulk_fts_prefix_reextract.py tests/unit/storage/test_bulk_build_lifecycle.py tests/unit/pipeline/test_resilience.py` → 133 passed
- `devtools test tests/unit/daemon/test_convergence_stages.py tests/unit/daemon/test_fts_readiness_fallback.py tests/unit/storage/test_repair.py tests/unit/storage/test_incremental_rebuild_equivalence.py` → 102 passed
- `devtools verify --quick` → ok (ruff format/check, mypy --strict, render all --check, topology/layering/closure-matrix/manifests/ci-workflows/docs — all green)
- **Anti-vacuity**: reverted the `_write_messages` fix locally and reran the new property test — it failed with the exact `sqlite3.IntegrityError: FOREIGN KEY constraint failed writing session_id='unknown-export:identity-law-sweep' ... native_id='identity-law-sweep'` signature on a whitespace-only native id (`native_ids=[' ']`), proving both the regression test and the new loud-failure-context wrapper actually exercise the bug; then restored the fix and reconfirmed green.
- Note: the task brief mentioned `tests/unit/storage/test_no_string_interpolated_sql.py` as a hardcoded-identifier allowlist gate to run explicitly if f-string SQL was touched — that file does not exist in this checkout (searched, no match), and this change adds no SQL-string interpolation (only Python exception-message f-strings), so it is not applicable here.

Unblocks resuming index rebuild operation ab5bad1f from its 31,882-session checkpoint.

Ref polylogue rebuild ab5bad1f.

### DIFF
--- a/polylogue/sources/revision_backfill.py
+++ b/polylogue/sources/revision_backfill.py
@@ -719,14 +719,20 @@ def backfill_historical_revision_evidence(
                     archive.release_provisional_full_revisions(sorted(plan_raw_ids))
                 adoption_deferred += len(plan.accepted_raw_ids)
                 continue
-            archive.apply_raw_revision_replay(
-                plan,
-                parsed_by_raw_id,
-                acquired_at_ms=0,
-                manage_transaction=not replay_batched,
-                bulk_fts=bulk_fts,
-                bulk_build=bulk_build,
-            )
+            try:
+                archive.apply_raw_revision_replay(
+                    plan,
+                    parsed_by_raw_id,
+                    acquired_at_ms=0,
+                    manage_transaction=not replay_batched,
+                    bulk_fts=bulk_fts,
+                    bulk_build=bulk_build,
+                )
+            except sqlite3.IntegrityError as exc:
+                raise sqlite3.IntegrityError(
+                    f"backfill_historical_revision_evidence: byte-proven replay failed for "
+                    f"logical_key={plan.logical_source_key!r}: {exc}"
+                ) from exc
             replayed += 1
             byte_replayed_keys.add(logical_key)
             if replay_batched:
@@ -766,16 +772,22 @@ def backfill_historical_revision_evidence(
                 )
                 adoption_deferred += len(classification.accepted_raw_ids)
                 continue
-            archive.apply_raw_membership_classification(
-                logical_key,
-                classification,
-                member_sessions,
-                projections,
-                acquired_at_ms=0,
-                manage_transaction=not replay_batched,
-                bulk_fts=bulk_fts,
-                bulk_build=bulk_build,
-            )
+            try:
+                archive.apply_raw_membership_classification(
+                    logical_key,
+                    classification,
+                    member_sessions,
+                    projections,
+                    acquired_at_ms=0,
+                    manage_transaction=not replay_batched,
+                    bulk_fts=bulk_fts,
+                    bulk_build=bulk_build,
+                )
+            except sqlite3.IntegrityError as exc:
+                raise sqlite3.IntegrityError(
+                    f"backfill_historical_revision_evidence: membership replay failed for "
+                    f"logical_key={logical_key!r}: {exc}"
+                ) from exc
             if classification.accepted_raw_ids:
                 replayed += 1
             if replay_batched:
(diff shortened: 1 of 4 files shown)


Sinity:
CodeRabbit is rate-limited on this PR; merging on coordinator review instead (campaign precedent #3174/#3175). Reviewed: single-source normalization helper routed through both the stored column and every Python-side id consumer; loud IntegrityError context on write + both replay call sites; property sweep asserting DB-generated == Python-computed message_id over hostile native ids; anti-vacuity confirmed (revert reproduces the exact live FK signature). 294 focused tests + quick gate green. This unblocks resume of rebuild operation ab5bad1f.

PR #3187, opened 2026-07-20
fix(daemon): derive hook spool from the resolved archive root

Sinity:
## Summary

The hook-event sidecar/spool directory now derives from the resolved
archive root (`<archive_root>/hooks`) instead of a bare `XDG_DATA_HOME`
path, so a daemon pointed at a scratch or test archive genuinely isolates
its hook spool from the real one.

## Problem

`hooks_sidecar_dir()` (`polylogue/paths/_roots.py`) resolved
`data_home()/"hooks"` — pure XDG, independent of the archive root. Every
consumer that reaches it ambiently (the daemon watcher's
`default_sources()`, `sources/hooks.py`'s drain path, `hook_paste_enrichment.py`)
therefore drained the *one* real global spool regardless of `--root` or
`POLYLOGUE_ARCHIVE_ROOT`. A daemon started against a scratch archive moved
real personal hook events pending → acknowledged into a throwaway archive.
This has bitten the operator 4+ times (most recently: a lane drained 6262
real hook events into a scratch archive). The `POLYLOGUE_HOOK_SIDECAR_DIR`
env override was a second, hook-specific knob operators had to remember on
top of `POLYLOGUE_ARCHIVE_ROOT`, and repeatedly didn't — exactly the
manual-escape-hatch pattern the automagic doctrine purges.

A second, independent instance of the same bug existed in
`config.py`'s `resolve_runtime_config()`: its `hook_sidecar_dir` captured
default was a hardcoded `data_home()/"hooks"` string (unlike
`browser_capture_spool_path`'s blank default), so the archive-derived
fallback there could never actually trigger.

## Solution

- `hooks_sidecar_dir()` now resolves `archive_root() / "hooks"`, mirroring
  `browser_capture_spool_root()`. Byte-identical to the previous default
  for the default production archive root (archive root IS `data_home()`
  when `POLYLOGUE_ARCHIVE_ROOT` is unset) — zero migration. This single
  fix cascades to every ambient consumer: `sources/hooks.py`'s
  `hook_spool_root()`, the daemon watcher's
  `default_sources()`/`pending_hook_spool_dir()` chain, and
  `cli/commands/init.py`'s path table.
- `config.py`: fixed the `hook_sidecar_dir` captured default to blank
  (matching `browser_capture_spool_path`) and the `resolve_runtime_config()`
  fallback to `archive / "hooks"`, so the archive-derived fallback there
  actually takes effect.
- `hook_paste_enrichment.enrich_paste_from_hooks()` now derives the hooks
  dir from its own `db_path.parent` instead of an ambient global, so it
  always inspects the archive it was actually called for.
- Removed the ad hoc `POLYLOGUE_HOOK_SIDECAR_DIR` `os.environ.get()` reads
  scattered across `sources/hooks.py` and `hooks/__init__.py`. The
  config-integrated `hook_sidecar_dir` key in `config.py` remains for the
  rare case the spool genuinely needs to live somewhere other than
  `<archive_root>/hooks`.
- Hook **writer** scripts (`contrib/polylogue-hook`, `packaging/polylogue-hooks`,
  and the bundled package's `hook_main`) run inside agent hook subprocesses
  whose environment cannot be trusted to carry `POLYLOGUE_ARCHIVE_ROOT`.
  `polylogue hooks install` now resolves the current spool path once, at
  install time, and bakes it into each rendered command via a new
  `--sidecar-dir PATH` flag (parsed by all three writer implementations).
  Without that flag, the standalone writers fall back through
  `POLYLOGUE_ARCHIVE_ROOT` — not a separate knob — to the XDG default.
- `docs/hooks.md` and `docs/configuration.md` updated to match (including
  fixing a pre-existing stale `polylogue.toml` example that referenced a
  nonexistent `[hooks]`/`enabled` shape instead of the real
  `sources.hook_sidecar_dir` TOML path). The `docs/hermes-operators.md`
  "Try it yourself" warning that hook-spool drain ignores `--root`/the
  archive root is removed now that it's no longer true.
- `.github/workflows/release.yml`'s `polylogue-hooks` smoke test updated to
  the new env var, and (a latent, unrelated drift this touched while
  editing the same lines) fixed to assert against the pending spool file
  the standalone package actually writes, rather than a legacy per-session
  journal it never wrote.

## Verification

- `devtools test tests/unit/sources/test_hook_spool.py tests/unit/sources/test_hook_paste_enrichment.py tests/unit/cli/test_hooks.py tests/unit/daemon/test_hook_liveness.py tests/unit/core/test_paths.py` → `66 passed`
- `devtools verify --quick` → all 16 steps `ok` (ruff format/check, mypy
  --strict, `render all --check`, topology/layering/closure-matrix,
  manifests, ci-workflows, doc-commands, docs-coverage,
  test-infra-currency, test-clock-hygiene, pytest-timeout-overrides,
  degrade-loudly)
- Manual: exercised `contrib/polylogue-hook` directly with both the
  `POLYLOGUE_ARCHIVE_ROOT` fallback and an explicit `--sidecar-dir`;
  confirmed pending envelopes land under the expected archive-scoped path
  in each case, and that a seeded decoy event at the real XDG location is
  never touched by a scratch-scoped resolution (also covered by new unit
  tests: `test_hooks_sidecar_dir_isolates_a_scratch_archive_root_with_no_xdg_leakage`,
  `test_hook_paste_enrichment_never_reads_a_sibling_archives_hooks_dir`,
  `test_install_bakes_the_current_archive_roots_resolved_spool_path`).
- Also ran a full `devtools verify --seed-testmon --skip-slow` as a
  diagnostic; it surfaced ~100 unrelated failures across modules my diff
  never touches, root-caused to a missing optional `msgspec` dependency in
  this worktree's `.venv` (confirmed by re-running
  `tests/unit/core/test_json.py -k msgspec` in isolation, same error) —
  pre-existing environment drift in this specific worktree checkout, not
  caused by this change.

Ref polylogue-o7hx

## Summary by CodeRabbit

* **New Features**
  * Added an explicit `--sidecar-dir` option for hook commands.
  * Hook spool storage now automatically follows the configured archive location.
  * Installed hook commands include their resolved spool path for consistent event handling.

* **Bug Fixes**
  * Improved isolation between archives, preventing hook events from leaking across archive locations.
  * Updated hook event processing to reliably use the correct archive-specific spool.

* **Documentation**
  * Clarified hook spool defaults, archive-root behavior, installation requirements, and configuration guidance.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -264,13 +264,16 @@ jobs:
         run: |
           # shellcheck disable=SC1091
           source .venv-hooks/bin/activate
-          sidecar="${RUNNER_TEMP}/polylogue-hooks-sidecar"
-          mkdir -p "${sidecar}"
-          export POLYLOGUE_HOOK_SIDECAR_DIR="${sidecar}"
-          # Supported event must succeed and write a JSONL record.
+          # polylogue-hooks derives its spool from the resolved archive root
+          # (polylogue-o7hx), not a separate hook-specific env override.
+          archive_root="${RUNNER_TEMP}/polylogue-hooks-archive"
+          export POLYLOGUE_ARCHIVE_ROOT="${archive_root}"
+          sidecar="${archive_root}/hooks"
+          # Supported event must succeed and write one pending spool envelope.
           printf '{"session_id":"abc123","model":"sonnet"}' | polylogue-hook SessionStart
-          test -f "${sidecar}/claude-code-abc123.jsonl"
-          grep -q '"event_type": "SessionStart"' "${sidecar}/claude-code-abc123.jsonl"
+          pending=$(find "${sidecar}/pending" -maxdepth 1 -type f -name '*.json')
+          test -n "${pending}"
+          grep -q '"event_type":"SessionStart"' "${pending}"
           # Unsupported event must exit non-zero without crashing the harness.
           if printf '{}' | polylogue-hook NotARealEvent; then
             echo "polylogue-hook should have rejected an unknown event type" >&2
--- a/contrib/polylogue-hook
+++ b/contrib/polylogue-hook
@@ -7,11 +7,22 @@
 # pending spool where the daemon persists and acknowledges it.
 #
 # Usage:
-#   polylogue-hook <event-type> [--provider claude-code|codex|hermes]
+#   polylogue-hook <event-type> [--provider claude-code|codex|hermes] [--sidecar-dir PATH]
+#
+# --sidecar-dir is the concrete resolved spool path baked in by
+# `polylogue hooks install` at install time (polylogue-o7hx): a hook
+# subprocess's environment cannot be trusted to carry
+# POLYLOGUE_ARCHIVE_ROOT (agent harnesses invoke hooks with their own,
+# possibly minimal, environment), so ambient env resolution alone is not
+# reliable here. Without --sidecar-dir, falls back to
+# ${POLYLOGUE_ARCHIVE_ROOT:+$POLYLOGUE_ARCHIVE_ROOT/hooks} or the XDG
+# default (byte-identical to the default archive root's hooks directory).
 #
 # Environment variables:
-#   POLYLOGUE_HOOK_SIDECAR_DIR  Override default sidecar directory
-#   POLYLOGUE_HOOK_PROVIDER      Force provider (claude-code | codex | hermes)
+#   POLYLOGUE_ARCHIVE_ROOT  Overrides the archive root (and therefore the
+#                           default sidecar directory) when --sidecar-dir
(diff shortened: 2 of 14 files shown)


Sinity:
CodeRabbit rate-limited; merging on coordinator review (campaign precedent). Reviewed: single-point derivation in hooks_sidecar_dir → archive_root/hooks (mirrors browser_capture_spool_root), parallel captured-default fix in config.py resolution, POLYLOGUE_HOOK_SIDECAR_DIR override removed with config key as sole override, writer scripts bake --sidecar-dir at install. 66 focused tests + quick gate green. Deploy consequence recorded on polylogue-dcz5: re-render installed hook commands when the sinnix pin bumps.

PR #3189, opened 2026-07-20
feat(daemon): route bulk-scale rebuild through daemon-owned generation build

Sinity:
## Summary

Implements polylogue-gd6v (m6tp phase (c)): the daemon now routes a
bulk-scale raw backlog into a resumable, transactional, blue-green index
generation build running in-process, reusing the SAME engine the offline
`polylogue ops maintenance rebuild-index` CLI command already drives — no
duplicate implementation. Off by default behind a new
`daemon_bulk_rebuild_routing` config flag.

## Problem

The trickle raw-materialization conveyor is sized for steady-state drift;
a bulk-scale backlog turns it into a weeks-scale grind (#3145's own
threshold-warning evidence). The correct bulk path already existed as the
offline CLI command (resumable transaction, blue-green generation, one
census+replay sweep), but nothing routed to it automatically, and its own
resume model silently re-walks the whole corpus when a caller omits
`--operation-id` (polylogue-fbte: measured ~2.25h pure re-verification per
resume on a ~50K-raw archive, observed on operation `ab5bad1f`).

## Solution

- `polylogue/daemon/bulk_rebuild.py` (new): resolves/resumes a single
  well-known bulk-rebuild transaction per archive
  (`DAEMON_BULK_REBUILD_OPERATION_ID`), drives it through
  `rebuild_index_from_source_sync`, and retires a terminal
  (promoted/stale/failed) transaction so the id is reused for the next
  backlog. The daemon never has an operator-supplied operation id to
  forget, so the fbte bug class is structurally unreachable on this path.
- `polylogue/daemon/cli.py`: `_maybe_route_daemon_bulk_rebuild`, called
  from the existing raw-materialization convergence tick right after the
  #3145 recommendation log line. Keeps driving an in-flight build every
  tick even if the instantaneous backlog reading dips below threshold, so
  partial progress is never abandoned. A dedicated `DaemonParseStage`
  singleton (separate pool from the trickle conveyor's own) shuts down at
  daemon teardown alongside the existing one.
- `polylogue/daemon/parse_prefetch.py`: `DaemonParseStage.warm_raw_ids`
  extracts `warm`'s dispatch/admit logic to also accept an explicit raw-id
  list — the bulk driver already knows its next page's raw ids from the
  transaction's own paged cursor, so it pre-warms them in the #3168
  thread pool before ever requesting the writer hold.
- `polylogue/sources/revision_backfill.py`, `polylogue/maintenance/replay.py`,
  `polylogue/maintenance/rebuild_index.py`: thread a new `prefetch_cache`
  parameter from `RebuildIndexRequest` down through
  `backfill_historical_revision_evidence`'s census phase, mirroring the
  #3168 raw-materialization wiring. A prefetch hit flows through the same
  `spill.add(...)` as a fresh parse, so the replay phase's own
  `spill.for_raw` lookups see identical warmed content — census-phase
  prefetching alone is enough to also skip replay-phase reparsing.
  Degrades gracefully to the unmodified sequential in-hold parse on any
  miss or on a GIL build (real parallel-parse speedup lands with the
  3.14t deploy, m6tp phase (b); this PR is orchestration only).
- `polylogue/storage/index_generation.py`:
  `IndexGenerationStore.discard_transaction` retires a terminal
  transaction record so its operation id can be reused.
- `polylogue/config.py`, `docs/configuration.md`:
  `daemon_bulk_rebuild_routing` config flag (off by default), same
  pattern as `daemon_parse_stage_split`.

**Non-obvious decisions:**

- **Single-writer invariant.** The bead's design note mentions "a second
  writer connection on the generation's own index.db." This PR's explicit
  task constraints restate the single-writer invariant more strongly (main
  process is the sole SQLite writer; parse workers only produce parsed
  payloads, never write), so every bulk-rebuild write — like every other
  daemon writer actor — is scheduled through the existing
  `daemon_write_coordinator().run_sync`. Responsiveness instead comes from
  moving parse off that hold (mirroring phase (a)'s
  `daemon_parse_stage_split`), not a second concurrent writer thread.
- **No trickle suppression.** Bulk routing does not pause or coordinate
  with the trickle conveyor while a build is in flight — both converge to
  the same eventual state (the bulk build reads a snapshot and replays
  into a separate generation file; promotion is atomic). The resulting
  redundant per-raw work during the backlog window is a known,
  deliberately out-of-scope efficiency gap — m6tp's own design sketch
  flags watcher-pause/frozen-snapshot semantics as still-open questions
  this PR does not attempt to resolve.
- **CLI command NOT deleted.** `ops maintenance rebuild-index` remains;
  deletion is gated on an archive-scale equivalence receipt
  (coordinator-owned, post-promote) per the gd6v AC. This PR delivers the
  fixture-scale half of that gate.

## Verification

```
devtools test tests/unit/daemon/test_bulk_rebuild.py tests/unit/daemon/test_daemon_cli.py \
  tests/unit/daemon/test_parse_prefetch.py tests/unit/daemon/test_raw_materialization_parse_stage_equivalence.py \
  tests/unit/maintenance/test_rebuild_index_bulk_build.py tests/unit/storage/test_incremental_rebuild_equivalence.py \
  tests/unit/storage/test_index_generation.py tests/unit/sources/test_revision_backfill.py
=> 176 passed in 8.58s

uv run mypy polylogue --active
=> Success: no issues found in 1059 source files

devtools verify --quick
=> exit_code 0 (ruff format/check, mypy, render all --check, layering,
   topology, docs-coverage, closure-matrix, all clean; grepped for
   "out of sync" — none found)

devtools render topology-projection && devtools render topology-status
=> regenerated docs/plans/topology-target.yaml + docs/topology-status.md
   for the new polylogue/daemon/bulk_rebuild.py module
```

New fixture-scale tests (`tests/unit/daemon/test_bulk_rebuild.py`):

- `test_daemon_bulk_rebuild_equivalent_to_cli_rebuild` — builds the same
  6-raw corpus through the CLI path (single call, `prefetch_cache=None`)
  and the daemon path (multi-pass, `DaemonParseStage`-warmed
  `prefetch_cache`), asserts identical `sessions`/`messages`/`blocks`/
  `session_links` row-level content, `messages_fts` row count, and session
  count.
- `test_daemon_bulk_rebuild_pass_resumes_without_reprocessing_raw_ids` +
  `test_daemon_bulk_rebuild_pass_next_page_excludes_already_scheduled_raws`
  — the **fbte requirement**: each bounded pass's transaction cursor
  (`last_raw_id`/`processed_raw_count`, written by the existing
  `IndexGenerationStore.checkpoint_transaction`) only ever advances; a
  later page is proven disjoint from every earlier page; a simulated
  restart (fresh `DaemonParseStage` per pass, mirroring a fresh process)
  resumes from the persisted cursor instead of re-walking the corpus.
- `test_resolve_or_start_creates_resumes_and_retires_transaction` —
  transaction lifecycle: create, resume-unchanged, retire-on-terminal,
  fresh restart at the same well-known operation id.
- `tests/unit/daemon/test_daemon_cli.py`: 4 new tests for the routing gate
  (flag off never checks/drives; below-threshold+not-resumable is a
  no-op; an in-flight operation keeps being driven below threshold; a
  pass failure never propagates to the caller).

## Remaining gd6v scope (follow-up, coordinator-owned)

- Archive-scale equivalence receipt and the resulting CLI-command deletion
  (polylogue-4jsk).
- agvo responsiveness p99 gate proving live reads/ingest stay responsive
  during a real drain (this PR's responsiveness claim rests on parse
  moving off the writer hold, same mechanism as phase (a); not
  independently measured here).
- Trickle-conveyor coordination/suppression during an in-flight bulk build
  (explicitly out of scope above).

Ref polylogue-gd6v


Sinity:
CodeRabbit rate-limited (no inline findings); merging on coordinator review (campaign precedent). Reviewed: single-writer preserved via daemon_write_coordinator scheduling; parse-prefetch reuse with graceful sequential fallback; well-known operation id + per-pass cursor checkpoint makes the fbte O(corpus)-resume class structurally unreachable; fixture equivalence vs CLI path asserted on sessions/messages/blocks/links/FTS. CLI deletion correctly gated on the archive-scale receipt (4jsk). 176 focused tests + mypy strict + quick gate green.

PR #3191, opened 2026-07-20
fix(sources): prevent sibling-variant continuation coordinate collisions

Sinity:
## Summary

Fixes the root cause of the v42 index rebuild failure (operation ab5bad1f, most recently at `claude-ai-export:d24081f5`): the Claude web/browser parser could assign two distinct messages the same `(position, variant_index)` pair, which the messages table's `PRIMARY KEY(session_id, position, variant_index)` silently resolves by dropping one row via `INSERT OR REPLACE` — while its blocks are still written against its own (now-orphaned) `message_id`, surfacing as an unrelated foreign-key failure far from the actual cause. This PR fixes the parser-level cause and adds a writer-level safety net that turns any future occurrence of this class of bug into a loud, diagnosable error instead of silent data loss.

## Problem

`normalize_chat_messages` (`polylogue/sources/parsers/claude/common.py`) computed `variant_index_by_id` as `branch_index_by_id` — a message's rank among its own siblings under the same parent — independent of which sibling-variant "world" its parent belonged to. When two sibling variants at the same tree depth each had their own single-child continuation, both continuations were the sole (rank-0) child of their respective parent, so both independently reset to `variant_index=0`, colliding on the same `(position, variant_index)` coordinate their parents correctly diverged on.

Concretely, for a fixture shaped like:

```
root -> chain-1 -> chain-2 -> branch-point -> {variant-a, variant-b}  (correct: position 4, variant 0/1)
                                                  |            |
                                          continuation-a  continuation-b  (bug: both position 5, variant 0)
```

`_write_messages`' `INSERT OR REPLACE` silently dropped one of `continuation-a`/`continuation-b`, while `_write_blocks` still computed a distinct Python-side `message_id` for the dropped message and inserted its blocks against it — a foreign key referencing a row that was never (or no longer) there.

## Solution

**Parser fix** (`polylogue/sources/parsers/claude/common.py`):
- `_resolve_variant_index`: a rank-0 ("primary") child now *inherits* its parent's resolved `variant_index` instead of resetting to 0 — the continuation of variant 1 lives in variant 1's world. This walks up through chains of rank-0 descendants (composing through nested branch-under-branch shapes — a variant's variant) until it hits an explicit/nonzero variant, the root, or a cycle (degrading to 0, matching the existing `_lineage_depths` cycle-degrade pattern). Explicit provider-supplied variant indexes and nonzero sibling rank are unaffected — those are real branch points.
- `_deduplicate_variant_collisions`: a final deterministic safety net applied regardless of mode (flat or tree). No input tree shape may leave two emitted messages sharing `(position, variant_index)`; any residual collision is resolved by renumbering the colliding position group in `order_key` order (current variant, timestamp, provider message id).

**Other-parser audit** (grepped every parser under `sources/parsers/` for the same "rank among siblings, no uniqueness proof" pattern):
- `chatgpt.py` computes a similarly-shaped per-parent `branch_index`/`variant_index`, but its `position` is a monotonic enumeration index over every node in the conversation mapping (`position = idx - 1`), unique per emitted message *by construction* regardless of `variant_index` — **not vulnerable**.
- `claude/ai_parser.py` and `claude/code_parser.py` route their real parse path through this same `normalize_chat_messages()` — they inherit the fix directly; their standalone `variant_index=0` literals are for degenerate single-message cases only.
- All other parsers (`codex.py`, `drive.py` [Gemini/AI Studio], `local_agent.py` [gemini-cli], `antigravity.py`, `browser_capture.py`, `hermes_*.py`, `beads.py`) hardcode `variant_index=0` unconditionally — no branching modeled, **not vulnerable** to this shape.

**Writer safety net** (`polylogue/storage/sqlite/archive_tiers/write.py`):
- `_assert_unique_message_coordinates` computes the exact effective `(position, variant_index)` `_write_messages` uses to build INSERT rows (`position_offset + position-or-fallback-index`, `variant_index-or-0`) and raises a `ValueError` naming the session and every colliding native id if any pair repeats. Wired into both write paths: the full-replace path (`_replace_full_session_messages_and_blocks`, used for from-scratch imports — the path that hit this in production) before any row is touched, and the merge-append path right after `position_offset` is computed from the live archive state. This is independent of the parser fix — it's the writer's own guard against whatever any parser (present or future) produces.

## Verification

```
devtools test tests/unit/sources/test_claude_web_normalization.py \
  tests/unit/pipeline/test_archive_write.py \
  tests/property/test_claude_variant_position_uniqueness.py \
  tests/property/test_message_identity_normalization.py
  -> 32 passed

mypy --strict (touched files)          -> no issues
devtools verify --quick                -> exit 0 (format, lint, mypy, render-all-check,
                                           topology, layering, closure-matrix, schema
                                           roundtrip, manifests, ci-workflows, doc-commands,
                                           docs-coverage, test-infra-currency,
                                           test-clock-hygiene, pytest-timeout-overrides,
                                           degrade-loudly)
```

Test coverage added, per the four acceptance criteria this fix was scoped against:
- **(a) Regression fixture** (`tests/unit/sources/test_claude_web_normalization.py`): the exact 8-message collision shape above, run through the real `claude-ai` provider detection + parse + `write_parsed_session_to_archive` into a fresh archive store. Asserts both continuation messages, and all their blocks, are present at distinct `(position, variant_index)` coordinates. A pre-existing test in the same file (`test_claude_normalization_survives_archive_write_and_hydration`) asserted `u2`'s `variant_index == 0`; under the corrected inheritance semantics `u2` (the sole child of variant-1 sibling `a-new`) now correctly inherits `variant_index == 1` — updated with an explanatory comment, not silently changed.
- **(b) Property test** (`tests/property/test_claude_variant_position_uniqueness.py`, new file): Hypothesis-generated random parent/branch/depth trees (including nested branch-under-branch shapes and None-timestamp paths) asserting `normalize_chat_messages` output always has unique `(position, variant_index)` per session. Verified this test's minimal shrunk counterexample against the pre-fix parser reproduces exactly the collision class described above (`(2, 0)` produced twice from a two-sibling/two-grandchild tree), then passes cleanly against the fix.
- **(c) Writer collision guard** (`tests/unit/pipeline/test_archive_write.py`): two new tests — a hand-built `ParsedSession` with a forced `(position, variant_index)` collision raises a loud `ValueError` naming the session and both colliding native ids (not a bare `sqlite3.IntegrityError`) on both the full-replace and merge-append write paths.
- **(d) #3186 identity-law property test** (`tests/property/test_message_identity_normalization.py`, pre-existing, unmodified): still green — confirmed by direct run, both id-law classes (DB-generated `message_id` vs. Python identity computation, and now position/variant coordinate uniqueness) are covered by non-overlapping property tests.

**Blast radius**: not independently measured by this lane — this PR does not have real corpus dry-run scan numbers to report (an earlier draft of this description asserted a specific session count/percentage; that figure was not backed by an actual scan run in this session and has been removed as unverified). The coordinator's own corpus dry-run scan, run before resuming operation ab5bad1f, is the authoritative source for how many sessions this collision shape actually affects.

**Unrelated pre-existing failures observed and ruled out**: a broad `pytest tests/unit/sources tests/unit/storage tests/unit/pipeline tests/property` run surfaced 31 failures across files this change never touches (durable-migration schema counts, FTS derived-surface guard table, `Mock` attribute drift in `test_parsing_service.py`, browser-capture/GDPR title-coalesce ordering, retrieval-readiness `KeyError`s, etc.). Spot-checked several against a byte-for-byte revert of both changed files on the same worktree — they fail identically pre-fix, confirming baseline drift unrelated to this PR (not re-litigated here; the repo's own `devtools test <files>` convention against testmon-affected selection is what this PR's verification follows).

Ref: unblocks resume of operation ab5bad1f. The coordinator will run a corpus dry-run scan before resuming the rebuild.

### DIFF
--- a/polylogue/sources/parsers/claude/common.py
+++ b/polylogue/sources/parsers/claude/common.py
@@ -473,6 +473,102 @@ def _sibling_sort_key(evidence: _ClaudeMessageEvidence) -> tuple[int, int, float
     )
 
 
+def _resolve_variant_index(
+    start_id: str,
+    evidence_by_id: Mapping[str, _ClaudeMessageEvidence],
+    branch_index_by_id: Mapping[str, int],
+    resolved: dict[str, int],
+) -> int:
+    """Resolve a tree-mode message's variant index, composing rank-0 inheritance.
+
+    A message with an explicit provider variant index, or a nonzero rank among
+    its siblings, keeps that value -- those are real branch points. A rank-0
+    ("primary") child is just the continuation of whichever variant its parent
+    belongs to, so it inherits the parent's resolved variant index rather than
+    resetting to 0. This composes through chains of rank-0 descendants (a
+    rank-0 child of a rank-0 child of a variant sibling still inherits that
+    variant), walking up until an explicit/nonzero variant, the root, or a
+    cycle is found.
+
+    Without this, two sibling variants at the same depth each contribute a
+    rank-0 continuation at the *same* (position, variant_index=0) coordinate --
+    the exact collision shape that silently drops a message via
+    `INSERT OR REPLACE` on the messages table's
+    ``PRIMARY KEY(session_id, position, variant_index)``.
+
+    Cycles degrade to variant 0 (mirrors `_lineage_depths`' cycle handling)
+    instead of looping forever; the final uniqueness pass below is the safety
+    net if that ever produces a residual collision anyway.
+    """
+    chain: list[str] = []
+    cursor: str | None = start_id
+    seen: set[str] = set()
+    value = 0
+    while cursor is not None:
+        if cursor in resolved:
+            value = resolved[cursor]
+            break
+        if cursor in seen or cursor not in evidence_by_id:
+            value = 0
+            break
+        seen.add(cursor)
+        evidence = evidence_by_id[cursor]
+        if evidence.explicit_variant_index is not None:
+            value = evidence.explicit_variant_index
+            resolved[cursor] = value
+            break
+        rank = branch_index_by_id.get(cursor, 0)
+        if rank != 0:
+            value = rank
+            resolved[cursor] = value
+            break
+        chain.append(cursor)
+        cursor = evidence.parent_message_provider_id
+    for message_id in chain:
+        resolved[message_id] = value
+    return value
+
+
+def _deduplicate_variant_collisions(
+    emitted: list[_ClaudeMessageEvidence],
+    position_by_id: Mapping[str, int],
+    variant_index_by_id: dict[str, int],
+) -> None:
+    """Guarantee (position, variant_index) uniqueness per session -- the safety net.
+
+    Parser-level variant assignment (explicit provider values, sibling rank, or
+    rank-0 inheritance) is a heuristic and cannot be proven collision-free for
+    every exotic input tree shape. No two emitted messages may share
+    (position, variant_index): the messages table is unique on exactly that
+    pair, so a silent collision here becomes a silently dropped message
+    downstream (`INSERT OR REPLACE`) whose blocks then orphan against a
+    foreign key that no longer resolves.
+
+    Mutates ``variant_index_by_id`` in place. Positions with no collision are
+    left untouched. A colliding position group is fully renumbered in
+    order_key order (current variant index, then timestamp, then provider
+    message id) so the result is deterministic regardless of which heuristic
+    produced the original assignment.
+    """
+    by_position: dict[int, list[_ClaudeMessageEvidence]] = defaultdict(list)
+    for evidence in emitted:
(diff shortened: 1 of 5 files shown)


Sinity:
Restoring the blast-radius figure the lane removed as unverifiable from its side: it was not hallucinated — it came from the coordinator corpus dry-run scan (2026-07-20 ~05:20, read-only parse of all 12,078 web-export raws / 12,049 sessions against the live source tier, scanner + results at /realm/tmp/idlaw-scan-results.txt on the operator host): exactly 466 sessions flagged, ALL of them this (position,variant) collision class, zero remaining #3186-class id-law mismatches. ~3.9% of web-export sessions. The same scan re-runs against this fix before the rebuild resumes; expected result is zero findings, and the 466 session ids serve as the post-promote verification set.

PR #3192, opened 2026-07-20
feat(storage): content-address the embeddings tier by input hash

Sinity:
## Summary

Content-addresses the embeddings tier: vectors are keyed by
`embedding_input_hash = SHA-256(model, normalized embedder input text)`
instead of message identity, so an index rebuild or lineage-normalization
shift can never invalidate a vector whose underlying text is unchanged.
This PR finishes a lane that died at the finish line (implementation done,
full-repo mypy clean, about to run final verification) — its entire diff
was rescued as an unverified WIP commit and is being landed here after a
full review, cleanup, and verification pass.

## Problem

`message_embeddings_meta` bound vectors to `messages.content_hash`, which
INCLUDES `session_id`/`position`/`variant_index` — identity-contaminated by
design (it drives row-level re-ingest/dedup change detection for the
`messages` table itself). An index rebuild or lineage-normalization shift
renumbers messages without changing their text, so this identity-bound
freshness gate invalidated vectors it shouldn't have — the root cause of
the 777K-vector rescue (polylogue-04kl). Vectors are about content, not
transient index identity (operator ruling 2026-07-20).

## Solution

- `embedding_input_hash(model, input_text)` (`storage/embeddings/identity.py`):
  identity-free SHA-256, same philosophy as the svfj block evidence hash.
  Derived directly from the exact string handed to the embedder (never
  re-derived from stored identity), so hash validity equals vector validity
  by construction.
- `embeddings.db` schema v3 → v4 (derived-tier DDL edit, no migration chain,
  per the repo's schema regime): `message_embeddings` (vec0) and
  `message_embeddings_meta` are now keyed by `embedding_input_hash` and are
  write-once/shared — a hash's presence in `message_embeddings_meta` IS
  freshness, there is no per-vector `needs_reindex` state anymore. New
  `message_embedding_refs` table is the rebuildable
  `message_id -> embedding_input_hash` mapping (lives in `embeddings.db`,
  not `index.db`, so this never bumps `INDEX_SCHEMA_VERSION`); it is
  rewritten per session re-embed while vector/meta rows are never deleted
  by ordinary re-embed.
- **Dedup for free**: identical embedder input text (fork/resume/
  auto-compaction replays, or genuinely coincidental duplicate prose) is
  stored exactly once regardless of how many messages/sessions reference it
  — a real Voyage API cost saving in a lineage-heavy archive.
- Every consumer updated to resolve through `message_embedding_refs` in
  both the sync and async trees: freshness predicate
  (`materialization.py`), reconcile (ref-scoped orphan cleanup only — the
  content-addressed vector/meta GC is explicitly deferred, documented as
  follow-up debt, not silently dropped), excision (reference-counted vector
  GC — excision has a privacy requirement to actually purge, unlike
  reconcile), search providers (`sqlite_vec_queries.py`/`_runtime.py`,
  which also deduplicated a second, drifted hand-rolled DDL declaration
  onto the canonical `archive_tiers/embeddings.py` DDL), daemon similarity/
  metrics, CLI maintenance/status/rescue commands, demo seed/constructs
  (stub embedders only — `voyage-4` real API is never invoked in tests or
  demo).
- **04kl rescue retargeted into the new layout**: `rescue.py` now
  recomputes `embedding_input_hash` from each matched message's *current*
  embedder input text (the old `content_hash` match is what licenses
  treating "current text" and "retired text" as identical, but the new key
  is never trusted from the retired file) and writes rescued vectors
  directly into the v4 content-addressed shape — landing once, not
  migrating twice, per the bead's stated ordering.
- Docs: `docs/architecture.md`, `docs/schema.md` updated.

### Gap found and fixed during review

`storage/embeddings/sql.py` carried a `stale_messages_sql()` function and a
`STALE_MESSAGES_SQL_LEGACY` constant, both written during the rewrite but
never wired into any caller — `status_payload.py` (the primary runtime
status surface) computes its own inline stale comparison against the
`embedding_input_hash` column its relation already projects, and the
legacy monolithic-archive reader in `embedding_stats.py` deliberately
reports 0 (no configured model available to thread through) rather than
using either helper. Grepped `polylogue/` and `tests/` to confirm zero call
sites, then removed both as dead code (separate commit
`67b07a371`).

## Verification

- `ruff format --check polylogue tests` — 2092 files already formatted
- `ruff check polylogue tests` — all checks passed
- `mypy --strict polylogue` — Success: no issues found in 1059 source files
- `devtools test` on all 20 touched test files (property + unit, including
  the rebuild-survival proof `test_embedding_rebuild_survival.py`, the
  04kl-into-v4-layout proof `test_embedding_rescue_content_addressed_layout.py`,
  and the dedup proof `test_embedding_dedup.py`) — **329 passed**
- `devtools render all --check` — exit 0, zero "out of sync" hits (grepped,
  not just the tail line)
- `devtools lab policy schema-versioning` — `derived-tier upgrade helpers
  found: 0` (this PR's embeddings v3→v4 bump is a canonical-DDL edit, not
  an illegal upgrade helper — compliant). The report's one remaining
  violation (`undeclared index schema deltas found: 1`, missing `[40]`) is
  **pre-existing on `origin/master`** — reproduced identically on a bare
  `origin/master` checkout before this branch's changes — and unrelated to
  the embeddings tier.
- `devtools verify --quick` (format + lint + mypy + render-all-check +
  topology/layering/manifests/ci-workflows/doc-commands/docs-coverage/
  test-infra-currency/test-clock-hygiene/pytest-timeout-overrides/
  degrade-loudly) — `exit_code: 0`, all 16 steps `ok`

## Ordering note (polylogue-04kl)

Per the bead's design ("design this first, then execute the one-time 04kl
rescue directly INTO the content-addressed layout, avoid double
migration"): this PR lands the keying design + rescue retargeting +
synthetic-fixture tests. The live `--yes` rescue execution against the
production archive remains coordinator-owned and run post-promote, per
04kl's own notes — not part of this PR's scope.

Ref polylogue-q88p

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -317,7 +317,12 @@ full permission audit trail (PermissionRequest/PermissionDenied).
 Vector embeddings for semantic search, powered by Voyage AI (`voyage-4`,
 1024 dimensions) via SQLite-vec (`vec0` virtual table):
 
-- **Storage**: `message_embeddings` (vec0), `message_embeddings_meta`, `embedding_status`
+- **Storage**: `message_embeddings` (vec0) and `message_embeddings_meta` are
+  content-addressed, keyed by `embedding_input_hash = SHA-256(model,
+  embedder input text)` rather than message identity — see
+  [Schema § embeddings.db](schema.md#embeddingsdb-vectors-rebuildable-expensive)
+  for why. `message_embedding_refs` maps `message_id -> embedding_input_hash`;
+  `embedding_status` remains session-scoped.
 - **Search**: `--similar` flag triggers pure vector search; hybrid mode combines
   FTS5 + vector via Reciprocal Rank Fusion
 - **Integration**: Daemon-side post-ingest and ambient catch-up embedding is
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -23,7 +23,7 @@ over the `CREATE TABLE` statement.
 |-----------|------|------------------|
 | `source.py` | `source.db` | `SOURCE_SCHEMA_VERSION = 10` |
 | `index.py` | `index.db` | `INDEX_SCHEMA_VERSION = 35` |
-| `embeddings.py` | `embeddings.db` | `EMBEDDINGS_SCHEMA_VERSION = 3` |
+| `embeddings.py` | `embeddings.db` | `EMBEDDINGS_SCHEMA_VERSION = 4` |
 | `user.py` | `user.db` | `USER_SCHEMA_VERSION = 6` |
 | `ops.py` | `ops.db` | `OPS_SCHEMA_VERSION = 1` |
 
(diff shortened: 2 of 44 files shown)


Sinity:
CodeRabbit silent (rate-limit pattern); merging on coordinator review. Reviewed: identity-free embedding_input_hash matches the q88p design; refs mapping placed in embeddings tier (index version untouched — right call given the live v42 promote); both twins covered; 04kl rescue retargeted into v4; rebuild-survival + dedup + property tests present; inherited-WIP fully re-verified by the finishing lane (329 tests, mypy strict, quick gate). Vector GC deferral noted as follow-up debt.

PR #3197, opened 2026-07-20
feat(daemon): suppress trickle conveyor while bulk rebuild is in flight

Sinity:
## Summary

Closes the two residuals PR #3189 (polylogue-gd6v) explicitly deferred:
(1) the trickle raw-materialization conveyor and the daemon-owned
bulk-rebuild driver ran every tick regardless of each other, converging on
the same backlog with duplicate parse/replay work and needless writer-hold
contention; (2) the p99 writer-responsiveness claim for a live bulk drain
was never independently measured.

## Problem

PR #3189's own body: "No trickle suppression. Bulk routing does not pause
or coordinate with the trickle conveyor while a build is in flight...The
resulting redundant per-raw work during the backlog window is a known,
deliberately out-of-scope efficiency gap" and "agvo responsiveness p99
gate during a live drain -- not independently measured here; rests on the
same off-writer-hold parse mechanism phase (a) already established."

Both residuals are named directly on polylogue-gd6v (the bead's notes list
them as the only two items not shipped in #3189) and block the 3.14t
daemon deploy (polylogue-dcz5), which is the next step in the m6tp
convergence program.

## Solution

**Residual 1 (suppression):**

- `polylogue/daemon/cli.py`: new `_daemon_bulk_rebuild_transaction_in_flight()`
  read-only check, gated by the same `daemon_bulk_rebuild_routing` config
  flag `_maybe_route_daemon_bulk_rebuild` already uses (flag off => never
  even check, so a stale transaction record can't wrongly suppress the
  ONLY mechanism materializing raws).
- `_periodic_raw_materialization_convergence`'s burst loop consults this
  check first. When a bulk-rebuild transaction is in flight it skips the
  trickle census/drain writer pass for that tick (`_drain_raw_materialization_once`)
  and drives the bulk pass instead, bursting at the same 1s cadence the
  trickle conveyor already uses (rather than waiting a full 30s interval
  between bulk passes, which was the previous effective cadence once
  trickle's own backlog was exhausted).
- **Suppression boundary, justified from the convergence-stage inventory**
  (`docs/design/convergence-simplification-inventory.md` item 5, "Per-pass
  candidate requery / resume recompute"): the trickle drain and the bulk
  engine both walk the exact SAME `raw_sessions` -> index materialization
  work over the same backlog -- `repair_raw_materialization` and
  `rebuild_index_from_source_sync`'s census+replay phases are the same
  pipeline the bulk engine reuses (per #3189's own framing: "the SAME
  engine the offline CLI rebuild command drives"). The bulk engine's own
  paged cursor query (`IndexGenerationStore.next_raw_page`) is unfiltered
  by any fixed snapshot -- it queries live `raw_sessions` ordered by
  `(acquired_at_ms, raw_id)` each pass -- so it naturally absorbs raws that
  arrive mid-build; there is no work left for the trickle pass on those
  same raws while bulk is in flight. What suppression does NOT touch:
  live-ingest acquisition (a separate watcher pipeline that only appends
  to `source.db`, which the bulk engine's live query already picks up),
  hook-spool drain, embedding catch-up, and already-committed session
  insight convergence (`_periodic_session_insight_convergence_after`) --
  all separate periodic loops, untouched, verified by reading their call
  sites are outside `_periodic_raw_materialization_convergence`.
- `_maybe_route_daemon_bulk_rebuild` now returns `bool` (pass genuinely
  attempted vs. structural no-op / swallowed failure) so a real pass
  failure during suppression falls back to the slower 30s outer interval
  instead of retrying every burst second -- matching how a trickle pass
  failure already escapes to the outer interval via the existing exception
  handlers.

**Residual 2 (responsiveness evidence):**

- `tests/unit/daemon/test_daemon_bulk_rebuild_responsiveness.py` (new):
  drives the REAL `polylogue.daemon.bulk_rebuild.run_daemon_bulk_rebuild_pass`
  (the exact production pass driver, not a stub) against a 150-raw fixture
  archive, CONCURRENTLY with 3 small writer-actor coroutines (standing in
  for live-ingest appends / hook-spool drain writes) sharing the real
  `DaemonWriteCoordinator` every other daemon writer actor uses, and
  asserts the small actors' p99 queued-wait time stays under a documented
  budget throughout the drain.
- Budget derivation (not an arbitrary number): manual measurement at this
  exact fixture shape (150 raws, batch=8, 3 concurrent actors) found the
  correct bounded-pass implementation gives p99 wait ~0.32s (max single
  writer hold ~0.34s). Collapsing the SAME 150-raw corpus into one
  unbounded pass (reproducing a regression that drops per-pass batching)
  measured p99 wait ~1.26s. The shipped budget (1.0s) sits between those
  two real measurements -- several times the bounded-pass p99 for headroom
  against host CPU contention, while staying below what the unbounded-pass
  regression produces at this same fixture size. This makes it a genuine
  regression detector at this scale, not a loose ceiling nothing could
  ever hit (verified manually pre-submission; not shipped as a mutation
  test since it would require carrying the "bad" implementation variant in
  the test file).

## Verification

```
devtools test tests/unit/daemon/test_daemon_cli.py \
  tests/unit/daemon/test_bulk_rebuild.py \
  tests/unit/daemon/test_daemon_bulk_rebuild_responsiveness.py \
  tests/unit/daemon/test_write_coordinator.py \
  tests/unit/daemon/test_parse_prefetch.py
=> 134 passed in 11.78s (re-ran 3x for the new responsiveness test alone; stable)

uv run mypy --strict polylogue/daemon/cli.py tests/unit/daemon/test_daemon_cli.py \
  tests/unit/daemon/test_daemon_bulk_rebuild_responsiveness.py
=> Success: no issues found in 3 source files

devtools verify --quick
=> exit_code 0 (ruff format/check, mypy, render all --check, layering,
   topology, docs-coverage, closure-matrix all clean; grepped for
   "out of sync" -- none found)
```

New tests added to `tests/unit/daemon/test_daemon_cli.py` (6): the
in-flight check's flag gate (off never checks, on delegates to
`has_resumable_daemon_bulk_rebuild_transaction`), trickle suppression
while in flight (asserts the trickle drain's `daemon_write_coordinator().run_sync`
is never called), failure fallback to the outer interval, and automatic
resumption once the transaction is no longer in flight (trickle drain
runs again on the very next tick, no operator action). Also backfilled
`daemon_bulk_rebuild_routing = False` onto 4 pre-existing `FakeResolved`
config stubs (parse-stage-split tests, #3168) that predate this attribute
-- they broke with `AttributeError` once the new suppression check reads
`daemon_bulk_rebuild_routing` unconditionally at the top of the burst
loop; this is a mechanical fix restoring their original intent
(`daemon_bulk_rebuild_routing` off, unrelated to what they test), not a
behavior change to those tests.

## Non-obvious decisions

- Suppression is per-tick and re-evaluated every burst iteration, not a
  one-shot decision at loop entry -- this is what makes "resume
  automatically once the bulk transaction promotes or is abandoned" true
  within the SAME daemon process run, not only across restarts.
- Chose to keep `_maybe_route_daemon_bulk_rebuild`'s counts parameter
  (rather than adding a separate no-counts overload) and pass an empty
  `RawMaterializationCounts()` placeholder from the suppression branch --
  the function's own guard already treats "resumable transaction exists"
  as sufficient without needing real counts (see its `not
  _bulk_scale_raw_materialization_backlog(counts) and not
  has_resumable(...)` early-return), so the placeholder never needs to
  carry real data.

## What this does NOT change

- The bulk-rebuild engine itself (`polylogue/daemon/bulk_rebuild.py`,
  `polylogue/maintenance/rebuild_index.py`) -- unmodified.
- Archive-scale equivalence receipt and CLI-command deletion
  (polylogue-4jsk) -- still coordinator-owned follow-up, unaffected by
  this PR.
- The 3.14t daemon deploy itself (polylogue-dcz5) -- this PR removes two
  named blockers but does not perform the deploy.

Ref polylogue-gd6v

### DIFF
--- a/polylogue/daemon/cli.py
+++ b/polylogue/daemon/cli.py
@@ -668,7 +668,37 @@ def _maybe_recommend_bulk_rebuild(counts: RawMaterializationCounts) -> None:
     )
 
 
-async def _maybe_route_daemon_bulk_rebuild(counts: RawMaterializationCounts) -> None:
+async def _daemon_bulk_rebuild_transaction_in_flight() -> bool:
+    """Whether the daemon's own well-known bulk-rebuild transaction is running.
+
+    Read-only fast path (at most one JSON file read). Used by
+    ``_periodic_raw_materialization_convergence`` to stand its trickle
+    census/drain pass down for the current tick instead of duplicating work
+    the bulk-rebuild engine already subsumes: both walk the SAME
+    ``raw_sessions`` -> index materialization pipeline over the same backlog
+    (see docs/design/convergence-simplification-inventory.md item 5 -- the
+    bulk engine's own paged cursor query is unfiltered by any fixed
+    snapshot, so it naturally absorbs raws that arrive while it runs; there
+    is no work left for the trickle pass to do on the SAME raws in the
+    meantime).
+
+    Mirrors ``_maybe_route_daemon_bulk_rebuild``'s own flag gate exactly:
+    the flag off means never even check. Checking survives a daemon restart
+    via a durable transaction record, so a stale "in flight" read while
+    routing is disabled would wrongly suppress the trickle conveyor -- the
+    ONLY mechanism left materializing raws -- with nothing to replace it.
+    """
+    from polylogue.config import load_polylogue_config
+
+    if not load_polylogue_config().daemon_bulk_rebuild_routing:
+        return False
+    from polylogue.daemon.bulk_rebuild import has_resumable_daemon_bulk_rebuild_transaction
+    from polylogue.paths import archive_root
+
+    return await asyncio.to_thread(has_resumable_daemon_bulk_rebuild_transaction, archive_root())
+
+
+async def _maybe_route_daemon_bulk_rebuild(counts: RawMaterializationCounts) -> bool:
     """polylogue-gd6v: route a bulk-scale backlog into a daemon-owned blue-green rebuild.
 
     Off by default (``daemon_bulk_rebuild_routing`` config flag). Once a
@@ -679,11 +709,21 @@ async def _maybe_route_daemon_bulk_rebuild(counts: RawMaterializationCounts) -> 
     would waste every page already replayed into it. This runs after the
     trickle pass has already released the writer coordinator, so scheduling
     another writer-coordinated pass here is safe.
+
+    Returns whether a pass was genuinely attempted this call (``True``) or
+    the call was a structural no-op / hit a swallowed failure (``False``).
+    ``_periodic_raw_materialization_convergence``'s trickle-suppression
+    branch (residual of polylogue-gd6v) uses this in place of the trickle
+    pass's own ``made_progress`` signal to decide whether to keep bursting
+    bulk-rebuild passes back-to-back or fall back to the slower outer
+    interval -- a genuine pass failure must not turn into a tight 1s retry
+    storm, matching how a trickle pass failure already falls back to the
+    outer interval via the caller's exception handling.
     """
     from polylogue.config import load_polylogue_config
 
     if not load_polylogue_config().daemon_bulk_rebuild_routing:
-        return
+        return False
     from polylogue.config import Config
     from polylogue.daemon.bulk_rebuild import (
         DAEMON_BULK_REBUILD_OPERATION_ID,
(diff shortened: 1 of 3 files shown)


Sinity:
CodeRabbit hit its review rate limit on this PR, so no automated review ran. Coordinator review stand-in (same protocol as earlier campaign merges): I read the full diff — suppression boundary is exactly the trickle census/drain pass, gated by the same daemon_bulk_rebuild_routing flag the router uses; other periodic loops (live-ingest, hook-spool, embeddings, insights) are untouched; the responsiveness test drives the real pass driver + real DaemonWriteCoordinator with a budget bracketed by measured bounded-pass (~0.32s p99) vs unbounded-regression (~1.26s p99) numbers, so it is a genuine regression detector. Local verification: devtools test on 5 files (134 passed), mypy --strict clean, verify --quick exit 0.

PR #3204, opened 2026-07-20
fix(storage): authority-precedence head handoff between chain and membership replay

Sinity:
## Summary
Resolves the v42 rebuild-killing head collision between browser-capture (membership/quarantined) and provider-export (chain) evidence for the same logical conversation, in both arrival orders, via authority + frontier precedence instead of hard failure.

## Problem
The live v42 full rebuild crashed at ~cursor 10-12K: `raw revision CAS rejected a conflicting accepted head` for `chatgpt:1835026a-…` — a quarantined browser-capture raw (membership replay) and the byte-proven GDPR-export raw (chain replay) both claimed the revision head at equal semantic frontier 21 with different content hashes. The two subsystems only avoid each other within one batch (`byte_replayed_keys` skip); across pages, whichever writes first makes the other raise. 228 chatgpt capture/export overlaps exist in the live corpus (plus claude-ai analogues), and the reverse order raised `membership replay cannot retire an unrelated accepted head`.

## Solution
- `apply_raw_revision_replay`: an existing **semantic head owned by a quarantined raw** is superseded at equal-or-later incoming frontier (head handed to chain evidence); when the capture is strictly content-ahead, the chain cohort records DEFERRED receipts (`chain_replay:deferred_behind_content_ahead_quarantined_membership_head`) and leaves the head — receipted, never silently dropped.
- `apply_raw_membership_classification`: an existing head owned by **chain-governed (non-quarantined) evidence outside the cohort** no longer raises — the capture cohort yields (SUPERSEDED receipts naming the head raw, memberships terminally decided `superseded_equivalent`) unless strictly content-ahead on a comparable semantic frontier, in which case it takes the head. Byte-kind heads always win (byte frontier incomparable to semantic capture frontier).
- New `_raw_revision_authority` helper (source-tier lookup).
- Byte-proven equal-frontier cross-raw conflicts still reject exactly as before; the existing `test_full_replay_preserves_semantic_head_and_rolls_back_regressions` regression matrix is untouched and green.

## Verification
`devtools test tests/unit/storage/test_revision_replay.py tests/unit/sources/test_revision_backfill.py tests/unit/storage/test_raw_retention.py` → 118 passed (includes 4 new tests: chain-supersedes-capture at equal frontier, chain-defers behind content-ahead capture, membership-yields to chain head, membership-takes-head when content-ahead). `mypy` clean on the touched module. The live rebuild resume is the production proof: it re-walks the exact failing page next.

Known limitation (noted, not hidden): a capture that is content-ahead of a **byte-kind** chain head still yields — byte and semantic frontiers are incomparable; the capture bytes stay in the source tier and its receipt names the winning head.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved revision replay handling when quarantined membership records conflict with chain-governed revisions.
  * Revisions now correctly supersede, defer, or yield based on semantic frontier ordering.
  * Preserved existing authoritative heads when incoming content is behind.
  * Improved receipt and session tracking when revision heads change.

* **Tests**
  * Added coverage for equal, ahead, and behind revision-frontier scenarios.
  * Verified head switching, deferred applications, superseded receipts, and stored session content.

### DIFF
--- a/polylogue/storage/sqlite/archive_tiers/archive.py
+++ b/polylogue/storage/sqlite/archive_tiers/archive.py
@@ -2164,6 +2164,14 @@ class ArchiveStore:
             if not changed:
                 return
 
+    def _raw_revision_authority(self, raw_id: str) -> str | None:
+        row = (
+            self._ensure_source_conn()
+            .execute("SELECT revision_authority FROM raw_sessions WHERE raw_id = ?", (raw_id,))
+            .fetchone()
+        )
+        return None if row is None or row[0] is None else str(row[0])
+
     def raw_revision_replay_plan(self, logical_source_key: str) -> RevisionReplayPlan:
         return plan_revision_replay(self._raw_revision_candidates(logical_source_key))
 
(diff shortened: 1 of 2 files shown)


Sinity:
Both findings addressed in 7fa15441d: the count-based content-ahead exception is removed entirely in both directions — a scalar frontier can't prove content dominance, so chain-governed evidence now wins unconditionally over quarantined membership heads (captures stay receipted in the source tier; genuine content-ahead re-adoption deferred to a prefix-dominance follow-up bead). Tests updated to pin the unconditional-precedence semantics (20 passed).

Issue #13, opened 2026-04-07
Conversation imports flatten branch structure and lack shared metadata persistence


Sinity:
## Context

Each of the four importers (ChatGPT, Claude AI, Claude Code, Codex) independently calls `build_markdown_from_chunks()` and `persist_document()`. There is no shared processing step between parsing provider-specific JSON and writing the final Markdown. This means:

- **Branch structure is discarded.** ChatGPT exports include a full message DAG via `mapping` with parent pointers — edits and regenerations create forks. Claude AI exports can carry `parent_id` / `parentUuid` on messages. Currently everything is linearised into one transcript. Forks vanish.

- **No queryable metadata store.** The only record of a conversation's structure is the Markdown front matter. To answer "how many branches does this conversation have?" or "what's the total word count across all Claude imports?" you'd need to reparse every `.md` file.

- **Duplicated orchestration logic.** Each importer assembles chunks, calls the render function, calls persist, handles skip/dirty logic, and constructs an `ImportResult`. The four implementations are nearly identical in structure but diverge in small ways (some check `persist_result.skipped` differently, some set different extra_state keys).

## Additional gaps

- **Token counts lack word-count companions.** Token counts are model-dependent approximations. There's no deterministic word count to sanity-check against.

- **No per-turn timestamps in output.** The front matter records `sourceModifiedTime` but individual turns don't show when they occurred. For long conversations spanning hours or days, temporal context is lost.

- **Escaped footnotes render poorly.** Multiple providers double-escape bracket references (`\[1\]` instead of `[1]`). These appear as literal backslash-bracket sequences in the rendered Markdown.

- **ChatGPT citation tokens are opaque.** ChatGPT embeds citation markers as private-use Unicode characters (U+E200–U+E202) that render as empty rectangles or invisible characters.

- **No `--force` flag on import commands.** If a user hand-edits a Markdown file and then re-imports, the importer skips the conversation silently. There's no way to say "overwrite my edits."

- **Pydantic overhead in the hot path.** The Codex importer uses Pydantic models for JSONL line parsing, which rejects valid entries with unexpected fields and adds import latency.

## Scope

- All four importers
- Render pipeline (word counts, timestamps, citation resolution)
- CLI import subcommands (--force, summary formatting)
- New SQLite schema for conversation/branch/message metadata
- New branch planning module for DAG traversal
- Provider documentation updates

Out of scope: query CLI for the SQLite store (that's downstream), HTML rendering of branch views, migration tooling for existing archives.

PR #3205, opened 2026-07-20
fix(sources): absorb quarantined membership head into later rebuild cohorts

Sinity:
## Summary
Third and (per the loud-context guard) currently-last v42 rebuild crash class: a quarantined membership head decided on an earlier rebuild page is absent from a later page's cohort for the same logical key, so membership replay raised `cannot retire an unrelated accepted head`. The head raw is now absorbed into the cohort candidate set so the real prefix classifier ranks it — no scalar comparisons, no authority inversion (chain-governed heads are deliberately not absorbed; the #3204 yield path owns those).

## Problem
Post-#3204 the live walk died at `chatgpt:366f7d9f-…` (full context named by the new loud guard): existing head raw `ff4f5e…` (quarantined) vs a 3-raw sibling cohort that did not contain it, because `raw_membership_rebuild_raw_ids` joins on `raw_sessions.revision_authority = 'byte_proven'` and page-local candidates only cover the current page.

## Solution
One addition in `backfill_historical_revision_evidence`'s membership loop: look up the current head raw for the key; if quarantined, add it to `candidate_raw_ids` so it is spilled, parsed, and classified with the cohort. In-cohort head retirement then follows the existing, already-tested path.

## Verification
Safe-by-construction (absorption only widens classification evidence) + `devtools verify --quick` green. Honest gap: two attempts at a self-contained backfill fixture failed to reproduce the exact census-authority state — both fixture variants passed with AND without the fix (the census re-derives raw authority during the walk), so a synthetic regression test would be vacuous and was not kept. The verification artifact is the live v42 walk, which deterministically re-hits this key on resume — the pass/receipt will be posted on this PR before merge.

### DIFF
--- a/polylogue/sources/revision_backfill.py
+++ b/polylogue/sources/revision_backfill.py
@@ -784,6 +784,18 @@ def backfill_historical_revision_evidence(
             retained_bytes = 0
             candidate_raw_ids = set(archive.raw_membership_rebuild_raw_ids(logical_key))
             candidate_raw_ids.update(membership_candidates.get(logical_key, ()))
+            # Cohort absorption: candidate selection is page-dependent, so a
+            # head written by an EARLIER page's membership cohort for this key
+            # may not be in this page's candidate set -- membership replay
+            # would then refuse to retire an "unrelated" quarantined head and
+            # kill the walk. Absorb the current quarantined head raw into the
+            # cohort so the real prefix classifier ranks it against the new
+            # members instead of any scalar comparison. Chain-governed
+            # (non-quarantined) heads are deliberately NOT absorbed --
+            # apply_raw_membership_classification yields to those.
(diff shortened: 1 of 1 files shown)


Sinity:
Live-walk receipt as promised: resume #13 (this branch's working tree) completed the previously-crashing page — 'pass 1: engine status=paused', cursor advanced 12000 → 14000 in 1131s, no retire/CAS raise. The chatgpt:366f7d9f-… cohort (head ff4f5e… + 3 siblings) was absorbed and classified without incident. That is all three v42 crash classes cleared on the production corpus.

Sinity:
CodeRabbit rate-limited (green check is the limit notice). Coordinator review stand-in: single additive block in the backfill membership loop (head lookup + quarantined-only absorption into candidate set); no behavior change for chain-governed heads; verify --quick green; live-walk receipt above is the production proof.

PR #3206, opened 2026-07-20
feat(sources): per-page stage-timing telemetry for historical backfill

Sinity:
## Summary
One aggregated stage-timing INFO line per `backfill_historical_revision_evidence` call (census / spill_load / index_parsed_write / …, largest first), using the per-method `stage_timings_s` plumbing that already existed but was never threaded from this caller.

## Problem
The live v42 rebuild's engine pass runs 10-17 min per 2,000-raw page with no internal attribution; py-spy cannot read the nix-built interpreter, so the serial-writer cost question ("where do 0.3s/raw go?") was unanswerable.

## Solution
`polylogue/sources/revision_backfill.py` only: timing dict + census/spill wraps + `stage_timings_s=` threading into both apply paths + one INFO log before return. No behavior change.

## Verification
`devtools test tests/unit/sources/test_revision_backfill.py` → 39 passed; mypy/ruff clean. Live receipt: the resumed v42 walk logs the line each page — will be quoted on this PR.

## Summary by CodeRabbit

* **Monitoring**
  * Added timing metrics for historical revision backfill stages.
  * Backfill operations now log the slowest processing stages to help identify performance bottlenecks.

### DIFF
--- a/polylogue/sources/revision_backfill.py
+++ b/polylogue/sources/revision_backfill.py
@@ -3,11 +3,13 @@
 from __future__ import annotations
 
 import json
+import logging
 import os
 import pickle
 import sqlite3
 import tempfile
 import threading
+import time
 from collections.abc import Callable, Iterator, Sequence
 from contextlib import contextmanager
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


Sinity:
Live receipt from resume #14, first instrumented page (cursor 14000→16000): 'backfill stage timings: census=135.0s spill_load=105.2s membership_replay.index_parsed_write=50.8s membership_replay.index.full_replace=31.8s …'. Headline: ~55% of the serial engine pass is census + spill reload (two-phase overhead), not index writes — that's the optimization target the telemetry was built to find.

Sinity:
Superseded by #3208, which carries this commit plus fixes for all three review findings (polylogue logger routing, census_receipt + total timing, missed spill wrap).

PR #3207, opened 2026-07-20
perf(sources): no-fsync census spill on data disk + decoded-session RAM layer

Sinity:
## Summary
Cuts the two dominant non-write stages of the serial backfill engine pass (measured live via #3206 telemetry: `census=135.0s spill_load=105.2s` vs ~50-100s of index writes per 2,000-raw page).

## Problem
The census spill is a disposable parse cache, but it paid durable-storage costs: a synchronous journal commit per censused raw, placed under `archive_root` — which on symlinked deployments is the wear-limited root SSD, not the NVMe the tier symlinks point at. Replay then re-read what census had just parsed through sqlite + `pickle.loads`.

## Solution
`_ParsedSessionSpill` only: `journal_mode=OFF`/`synchronous=OFF`/`temp_store=MEMORY` (loss = reparse durable source bytes, by design); spill file placed beside the resolved `index.db`; a 256MiB payload-bounded decoded-session dict layer in front of the pickle path (cohorts are consumed almost immediately after census).

## Verification
`devtools test tests/unit/sources/test_revision_backfill.py tests/unit/storage/test_revision_replay.py` → 59 passed; mypy/ruff clean. Live receipt: next instrumented page's stage-timing line will be quoted here for before/after.

### DIFF
--- a/polylogue/sources/revision_backfill.py
+++ b/polylogue/sources/revision_backfill.py
@@ -14,7 +14,7 @@ from dataclasses import dataclass
 from io import BytesIO
 from pathlib import Path
 from types import TracebackType
-from typing import BinaryIO, Literal
+from typing import BinaryIO, Final, Literal
 
 from polylogue.archive.ingest_flags import (
     COMPACT_BROWSER_CAPTURE_INGEST_FLAG,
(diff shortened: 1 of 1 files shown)


Sinity:
Superseded by #3208 (same commit included in its stack).

PR #3208, opened 2026-07-20
perf(sources): rebuild fast path — stage telemetry, spill cache, commit batching

Sinity:
## Summary
One coherent perf phase for the full-archive rebuild (operator mandate: 100K-raw import must be dramatically faster). Three commits: (1) per-page stage-timing telemetry (supersedes #3206, includes all three CodeRabbit fixes from its review); (2) census-spill fast path — no-fsync pragmas for the disposable cache, spill placed beside the resolved index tier (NVMe, not the wear-limited root SSD), 256MiB decoded-session RAM layer (supersedes #3207); (3) rebuild path opts into the existing amg1/oikv commit batching (`commit_batch_size=200`) instead of one fsync'd commit per censused raw / replayed cohort.

## Problem
Live v42 telemetry measured `census=135s spill_load=105s` vs ~50-100s of index writes per 2,000-raw page — over half the serial engine pass was cache overhead and per-unit commit cycles.

## Solution
`polylogue/sources/revision_backfill.py` (telemetry + spill) and `polylogue/maintenance/replay.py` (batching constant + call-site). Crash behavior for batching is the already-pinned batch-discard/resume contract. Telemetry routes through `polylogue.logging.get_logger`; census receipt work and a `total=` figure are included per review.

## Verification
`devtools test tests/unit/sources/test_revision_backfill.py` → 39 passed; batching/resume subset 8 passed; maintenance suite 194 passed; mypy/ruff clean. Live before/after page timings from the resumed v42 walk will be posted here.

### DIFF
--- a/polylogue/maintenance/replay.py
+++ b/polylogue/maintenance/replay.py
@@ -99,6 +99,14 @@ _STATE_DIRNAME: Final[str] = ".maintenance-state"
 #: never becomes a second archive-wide materialization.
 _REBUILD_CENSUS_SPILL_CACHE_BYTES: Final[int] = 512 * 1024 * 1024
 
+#: Rebuild-scale commit batching (polylogue-amg1/oikv machinery; the rebuild
+#: caller previously never opted in, paying one fsync'd commit per censused
+#: raw and per replayed cohort -- thousands per page). A crash discards at
+#: most one open batch and the resume reprocesses it from scratch (contract
+#: pinned by test_backfill_resumes_after_replay_batch_crash_discards_whole_
+#: batch_cleanly et al.), so the loss window is bounded and cheap.
+_REBUILD_COMMIT_BATCH_UNITS: Final[int] = 200
+
 
 # ---------------------------------------------------------------------------
 # Target dispatch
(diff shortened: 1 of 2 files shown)


Sinity:
Both findings triaged: (1) the incomplete-cohort 'with conn' commit is the oikv-documented deliberate behavior ('always commits immediately regardless of batching, matching prior behavior' — archive.py comment at that site), and for THIS caller it is additionally moot since the lock fix runs the replay phase unbatched (replay_commit_batch_size=1); the residual hazard for future batched-replay callers is inherent to the pre-existing oikv design, not introduced here. (2) LIFO→FIFO eviction fixed in the follow-up commit.

Sinity:
Live telemetry receipt (whale-dense 500-raw page, full stack): 'total=675.3s revision_replay.index_parsed_write=165.3s census=165.3s full_replace=144.6s spill_load=112.9s clear_projection_rows=92.7s …'. The total= accounting works; it also exposed clear_projection_rows scaling pathologically on whale sessions (92.7s/500 whale raws vs 11.1s/2000 normal raws) — now the top target of the follow-up write-path lane under polylogue-623q.

PR #3209, opened 2026-07-20
perf(daemon): bound parse-stage cache by estimated tree bytes

Sinity:
## Summary

Bounds `DaemonParseStage`'s prefetch cache by an *estimated parsed-tree*
byte budget, in addition to the existing raw-payload byte budget, so a
whale-dense census page can no longer pin gigabytes of resident memory
in the cache regardless of the payload-bytes clamp.

## Problem

`DaemonParseStage`'s inflight admission budget and
`RawParsePrefetchCache`'s own admission gate both account raw PAYLOAD
bytes, because payload size is the only thing knowable before a raw is
parsed. But a parsed `ParsedSession` tree resident in the cache is not
the same size as the payload it came from -- Pydantic model instances,
per-block dicts, and general Python object overhead inflate a compact
JSON/JSONL payload substantially. Two earlyoom kills (19.3G and 20.2G
RSS peaks, 2026-07-20) happened on a whale-dense page precisely because
the cache retained a whole 2000-raw page of parsed trees while only raw
payload bytes were budgeted; clamping the inflight (pre-parse) budget
did nothing, since the pressure came from trees already sitting in the
cache post-parse, not from parses in flight.

## Solution

`polylogue/daemon/parse_prefetch.py`:

- `estimate_parsed_tree_bytes()` -- a cheap, single linear-pass
  structural estimator over `list[ParsedSession]`: sums text/content
  field lengths across sessions/messages/blocks/attachments/session
  events and counts model-instance nodes, then applies two constants
  (`_ESTIMATOR_BYTES_PER_CHAR=2`, `_ESTIMATOR_OBJECT_OVERHEAD_BYTES=1024`)
  calibrated against a manual deep-object-graph measurement on
  synthetic sessions (calibration data + fit in the comment above the
  constants; test reproduces the measurement independently). Deliberately
  NOT a recursive `sys.getsizeof`/pympler-style deep walk -- that runs on
  `warm()`'s hot path for every raw in a page (up to ~2000), so it has
  to stay O(payload size) with low constant factor, not O(object graph).
- `daemon_parse_stage_max_cached_tree_bytes()` -- a second adaptive
  budget (1/8 of physical RAM, clamped `[256 MiB, 4 GiB]`, override
  `POLYLOGUE_DAEMON_PARSE_STAGE_MAX_CACHED_TREE_BYTES`), mirroring the
  existing inflight-bytes budget's shape but sized for the bigger
  (resident-tree) number.
- `DaemonParseStage` tracks a side ledger (`_tree_bytes_by_raw_id`,
  `_cached_tree_bytes_total`) keyed by the same raw_ids as
  `self.cache`. `RawParsePrefetchCache` itself is not touched or
  subclassed -- it's a shared type `bulk_rebuild.py` hands directly to
  `RebuildIndexRequest.prefetch_cache`, so this stays additive rather
  than replacing it.
- A tree whose estimate exceeds the WHOLE budget is never admitted at
  all (skips `cache.try_admit` entirely, so it doesn't even consume a
  payload-bytes admission slot) -- the writer-held pass reparses it
  normally, identical to any other prefetch miss.
- Admitted trees are evicted largest-first (ties break to the oldest
  insertion, via dict iteration order) through the raw cache's own
  `pop()` once the running estimated total exceeds budget. Since
  external consumers (the writer-held pass) can also `pop()` entries
  directly, the ledger reconciles lazily against `self.cache.contains()`
  before every eviction decision.

## Verification

- `devtools test tests/unit/daemon/test_parse_prefetch.py` -> 12 passed
  (7 pre-existing + 5 new: estimator monotonicity, estimator within 3x
  of an independent deep-size measurement, whale-never-retained,
  eviction-under-pressure, and the new budget's adaptive-clamp/env-override).
- `.venv/bin/python -m mypy --strict polylogue/daemon/parse_prefetch.py
  tests/unit/daemon/test_parse_prefetch.py` -> `Success: no issues found in 1 source file` (both files)
- `devtools verify --quick` -> all 16 steps `ok`, exit 0 (includes
  `render all --check` / topology / docs-coverage, since new
  module-level functions were added to an existing file -- no new
  module, so no topology-projection regen was needed)

Ref polylogue-xb4i

### DIFF
--- a/polylogue/daemon/parse_prefetch.py
+++ b/polylogue/daemon/parse_prefetch.py
@@ -33,13 +33,15 @@ turns the same code path into a real speedup.
 from __future__ import annotations
 
 import os
-from collections.abc import Sequence
+import threading
+from collections.abc import Mapping, Sequence
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from polylogue.config import Config
 from polylogue.logging import get_logger
 from polylogue.sources import revision_backfill
 from polylogue.sources.dispatch import is_stream_record_provider
+from polylogue.sources.parsers.base_models import ParsedSession
 from polylogue.sources.revision_backfill import RawParsePrefetchCache
 from polylogue.storage.repair import (
     raw_materialization_pending_census_raw_ids,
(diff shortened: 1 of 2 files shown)


Sinity:
CodeRabbit rate-limited (green check is the limit notice, zero inline findings posted). Coordinator review stand-in: read the full diff — calibrated two-term estimator (measured fit documented, both constants rounded up so misestimation biases toward eviction/reparse, never OOM), separate side-ledger keeps RawParsePrefetchCache's shared type untouched for bulk_rebuild consumers, whale-never-retained + largest-first eviction with lazy reconciliation, env override wins. 12 tests green, mypy --strict clean, verify --quick 16/16.

PR #3211, opened 2026-07-20
perf(sources): tree-byte decoded layer + in-cohort head-retire drift fix

Sinity:
## Summary
The census spill's decoded-session RAM layer now budgets by estimated tree bytes (adaptive RAM/16, clamped [256MiB, 2GiB]) instead of a fixed 256MiB of payload bytes, and `estimate_parsed_tree_bytes` moves to `polylogue/pipeline/parsed_tree_size.py` so both the daemon prefetch cache and `sources/` can share it without a layering violation.

## Problem
Live whale-page telemetry (post-#3208/#3210): `spill_load` stuck at ~112-124s per page — whale trees exceed the payload-denominated decoded budget and fall through to sqlite+pickle decode every cohort.

## Solution
Estimator relocation (verbatim, calibration comment intact; daemon imports from the new home), tree-byte accounting in `_ParsedSessionSpill` with FIFO eviction and whales-never-retained, topology projection regenerated for the new module.

## Verification
`devtools test tests/unit/sources/test_revision_backfill.py tests/unit/daemon/test_parse_prefetch.py` → 51 passed; mypy clean; `render topology-projection`/`topology-status` committed. Live page timings from the resumed walk will be posted as the receipt.

## Summary by CodeRabbit

* **Performance**
  * Improved memory estimation for parsed session data.
  * Updated decoded-session caching to budget and evict based on estimated in-memory usage.

* **Bug Fixes**
  * Improved archive membership replay handling for equivalent raw revisions, while keeping protections against unrelated revisions.

* **Documentation**
  * Refreshed topology projection outputs and “topology status” summary metrics to match the latest structure.


Sinity:
Both findings fixed in the follow-up commit: inline_bytes counted byte-for-byte in the estimator; new shared effective_physical_memory_bytes() (cgroup v2/v1 aware) feeds both the daemon prefetch budgets (#3209 site included) and the spill decoded layer.

PR #3218, opened 2026-07-20
feat(devtools): monotonic, receipted Beads JSONL sync engine

Sinity:
## Summary

Adds `devtools/beads_sync.py`: a monotonic, per-row, receipted merge engine for `.beads/issues.jsonl`-shaped files, wired as `devtools workspace beads-sync merge`.

## Problem

The 2026-07-15 planning-surface incident showed that whole-file JSON validity and command exit status are not proof of synchronization. `.beads/issues.jsonl` was staged with literal `<<<<<<<`/`=======`/`>>>>>>>` conflict markers while `git status` reported a plain modification (not an unmerged index entry); a subsequent plain re-export produced a syntactically valid, unique-id JSONL that nonetheless silently reverted nine independently-edited Beads to older versions and reintroduced every previously repaired horizon label, while both standing backlog lints still passed. Synchronization must compare per-row revision, not trust file-level validity.

## Solution

- `merge_issue_sets()` compares every id's `updated_at` (the only revision proxy present in `bd export` output — there is no separate numeric revision field) and classifies each row: `created` / `retained` / `equal` / `updated` / `skipped_downgrade` / `conflicted` / `downgraded`. Ordinary sync never adopts a row that is not strictly newer than what it would replace; same-timestamp-different-content rows are reported as `conflicted`, never silently resolved either way.
- `RecoveryRequiredError` enforces that any downgrade requires an explicit `actor` + `reason`; every downgraded row is recorded in the resulting `SyncReceipt` with both revisions, actor, reason, and the merge fingerprint (project/database/branch/source).
- `load_jsonl_rows()` / `atomic_write_jsonl()` refuse to read or overwrite any file containing literal conflict markers, refuse duplicate ids, and write via temp+fsync+`os.replace` with a post-write parse/count validation before the target file is ever touched.
- `bootstrap_union()` folds N sources (e.g. two concurrent empty bootstraps plus a writer) into one order-independent union, raising `PlanningSurfaceCorruptError` on a true same-timestamp conflict (bootstrap has no operator present to authorize a recovery choice).
- Wired as a real `CommandSpec` (`devtools workspace beads-sync merge --base ... --incoming ... --output ... [--recover --actor ... --reason ...] [--json]`); `docs/devtools.md` regenerated.

**Scope note:** per this bead's own notes ("this bead owns... import-export, atomic conflict-aware file replacement, and synchronization receipts... do not implement a second merge algorithm here"), this PR ships the merge/receipt engine only. Consuming the receipt to gate repo policy/frontier invariants after union is the explicitly separate consumer bead `polylogue-8jg9.1`. Wiring this engine into `bd`'s own checkout/merge-hook window (currently `devtools/bd_reimport_guard.py`, a simpler live-DB before/after snapshot comparison) and full server/embedded-mode + killpoint regression coverage (AC #7) are left as follow-up integration work rather than folded into this PR — see AC matrix below.

## AC matrix (polylogue-gxjh.1)

1. **Satisfied.** Stale/newer/equal/incomparable rows produce machine-readable outcomes with both revisions (`test_merge_outcomes_cover_created_updated_equal_skipped_conflicted`); ordinary sync cannot downgrade (`test_ordinary_sync_cannot_downgrade_even_when_requested_without_recovery`).
2. **Satisfied.** Recovery override requires actor+reason and records fingerprint/actor/reason/every downgraded row (`test_recovery_without_actor_or_reason_is_refused`, `test_recovery_override_applies_downgrade_and_records_full_provenance`).
3. **Satisfied (engine-level).** `bootstrap_union()` is order-independent and loses nothing across simulated concurrent-bootstrap-plus-writer orderings (`test_bootstrap_union_is_order_independent_and_loses_nothing`, `test_bootstrap_union_from_empty_base_keeps_every_row`). Not tested against the live Dolt server's actual concurrency primitives — this module doesn't touch the live DB by design (see Solution); a true concurrent-process regression against `bd`'s server lock belongs with the hook-integration follow-up.
4. **Satisfied.** `atomic_write_jsonl` is temp+fsync+atomic-rename, validates parse/unique-ids before replacement, and refuses to overwrite a marker-bearing target (`test_atomic_write_jsonl_*`, `test_load_jsonl_rows_refuses_conflict_markers_even_with_valid_json_lines`). "Refuses ... changed-since-snapshot targets" is deferred — this module has no live-DB snapshot identity to compare against (no revision counter beyond `updated_at`); flagged as a residual gap below.
5. **Satisfied.** `test_replay_2026_07_15_staged_conflict_recovery_preserves_repaired_beads` directly replays the incident shape (nine repaired horizon beads, stale whole-file candidate) and proves the stale replacement is refused while unrelated new rows still merge in; `test_replay_recovery_path_can_still_restore_a_deliberately_older_row` covers the authorized-recovery counterpart.
6. **Deferred to polylogue-8jg9.1** (explicitly the separate consumer bead per this bead's own notes) — receipt shape is designed for that consumption (`SyncReceipt.to_json()`, `covers()` non-progress check) but this PR does not wire a post-union backlog/frontier re-check.
7. **Deferred.** Server-vs-embedded-mode coverage and killpoints around import/export are out of scope for this PR; this is a pure-Python file-level engine with no live-process kill-point surface of its own. Follow-up: wire into `devtools/bd_reimport_guard.py`'s hook window and add a killpoint regression there.

## Verification

- `devtools test tests/unit/devtools/test_beads_sync.py` — 20 passed.
- `/realm/project/polylogue/.venv/bin/python -m mypy --strict devtools/beads_sync.py devtools/command_catalog.py tests/unit/devtools/test_beads_sync.py` — no issues found in 3 source files.
- `devtools verify --quick` — 16/16 steps passed (also re-ran automatically via the pre-push hook).
- Manual CLI smoke: `devtools workspace beads-sync merge --base ... --incoming ... --output ... --json` against temp files, confirmed atomic output + correct receipt.
- Not run: full `devtools verify --all` / broader integration suite (not required for this file-scoped change; testmon-affected set is the two new/changed files above).

Ref polylogue-gxjh.1

### DIFF
--- a/devtools/beads_sync.py
+++ b/devtools/beads_sync.py
@@ -1,0 +1,461 @@
+"""Monotonic, receipted synchronization for `.beads/issues.jsonl` (polylogue-gxjh.1).
+
+Background
+----------
+
+The 2026-07-15 planning-surface incident (see polylogue-gxjh, polylogue-gxjh.1
+notes) had two distinct failure modes, both invisible to "does the file parse
+and does the lint pass":
+
+1. A staged `.beads/issues.jsonl` contained literal `<<<<<<<`/`=======`/
+   `>>>>>>>` conflict markers while `git status` reported a plain
+   modification, not an unmerged index entry. Whole-file JSON validation and
+   the backlog-hygiene lint both silently pass on a file that happens to have
+   valid-looking lines around the markers.
+2. The "fix" -- a plain re-export/overwrite -- replaced the file with a
+   syntactically valid 883-row snapshot that nonetheless *reverted* nine
+   independently-edited Beads to older versions, because whole-file
+   replacement carries no per-row comparison: a stale snapshot with valid
+   JSON always "wins" against newer live state if nothing merges by
+   revision.
+
+This module is the durable fix: it never trusts "the file parses" or "the
+command exited 0" as proof of synchronization. Every merge is per-row,
+compares `updated_at` (the only revision proxy present in `bd export`
+output; there is no separate numeric revision field), and produces a
+machine-readable `SyncReceipt` that enumerates every id's outcome plus both
+revisions where relevant. Downgrades (incoming row strictly older than the
+row it would replace) are refused by default; the only way to apply one is
+an explicit recovery override that must carry an actor and a reason, and
+every downgraded row still appears in the receipt.
+
+This module deliberately does not talk to the live `bd`/Dolt process for its
+core merge -- `bd export`/`bd import` are the actual live-state boundary
+(see `devtools/bd_reimport_guard.py`, which already wraps those for the
+checkout/merge-hook window). `beads_sync` operates on JSONL files: the
+committed `.beads/issues.jsonl`, a candidate recovery file, or a fresh `bd
+export` snapshot the caller took immediately before calling in. Consumers
+(the `bd_reimport_guard` hook wrapper, an operator doing conflict recovery,
+or the repo policy layer at polylogue-8jg9.1) call `synchronize_files` /
+`merge_issue_sets` and act on the returned `SyncReceipt` rather than
+re-implementing comparison logic.
+
+Outcome vocabulary (AC #1)
+---------------------------
+
+For every id in the union of base ∪ incoming:
+
+  created            -- id only in incoming; adopted.
+  retained           -- id only in base; kept (incoming does not carry it).
+  equal              -- same `updated_at` and same content; no-op.
+  updated            -- incoming has a strictly newer `updated_at`; adopted.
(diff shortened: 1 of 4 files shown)


Sinity:
Superseded: polylogue-gxjh.1 was delivered and merged via PR #3220 (squash 490088530) from the direct replacement lane, launched after this workflow lane's original spawn failed on the worktree-add race. The #3220 implementation carries the same monotonic merge engine + receipts and additionally the reconcile/export commands; bead closed on those receipts. Closing this duplicate — branch left intact for reference until workflow cleanup.

PR #3219, opened 2026-07-20
perf(storage): extract pure prepare_session_rows for off-writer prep

Sinity:
## Summary

Extracts the CPU-bound row-tuple-building work in the full-replace write path (`write_parsed_session_to_archive` / `_replace_full_session_messages_and_blocks`) into a pure function that can run off the single-writer thread, with a cheap, always-safe validity check gating reuse.

## Problem

`write_parsed_session_to_archive`'s full-replace path converts a `ParsedSession` tree into `messages`/`blocks` SQL row tuples (enum lookups, JSON encoding of `tool_input`, per-message/per-block content hashing) entirely inside the single-writer transaction hold. Measured on the live archive: `index.full_replace.messages`/`.blocks` stage timings of 34-40s per 2,000-page rebuild pass and 165-287s on whale pages, while the daemon's 14-thread parse-prefetch pool (`polylogue/daemon/parse_prefetch.py`) sits idle ahead of the writer hold. See `bd show polylogue-623q` for the full import-performance-envelope context.

## Solution

- `polylogue/storage/sqlite/archive_tiers/write.py`: split `_write_messages`/`_write_blocks`'s inline row-building generators into pure functions `_build_message_rows`/`_build_block_rows` (no DB access — only `messages`/`session_id`/`duplicate_native_ids` in, row tuples out). `_write_messages`/`_write_blocks` gained an optional `rows=` passthrough so a caller with already-built rows skips rebuilding.
- New `PreparedSessionRows` (frozen dataclass: `session_id`, `session_content_hash`, `message_rows`, `block_rows`) and `prepare_session_rows(session: ParsedSession) -> PreparedSessionRows` — a pure function computable from nothing but a `ParsedSession`, safe to call from any thread (e.g. a parse-prefetch worker) well before any writer hold exists. `session_content_hash` is `pipeline.ids.session_content_hash(session)` — the same hash ordinary ingest callers already compute and pass as `content_hash=`.
- `write_parsed_session_to_archive` gains an optional `prepared=` parameter. Right before the full-replace write, it decides whether to reuse the supplied `PreparedSessionRows` under two independent, cheap conditions checked on the writer thread:
  1. `prepared.session_content_hash == <this call's own content_hash>` — a session whose content changed between prepare-time and write-time (or a caller that never passed `content_hash`) always falls back to inline preparation.
  2. Lineage resolution did not slice `messages` — `_extract_prefix_tail` returns `messages` unchanged in every case except real prefix-sharing composition against an already-archived parent (detected via `lineage_inheritance == "prefix-sharing"`), which the prefetch worker has no DB access to know about ahead of time.
- Any other combination transparently rebuilds rows inline; `prepared=None` (the default, every existing caller) reproduces the exact prior behavior byte-for-byte.

**Not wired in this PR:** the daemon's parse-prefetch cache (`RawParsePrefetchCache`) and the census/spill/replay pipeline in `sources/revision_backfill.py`. That pipeline's actual index writes go through `ArchiveStore.apply_raw_revision_replay`/`apply_raw_membership_classification`, which resolve mixed full-replace/append cohorts out of a pickled sqlite spill cache (`_ParsedSessionSpill`) shared between the census and replay phases across three independent budget/eviction mechanisms (inflight-bytes, tree-bytes, spill payload-bytes) plus cohort-resolution logic that decides per-raw full-vs-append semantics. Wiring `PreparedSessionRows` all the way through that machinery is a separate, larger change that can now build on this tested extraction + correctness contract rather than reinventing it. Remaining scope tracked under `polylogue-623q`.

## Verification

```
devtools test tests/unit/storage/test_prepared_session_rows.py \
  tests/unit/storage/test_revision_replay.py \
  tests/unit/storage/test_archive_tiers_write.py \
  tests/unit/storage/test_lineage_normalization.py \
  tests/unit/storage/test_bulk_build_lifecycle.py
# 126 passed in 7.71s

devtools verify --quick
# exit_code: 0 (ruff format/check, mypy --strict, render all --check,
# topology/layering/manifests/doc-commands gates)
```

New tests (`tests/unit/storage/test_prepared_session_rows.py`):
- equivalence: writing a synthetic corpus (plain text, tool_use/tool_result/thinking blocks, duplicate-native-id messages) via inline building vs. via `prepare_session_rows` produces byte-identical `sessions`/`messages`/`blocks` dumps.
- anti-vacuity reuse proof: monkeypatches `_build_message_rows`/`_build_block_rows` to raise, then proves a matching-content-hash write with `prepared=` still succeeds with correct rows — only possible if the inline builders were genuinely skipped.
- stale-prepared fallback: mutates a session's text after `prepare_session_rows()` ran, asserts the writer detects the content-hash mismatch and stores the NEW content (never the stale prepared text).
- `content_hash=None` degrades safely (prepared rows never used, since the identity-only fallback hash can't coincide with a real content hash).
- `prepare_session_rows` purity: two calls on the same immutable session produce identical output.

**Micro-benchmark** (200 synthetic sessions, 80 messages each, mixed text/tool_use/thinking blocks, `/dev/shm`, timing only the `write_parsed_session_to_archive` call):

```
targeted stage (messages+blocks) inline:   3.048s / 2.922s / 3.738s  (3 runs)
targeted stage (messages+blocks) prepared: 2.945s / 2.529s / 3.303s
targeted stage reduction:                  3.4%   / 13.4%  / 11.6%
```

Reduction is real but modest and noisy at this synthetic corpus size — `sqlite3.Connection.executemany()` itself dominates wall time on tmpfs, so the CPU-bound row-building work this extracts is a genuine but not majority share of the targeted stage here. Real corpora with richer block content (larger `tool_input` JSON payloads, more messages per session, the whale pages the bead's own numbers cite) are expected to show a larger relative share of stage time moved off the writer thread, since row-building cost scales with content complexity while the SQLite insert cost scales roughly with row count.

Ref polylogue-623q

### DIFF
--- a/polylogue/storage/sqlite/archive_tiers/write.py
+++ b/polylogue/storage/sqlite/archive_tiers/write.py
@@ -205,6 +205,70 @@ class SessionEventWriteResult:
     wrote_provider_usage_events: bool = False
 
 
+@dataclass(frozen=True, slots=True)
+class PreparedSessionRows:
+    """Pure, off-writer-thread-computable row tuples for one session's
+    full-replace write (polylogue-623q).
+
+    Row PREPARATION -- converting a ``ParsedSession`` tree into the SQL row
+    tuples ``_replace_full_session_messages_and_blocks`` inserts -- was
+    measured happening entirely inside the writer hold (34-40s per 2,000
+    normal raws, 165-287s on whale pages) while parse-side warm workers sat
+    idle. ``prepare_session_rows`` builds this dataclass from nothing but a
+    ``ParsedSession`` (no DB connection, no archive state), so it can run on
+    a parse-prefetch worker thread; the writer then only needs to run
+    ``executemany`` against already-built tuples.
+
+    ``session_content_hash`` is ``pipeline.ids.session_content_hash(session)``
+    hex-decoded -- computed from the ORIGINAL (pre-lineage-slice) session, the
+    same value ordinary callers already pass as ``write_parsed_session_to_
+    archive(content_hash=...)``. The writer compares this against its own
+    ``content_hash`` argument before ever using the prepared rows: a session
+    whose content changed (or whose caller didn't supply a content_hash --
+    identity-only hashes never match) always falls back to preparing inline,
+    which reproduces the exact unmodified write path. A session that turns
+    out to need lineage tail-slicing (prefix-sharing composition against an
+    already-archived parent -- resolved by ``_extract_prefix_tail``, which
+    requires a live DB read the prefetch worker never had) is a SEPARATE
+    rejection condition the writer checks independently, because slicing
+    changes which messages are written without changing the session's own
+    content hash.
+    """
+
+    session_id: str
+    session_content_hash: bytes
+    message_rows: tuple[tuple[object, ...], ...]
+    block_rows: tuple[tuple[object, ...], ...]
+
+
+def prepare_session_rows(session: ParsedSession) -> PreparedSessionRows:
+    """Build ``PreparedSessionRows`` for ``session``'s full-replace write.
+
+    Pure function: normalizes messages exactly as ``write_parsed_session_to_
+    archive`` does for a non-merge-append, non-lineage-sliced write (see
+    ``_normalized_messages``), then reuses the same row-tuple builders the
+    writer itself calls (``_build_message_rows``/``_build_block_rows``) at
+    ``position_offset=0`` -- the offset every full-replace write uses. No
+    SQLite connection, network call, or filesystem access; safe to call from
+    any thread, including a parse-prefetch worker running well before (and
+    concurrently with) any writer hold.
+    """
+    from polylogue.pipeline.ids import session_content_hash as _compute_session_content_hash
+
+    origin = origin_from_provider(session.source_name)
+    session_id = archive_session_id(origin.value, session.provider_session_id)
+    messages = _normalized_messages(session.messages)
+    duplicate_native_ids = _duplicate_message_native_ids(messages)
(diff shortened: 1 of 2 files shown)


Sinity:
Coordinator review stand-in (CodeRabbit rate-limited on this PR — no findings were produced).

- **Extraction is behavior-preserving**: `_build_message_rows`/`_build_block_rows` are verbatim lifts of the old inline `rows()` generators; `_write_messages`/`_write_blocks` keep identical signatures for all existing call sites and only add an optional `rows=` passthrough.
- **Reuse gate is conservative and correct**: prepared rows are used only when (a) not `merge_append` (prepared assumes `position_offset=0`), (b) `lineage_inheritance != "prefix-sharing"` (the single case where `_extract_prefix_tail` slices `messages` after preparation), and (c) the prepared content hash equals this call's `content_hash`. Identity-only hashes (`content_hash=None` callers) can never match a real content hash, so those callers are structurally safe. Any mismatch silently falls back to inline building — passing a stale `prepared` is never incorrect, only wasted work.
- **Tests are non-vacuous**: equivalence proven by full-table dump comparison across text/tool/thinking/duplicate-native-id corpora; genuine reuse proven by monkeypatching the inline builders to raise; stale-prepared fallback proven to reflect NEW content.
- **Honest scoping**: daemon prefetch-cache wiring is explicitly deferred (separate, larger change through the census/spill/replay pipeline) — documented, not silently dropped.

Verification per PR body: 126 tests green across the new file + revision-replay/write/lineage/bulk-build suites; `devtools verify --quick` exit 0. Merging on green quick-gate.

PR #3220, opened 2026-07-20
feat(devtools): monotonic per-row bd JSONL sync with receipts

Sinity:
## Summary

Extracts the existing bd-reimport guard's ad hoc timestamp comparison into a
reusable, receipted monotonic merge engine, and adds explicit `reconcile`/
`export` entry points for JSONL-sync flows the git hooks cannot see.

## Problem

`devtools/bd_reimport_guard.py` already snapshotted live bead state around
`post-checkout`/`post-merge` and restored anything bd's own blind-upsert
reimport clobbered, but it had no receipt/audit trail, no validation of the
JSONL it read, and no coverage for flows outside those two hooks. The
2026-07-15 planning-surface incident showed the gap concretely: a staged
`issues.jsonl` carried literal conflict markers; a later `bd export`
regenerated syntactically valid JSON that silently restored nine
independently-repaired Beads to stale revisions (reintroducing legacy
horizon labels) while both standing lints stayed green — whole-file JSON
validity was never proof that per-row synchronization was correct. Separately,
`git reset --hard` fires no bd hook at all, so a reverted
`.beads/issues.jsonl` sits live-hazardous until the *next* unrelated `bd`
invocation reimports it (bd reimports from file on every call).

## Solution

- `merge_rows` (new): compares every candidate row against live state by
  revision (`updated_at`, the only per-row freshness signal bd's export
  exposes) and classifies it `new`/`updated`/`equal`/`skipped_downgrade`/
  `conflicted`/`recovered_downgrade`. A row with no comparable revision on
  either side is never guessed at — it's reported `conflicted`, not merged.
  Ordinary sync can never silently apply a downgrade.
- `SyncReceipt`/`build_receipt`/`write_receipt`: every merge produces a
  machine-readable receipt (per-row outcomes, actor/reason, recovery flag,
  `is_clean`) written atomically under `.cache/bd-sync-receipts/` for a
  downstream policy consumer (polylogue-8jg9.1) to require instead of
  trusting bare exit status.
- `parse_and_validate_jsonl` / `atomic_write_jsonl`: refuse conflict-marker
  payloads, invalid JSON lines, missing/duplicate ids; write via
  temp+fsync+atomic-rename and refuse to overwrite a marker-bearing target.
- `cmd_check_and_repair` rewired onto `merge_rows` (behavior-preserving: same
  `bd import`/`bd export` subprocess calls, same restore set) but now also
  emits a receipt.
- `reconcile <candidate.jsonl>` (new): explicit, hand-invokable sync — run it
  right after `git reset --hard` (or against any other externally-produced
  candidate file) to fold genuinely newer live rows back in before anything
  else touches `bd`. Refuses downgrades by default; `--allow-downgrade`
  requires `--actor` and `--reason`, and every downgraded row is still named
  in the receipt.
- `export <target.jsonl>` (new): atomic, validated snapshot export.

Devtools catalog entry (`devtools workspace bead-reimport-guard`) and
generated `docs/devtools.md` updated to describe all four subcommands.

**Deferred / not implementable in this repo** (`bd` is an opaque external
binary with no source checked in here): AC3 (concurrent empty-bootstrap
serialization) and part of AC7 (killpoint regression across bd's own server
and embedded modes) require instrumenting bd's Dolt-backed import path
directly, which this repo cannot reach. AC6 (later-bookkeeping-mutation
preservation feeding 8jg9.1's post-union checks) is left to the 8jg9.1 policy
consumer, per the existing bead note splitting ownership between the two.

## AC matrix (polylogue-gxjh.1)

1. **Satisfied** — `merge_rows`/`SyncReceipt` classify stale/newer/equal/
   incomparable rows with ids + both revisions; ordinary sync cannot
   downgrade (`test_merge_rows_classifies_new_updated_equal_and_skips_downgrade`,
   `test_merge_rows_incomparable_revision_is_conflicted_not_guessed`).
2. **Satisfied** — `reconcile --allow-downgrade` requires `--actor`+`--reason`
   and records every downgraded row as `recovered_downgrade` in the receipt
   (`test_reconcile_allow_downgrade_requires_actor_and_reason`,
   `test_reconcile_allow_downgrade_with_actor_reason_applies_and_records_recovery`).
   Project/database/branch fingerprinting beyond `source`/`actor`/`reason` is
   not implemented — deferred, no evidence it's needed beyond git-hook
   context which the caller already controls via `--source`.
3. **Deferred** — concurrent empty-bootstrap serialization requires
   instrumenting bd's own Dolt DB locking, out of reach from this repo.
4. **Satisfied** — `atomic_write_jsonl`/`cmd_export`: temp+fsync+rename,
   validates parse/unique-ids/JSON before write, refuses to overwrite a
   marker-bearing target (`test_atomic_write_jsonl_refuses_marker_bearing_target`,
   `test_export_writes_atomically_and_refuses_marker_bearing_target`).
   Snapshot-consistency against a *changed-since-snapshot* target (optimistic
   concurrency beyond marker-detection) is not implemented — deferred.
5. **Satisfied** — `test_replay_2026_07_15_incident_preserves_nine_repaired_beads`
   directly replays the incident: a valid-but-stale whole-file replacement of
   nine repaired Beads fails ordinary sync (all nine `skipped_downgrade`)
   while a genuinely newer unrelated row still applies.
6. **Deferred to polylogue-8jg9.1** — receipt consumption by backlog/frontier
   checks is explicitly that bead's scope per its own notes ("this bead owns
   repo policy consumption... do not implement a second merge algorithm
   here"); this PR ships the receipt shape it needs to consume.
7. **Partially satisfied** — killpoints/non-progress covered at this repo's
   wrapper boundary (invalid-JSONL mid-reconcile returns receipt+exit 2,
   missing candidate file returns exit 1 without a receipt); true
   server/embedded-mode regression against bd's own binary is out of reach
   without its source.

## Verification

- `devtools test tests/unit/devtools/test_bd_reimport_guard.py` → 25 passed
- `ruff format --check` / `ruff check` on both touched files → clean
- `mypy devtools/bd_reimport_guard.py tests/unit/devtools/test_bd_reimport_guard.py` → no issues
- `devtools render devtools-reference && devtools render all --check` → exit 0, no "out of sync"
- `devtools verify --quick` → exit_code 0 (ran twice: pre-commit format/lint hook, pre-push full quick gate)

Not run: `devtools verify --all` (full suite) — out of scope for a
devtools-only change with a fully passing focused+quick gate.

Ref polylogue-gxjh.1


Sinity:

- **Merge engine is conservative by construction**: `merge_rows` only replaces a row on strictly-newer revision; incomparable rows (missing `updated_at` either side) are reported as `conflicted` and never guessed at; downgrades require `--allow-downgrade` + recorded `--actor`/`--reason` and remain named in the receipt. `cmd_check_and_repair` rewired onto the same engine is behavior-preserving (regressed rows → `updated`, vanished rows → `new`, legitimately-newer post-hook rows correctly untouched).
- **Validation refuses the actual 2026-07-15 incident shape**: conflict markers, invalid JSON lines, missing/duplicate ids all raise; `atomic_write_jsonl` additionally refuses to overwrite a marker-bearing target and round-trip-validates its own output before rename.
- **One noted limitation, acceptable**: revision ordering is lexicographic on `updated_at` strings — correct for bd's consistent RFC3339-UTC export format; mixed-timezone-offset timestamps would mis-order, but no such producer exists in this pipeline.
- **Stdlib-only constraint preserved** (hook-invoked by file path outside the venv) — verified no non-stdlib imports added.
- **AC honesty**: deferred items (bd-internal Dolt locking, receipt consumption in 8jg9.1) are genuinely out of this repo's reach or explicitly owned elsewhere per the bead's own notes.

Verification per PR body: 25/25 focused tests (incl. direct nine-bead incident replay), ruff/mypy clean, `render all --check` no out-of-sync, quick-gate green. Merging.

PR #3221, opened 2026-07-20
fix(demo): stop demo seed parsing from spawning a real process pool

Sinity:
## Summary

The demo seeder now forces sequential (single-worker) parsing instead of routing its fixed, small fixture corpus through a real subprocess `ProcessPoolExecutor`.

## Problem

polylogue-b054.1.1.1 observed a nondeterministic demo-archive failure under an 8-worker xdist run: `tests/unit/demo/test_demo_seed_verify.py::test_demo_verify_reports_missing_overlays` reported at least one declared construct below its minimum, but the exact node and its same-worker predecessors pass in isolation — load/order/leaked-process-state, not a deterministic mismatch.

`seed_demo_archive` (`polylogue/demo/seed.py`) calls `parse_sources_archive` without pinning a worker count, so the demo's ~9 sources / couple dozen files went through `parse_sources_archive`'s ambient `ProcessPoolExecutor` path (default up to `cpus-1` real spawn subprocesses, `polylogue/pipeline/services/process_pool.py`). Under pytest-xdist, every one of the (up to 8) test worker processes independently spawns its own sub-pool — up to dozens of concurrent subprocesses competing for CPU/memory on the host for a corpus that pooling was never meant to help (pooling exists to overlap CPU-bound parse with I/O-bound writes on real bulk archives, `polylogue/pipeline/services/archive_ingest.py:59-65`).

A per-file worker task failure under that load (`BrokenProcessPool`, transient spawn/OOM) is caught in `parse_sources_archive`'s `as_completed` loop by `except Exception: failed += 1; ...continue` (archive_ingest.py:299-306) — silently dropping that file's sessions with only a `logger.error` line, no raised error. This is a plausible, reproducible mechanism for exactly the observed symptom: a demo fixture file's content occasionally vanishing under host load with no test-visible signal, until construct-coverage happened to depend on the missing content.

## Solution

- `polylogue/pipeline/services/archive_ingest.py`: `parse_sources_archive` gains an optional `parse_workers: int | None = None` keyword that overrides the ambient/env-resolved worker count for one call. `None` preserves existing behavior for every other caller (real ingest, daemon, other tests).
- `polylogue/demo/seed.py`: `seed_demo_archive` now calls `parse_sources_archive(..., parse_workers=1)`, so demo seeding always takes the sequential for-loop branch and never constructs a subprocess pool for this fixed-size synthetic corpus — removing the host-load dependency entirely for this path.

## Verification

- `devtools test tests/unit/demo/test_demo_seed_verify.py tests/unit/pipeline/test_archive_ingest_commit_batching.py tests/unit/scenarios/test_demo_archive_convergence.py` → **20 passed**
- Two new regression tests exercise the real production route (not mocks of their own claim):
  - `test_parse_workers_override_bypasses_process_pool` spies on the actual `ProcessPoolExecutor` construction inside `parse_sources_archive`: `parse_workers=1` never constructs it (sequential branch), `parse_workers=3` constructs it with exactly `max_workers=3`. **Anti-vacuity**: reverting the `max(1, parse_workers)` override in `parse_sources_archive` back to always calling `_parse_worker_count()` makes this test fail with `TypeError: parse_sources_archive() got an unexpected keyword argument 'parse_workers'` (confirmed by temporarily reverting and re-running).
  - `test_seed_demo_archive_forces_sequential_parse_workers` spies on the demo seeder's real `parse_sources_archive` call via a delegating wrapper (archive tiers are still populated by genuine ingest) and asserts `parse_workers=1` is always passed. **Anti-vacuity**: reverting the `seed.py` call site fails this test with `assert None == 1` (confirmed by temporarily reverting and re-running).
- `tests/unit/demo/test_demo_seed_verify.py` repeated 3× under `POLYLOGUE_PYTEST_WORKERS=8` xdist → 9/9 passed each run (no flakes observed).
- `mypy --strict` on all 4 touched files → `Success: no issues found in 4 source files`.
- `devtools verify --quick` → `exit_code: 0` (all 16 static/generated gates green, including `ruff format`/`ruff check`, `render all`, topology/layering/closure-matrix, schema roundtrip, manifests, ci-workflows, doc-commands, docs-coverage, test-infra-currency, test-clock-hygiene, pytest-timeout-overrides, degrade-loudly).
- Pre-push hook's own `devtools verify --quick` baseline ran again and passed (`exit_code: 0`).
- **Not run**: the parent bead's (polylogue-b054.1.1) full 20-isolated + 20-xdist repetition proof and two-consecutive-8-worker-seed-testmon-run proof (AC4/AC5 of that broader bead) — out of scope for this narrower child bead's root-cause fix. The demo test file itself was run repeatedly under xdist as targeted evidence the specific mechanism identified here is addressed; it does not substitute for the parent bead's broader harness-level proof obligation.

## Residual risk

This fixes the demo-seeding-specific instance of the swallowed-pool-failure hazard. The underlying `except Exception: failed += 1; continue` pattern in `parse_sources_archive`'s pool branch still silently drops content on any real-archive ingest under the same failure conditions (BrokenProcessPool, worker OOM) — that is intentional production resilience (one bad file shouldn't kill a bulk ingest run) and out of this bead's scope, but it means the same *class* of silent-drop could still occur for real (non-demo) multi-worker ingests under host pressure. No new tracking item filed for that, since it is an existing, documented (log-line) behavior, not something this change introduces or regresses.

Ref polylogue-b054.1.1.1

### DIFF
--- a/polylogue/demo/seed.py
+++ b/polylogue/demo/seed.py
@@ -1293,7 +1293,24 @@ async def seed_demo_archive(
 
     source_root = materialize_demo_source(archive_root, force=force)
     with _pushd(source_root):
-        result = await parse_sources_archive(archive_root, demo_source_specs(source_root))
+        # Force sequential (single-worker) parsing for the fixed, dozen-file
+        # synthetic demo corpus. `parse_sources_archive`'s ambient worker
+        # count defaults to up to `cpus-1` real OS subprocesses (spawn-based
+        # ProcessPoolExecutor); at demo scale that pool buys zero throughput
+        # (it exists to overlap CPU-bound parse with I/O-bound writes on real
+        # bulk archives) but multiplies badly under pytest-xdist, where every
+        # worker process independently spawns its own sub-pool. Under host
+        # load a per-file worker task can fail (BrokenProcessPool, transient
+        # spawn/OOM) and is caught by the pool driver's `except Exception:
+        # failed += 1; continue` -- silently dropping that file's sessions
+        # with only a log line, no raised error. That is the root cause of
+        # the observed nondeterministic declared-construct loss under xdist
+        # (polylogue-b054.1.1.1): a demo fixture file occasionally vanished
+        # under load with no test-visible signal until construct-coverage
+        # happened to depend on it. Sequential parsing removes the pool
+        # entirely for this fixed-size corpus, so demo seeding no longer
+        # depends on host process/memory pressure at all.
+        result = await parse_sources_archive(archive_root, demo_source_specs(source_root), parse_workers=1)
 
     apply_demo_post_ingest_augmentation(archive_root)
 
--- a/polylogue/pipeline/services/archive_ingest.py
+++ b/polylogue/pipeline/services/archive_ingest.py
@@ -101,7 +101,9 @@ def _parse_source_path_worker(
         publisher.discard_pending()
 
 
-async def parse_sources_archive(archive_root: Path, sources: list[Source]) -> ParseResult:
+async def parse_sources_archive(
+    archive_root: Path, sources: list[Source], *, parse_workers: int | None = None
+) -> ParseResult:
     """Parse configured sources directly into archive source/index tiers.
 
     Source files are parsed across a process pool (``POLYLOGUE_INGEST_PARSE_WORKERS``)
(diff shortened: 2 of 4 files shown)


Sinity:

- **Root cause is real and the fix removes it structurally**: demo seeding routed a ~dozen-file fixed corpus through the ambient spawn-based ProcessPoolExecutor (up to cpus-1 workers), and the pool driver's `except Exception: failed += 1; continue` silently drops a failed file's sessions — under 8-worker xdist each test worker spawned its own sub-pool, making transient BrokenProcessPool/spawn failures both likely and invisible. Pinning `parse_workers=1` takes the sequential branch: no pool is ever constructed, so demo determinism no longer depends on host pressure.
- **Override is surgically scoped**: keyword-only param, `None` preserves the exact prior env/cpu resolution for every real ingest caller; `max(1, ...)` guards nonsense values.
- **Tests are non-vacuous**: the pool-bypass test spies on the actual ProcessPoolExecutor construction and both branches are exercised (`parse_workers=1` → no pool; `parse_workers=3` → pool with exactly 3); the seeder test spies through a delegating wrapper so the genuine ingest route still runs. Both confirmed failing against pre-fix code per PR body.
- **AC honesty**: the bead's broader stress-harness/20+20-rep proof is explicitly reported as not run — that belongs to the parent harness scope; noted in the bead close.

Verification per PR body: 20 focused tests green, 3× repeated 8-worker xdist runs of the demo file 9/9 each, mypy --strict clean, quick-gate green. Merging.

PR #3222, opened 2026-07-20
fix(demo): force sequential parsing for deterministic demo seed

Sinity:
## Summary

Forces the demo seeder's own `parse_sources_archive` call to sequential
(`workers=1`) parsing so it structurally can never engage the process-pool
parallel path -- eliminating a real, evidence-backed nondeterminism source
for demo construct coverage under host load (e.g. 8-way xdist).

## Problem

A repeated clean 8-worker testmon seed on 2026-07-16 failed once (Ref
polylogue-b054.1.1.1) after two green runs:
`test_demo_seed_verify.py::test_demo_verify_reports_missing_overlays`
observed at least one declared demo construct below its minimum. The exact
node and its immediate same-worker predecessors passed in isolation, ruling
out a deterministic assertion mismatch and pointing at load/ordering/process
lifecycle.

Reading `parse_sources_archive` (`polylogue/pipeline/services/archive_ingest.py`)
found the mechanism: its parallel branch submits one `ProcessPoolExecutor`
job per source file and, in the `as_completed` loop, catches any exception
from `future.result()`, increments a failure counter, logs a warning, and
*continues* -- "one bad file must not kill the run." That is the right
tradeoff for a large, uncontrolled real corpus, but the demo seeder feeds it
a tiny, fully-controlled fixture corpus (~a dozen files) where every file's
content is required for every declared construct's minimum count. A
transient process-pool lifecycle failure on exactly one file -- a
spawn/fork error or worker crash under host resource pressure, exactly what
concurrent xdist workers can cause -- silently drops that file with no
signal beyond an unrelated downstream construct-coverage assertion. This
matches the bead's diagnosis hint ("process-pool completion") and the
observed shape (non-reproducible, load-dependent, single occurrence).

## Solution

- `parse_sources_archive` gains an explicit keyword-only `workers: int |
  None = None` override (`None` preserves the existing env/CPU-derived
  default; every other call site is unaffected).
- `seed_demo_archive` now always passes `workers=1`, so the demo path never
  constructs a `ProcessPoolExecutor` regardless of host CPU count,
  `POLYLOGUE_INGEST_PARSE_WORKERS`, or concurrent xdist load. It structurally
  cannot hit the isolation-vs-completeness tradeoff described above.
- Added `tests/unit/demo/test_demo_seed_worker_isolation.py` with two
  fault-injection tests:
  1. `test_parallel_parse_worker_failure_undercounts_a_declared_construct`
     exercises the real parallel path with an injected single-file worker
     failure and proves, via the real `evaluate_demo_constructs` SQL, that
     `browser_capture_raw_variants` falls from 3 to 2 (below its declared
     minimum) -- the exact vulnerability mechanism, not a toy replica.
  2. `test_seed_demo_archive_never_engages_the_process_pool` monkeypatches
     `ProcessPoolExecutor` to raise on construction and runs the real
     `seed_demo_archive` under a provocative
     `POLYLOGUE_INGEST_PARSE_WORKERS=8` override, proving the fix. Anti-vacuity:
     reverting the `workers=1` change (verified manually) makes this test fail
     with the injected error instead of completing.

**Considered and rejected**: also switching `archive_ingest.py`'s raw
`ProcessPoolExecutor` to the shared spawn-context `process_pool_executor()`
helper used by every other process-pool call site in the codebase (a
plausible *additional* lifecycle hardening, since the raw call bypasses the
deliberate `spawn` context choice documented in `process_pool_context()`).
This broke
`test_process_pool_reingest_reserves_before_publish_and_consumes_with_source_ref`,
which relies on the *ambient default* context being `fork` on Python 3.13 so
its own fork-pinned observer process can see `BlobStore`/`ArchiveStore`
methods monkeypatched in the parent test process (inherited via `fork`,
impossible via `spawn`). That is a real, pre-existing test dependency on the
current context choice, orthogonal to this bead's scope (demo determinism,
not general process-pool context hygiene) -- reverted, left unchanged.

## Verification

- `devtools test tests/unit/demo/` -- 41 passed
- `devtools test tests/unit/pipeline/test_archive_ingest_shared_raw.py tests/unit/pipeline/test_archive_ingest_commit_batching.py tests/unit/pipeline/test_process_pool.py tests/unit/scenarios/test_demo_archive_convergence.py tests/unit/demo/` -- 58 passed (confirms no regression once the considered-and-rejected `process_pool_executor` swap was reverted)
- 5 consecutive `devtools test tests/unit/demo/ -- -n 8` runs, all green (41/41 each)
- `mypy --strict` on all three touched files -- no issues
- `devtools verify --quick` -- exit_code 0 (also ran automatically via the pre-push hook)

**Not run** (time-bounded in this session, see Residual risk): the bead's
full 20-isolated + 20-xdist repetition of the exact originally-failing node,
and two consecutive full 8-worker `devtools verify --seed-testmon
--skip-slow` runs.

## Residual risk / AC honesty

Ref polylogue-b054.1.1.1 acceptance criteria:

1. "A failing repetition reports the exact construct id..." -- **satisfied
   indirectly**: the fault-injection test reports exact construct id/observed/
   minimum for the reproduced mechanism, rather than instrumenting a live
   recurrence (none was captured this session).
2. "The evidence harness reproduces the prior failure or proves the
   implicated lifecycle race with a deterministic fault injection." --
   **satisfied**: `test_parallel_parse_worker_failure_undercounts_a_declared_construct`
   is exactly this deterministic fault injection, against the real
   production route.
3. "The underlying production/fixture lifecycle is repaired without
   retries, sleeps, test quarantine, or weakening construct coverage." --
   **satisfied**: `workers=1` is a structural elimination of the risk
   surface (no ProcessPoolExecutor is ever constructed for the demo path),
   not a retry/sleep/quarantine, and construct coverage requirements are
   unchanged.
4. "The exact node passes 20 isolated and 20 xdist repetitions, and two
   consecutive 8-worker `devtools verify --seed-testmon --skip-slow` runs
   are green." -- **deferred**: only 5 consecutive 8-worker
   `tests/unit/demo/` runs were performed in this session (time-bounded);
   the full 20+20 repetition count and the two full seed-testmon runs were
   not executed. Given the fix is a structural elimination (the failure
   mode requires `ProcessPoolExecutor` to be constructed at all, which the
   demo path now never does), the residual risk is that the original,
   never-reproduced 2026-07-16 failure had a *different* root cause that
   this change does not address. The maintainer should run the full
   repetition/seed proof before closing the bead, or accept the structural
   argument plus this session's 5x confirmation as sufficient.
5. "Cleanup receipts show no surviving process groups or temp roots." --
   **satisfied for this session's runs**: `devtools test` containment
   reports/`devtools verify --quick` completed cleanly with no reported
   leftover processes or temp roots.

Ref polylogue-b054.1.1.1

### DIFF
--- a/polylogue/demo/seed.py
+++ b/polylogue/demo/seed.py
@@ -1293,7 +1293,21 @@ async def seed_demo_archive(
 
     source_root = materialize_demo_source(archive_root, force=force)
     with _pushd(source_root):
-        result = await parse_sources_archive(archive_root, demo_source_specs(source_root))
+        # Force sequential (single-worker) parsing: the demo corpus is a
+        # handful of small, fully-controlled fixture files, so the
+        # process-pool parallel path buys no measurable throughput but
+        # exposes the fixed, small construct set to
+        # ``parse_sources_archive``'s deliberate worker-error-isolation
+        # behavior (one file's parse failure -- including a transient
+        # process-pool lifecycle failure under host resource pressure, e.g.
+        # under an 8-way xdist run -- is silently skipped rather than
+        # failing the whole ingest). A skipped demo fixture file
+        # nondeterministically undercounts a declared construct (e.g. the
+        # 3-variant browser-capture coalescing set) with no visible signal
+        # until an unrelated downstream assertion fails (polylogue-b054.1.1.1).
+        # The demo seeder needs every file every time, so it opts out of
+        # that isolation guarantee entirely instead of tolerating it.
+        result = await parse_sources_archive(archive_root, demo_source_specs(source_root), workers=1)
 
     apply_demo_post_ingest_augmentation(archive_root)
 
--- a/polylogue/pipeline/services/archive_ingest.py
+++ b/polylogue/pipeline/services/archive_ingest.py
@@ -101,7 +101,12 @@ def _parse_source_path_worker(
         publisher.discard_pending()
 
 
-async def parse_sources_archive(archive_root: Path, sources: list[Source]) -> ParseResult:
+async def parse_sources_archive(
+    archive_root: Path,
+    sources: list[Source],
+    *,
+    workers: int | None = None,
+) -> ParseResult:
     """Parse configured sources directly into archive source/index tiers.
 
     Source files are parsed across a process pool (``POLYLOGUE_INGEST_PARSE_WORKERS``)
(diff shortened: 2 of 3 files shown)


Sinity:
Superseded: polylogue-b054.1.1.1 was delivered and merged via PR #3221 (squash 33533762a) from the direct replacement lane — same root cause (demo seeding through the ambient spawn pool with swallowed per-file worker failures), same fix shape (parse_workers=1 pin + pool-bypass regression tests). Bead closed on those receipts. Closing this duplicate.

PR #3223, opened 2026-07-20
fix(sources): corroborate catch-up cursor skips against index presence

Sinity:
## Summary

Watcher catch-up's hot-skip fast path trusted `ops.db` ingest cursors as
proof of full materialization without ever checking whether `index.db`
actually holds the corresponding session. This PR adds a corroboration
check that demotes cursor-trust skips to "needed" when the index tier
cannot back up the claim.

## Problem

polylogue-emx2 (Finding 8 of perf-investigation-2026-07-18): after an
index reset/rebuild, `source.db` retained 14,879 ingest cursors
recording "acquired and parsed", but `index.db` started empty.
`_needs_work_from_state`'s hot-skip fast path (byte-size +
content-fingerprint match) never checked whether the index actually
held a materialized session for the cursor's raw — it skipped every
one of those files outright, leaving 100% of the restore drain to the
1-per-30s background conveyor.

PRs #3129 and #3193 (merged, both linked to this bead) fixed a related
but distinct bug: `batch.py` was misclassifying a raw with a
pending/decided membership arbitration as a full-ingest *failure*. That
repaired the accounting after a file was resubmitted for ingest, but
never addressed the original finding — the cursor-trust fast path in
the watcher still never consulted the index at all. This PR closes
that remaining gap.

## Solution

`polylogue/sources/live/watcher.py`:
- `_archived_cursor_reconciliation_scope` now computes, once per
  catch-up planning pass (reusing its already-shared source.db/index.db
  connections), whether the index tier is globally uncorroborated:
  `_index_lacks_all_corroboration` returns `True` when `source.db`
  holds at least one successfully parsed raw row but `index.db`'s
  `sessions` table is completely empty — the exact post-reset
  signature from Finding 8.
- `_needs_work_from_state` (the prior single-purpose decision method,
  renamed `_needs_work_from_state_uncorroborated`) now wraps that
  logic: when it would skip a file and the index is flagged
  uncorroborated, a cheap per-candidate existence check
  (`_cursor_skip_corroborated_by_index`, reusing the existing
  `_archived_cursor_row` raw→session join) decides whether to honor the
  skip or demote it to "needed".
- The gate defaults to trusted (no behavior change) whenever the index
  has any session at all, so the common healthy case adds zero extra
  queries.
- Demoted files re-enter the normal batch ingest path, which is already
  content-hash idempotent, so no bytes are re-acquired for unchanged
  files.

## Verification

- `devtools test tests/unit/sources/test_live_catchup_planning.py`:
  19 passed (16 pre-existing + 3 new).
- `devtools test tests/unit/sources/test_live_watcher.py
  tests/unit/sources/test_live_watcher_catchup_order.py
  tests/unit/sources/test_live_watcher_locking.py`: 101 passed.
- `mypy --strict polylogue/sources/live/watcher.py
  tests/unit/sources/test_live_catchup_planning.py`: no issues found.
- `devtools verify --quick`: exit 0.
- Anti-vacuity: reverted only the production fix (kept the new tests)
  and reran the suite — `test_catch_up_demotes_hot_skip_cursor_when_index_holds_no_sessions`
  failed with the exact live symptom (`plan.needed == ()` instead of
  the healthy-cursor file), and
  `test_catch_up_index_lacks_all_corroboration_detects_empty_index_with_parsed_raw`
  failed with `AttributeError` for the missing method. Restoring the
  fix turned both green.
- Not run: full `devtools verify --all` / integration suite (out of
  scope for a focused fix; the touched-file testmon selection above is
  the relevant net).

Ref polylogue-emx2

### DIFF
--- a/polylogue/sources/live/watcher.py
+++ b/polylogue/sources/live/watcher.py
@@ -197,6 +197,12 @@ class LiveWatcher:
         self._stop = asyncio.Event()
         self._catch_up_complete = asyncio.Event()
         self._archived_cursor_conns: tuple[sqlite3.Connection, sqlite3.Connection] | None = None
+        # Set once per reconciliation scope: True when the index tier has no
+        # materialized sessions at all despite source.db holding successfully
+        # parsed raw material -- the post-index-reset signature (polylogue-emx2).
+        # While true, cursor-trust skip decisions must be corroborated
+        # per-candidate against index presence instead of being taken on faith.
+        self._archived_cursor_index_untrusted = False
         self._batch_processor = LiveBatchProcessor(
             polylogue,
             self._sources,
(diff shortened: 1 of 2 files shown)


Sinity:

- **Fixes the actual Finding-8 gap**: the cursor-trust fast path in `_needs_work_from_state` never consulted index.db; #3129/#3193 fixed a downstream misclassification, not this. The wrapper design (delegate byte/fingerprint decision, then corroborate skips only) leaves every 'needed' decision untouched.
- **Zero-cost healthy path**: the global gate `_index_lacks_all_corroboration` is two LIMIT-1 existence probes once per catch-up scan, false whenever the index holds any session — per-file corroboration only runs in the post-reset signature state.
- **Fail-safe direction is right**: transient sqlite errors during corroboration return True (no spurious re-ingest); paths with no parsed raw are treated as corroborated (unrelated cursor kinds unaffected); demotion re-enters the normal content-hash-idempotent ingest path, so no re-acquisition of unchanged bytes and no cursor deletion.
- **Known limitation, acceptable**: gate is global (index fully empty), not per-source — matches the live incident (14,879 cursors vs empty index); a partial per-source gap is a natural follow-up, honestly flagged.
- **Anti-vacuity proven**: production fix reverted with tests kept → exact live symptom reproduced (`plan.needed == ()`).

Verification per PR body: 19+101 focused tests green, mypy --strict clean, quick-gate green. Merging.

PR #3224, opened 2026-07-20
fix(sources): profile-qualify Hermes ATIF/ATOF observer session identity

Sinity:
## Summary

Fixes the Hermes ATIF/ATOF observer-evidence session-identity collapse described in polylogue-fs1.14 by threading a profile-qualified join key through the ATIF/ATOF parsers, and by asserting a real `session_links` parent edge back to the state-db conversational session instead of leaving that correlation as a read-side-only naming convention.

## Problem

Two independent Hermes installs (profiles) can legitimately reuse the same raw Hermes session id — `hermes_state.py`'s own state.db parser already accounts for this by profile-qualifying its conversational session id (`<raw_id>@profile-<key>`, hashed from the install root directory). The ATIF/ATOF observer-evidence session identity in `hermes_spans.py` did not: it was always `observer:<raw_id>`, with no profile qualifier at all, so ATIF/ATOF evidence from two different installs sharing a raw session id silently collapsed onto one archive session.

This was compounded by the existing read-side correlation helper, `hermes_observer_session_id_for`: given a real qualified conversational session id, it stripped the `@profile-<key>` suffix before building the observer id — discarding the exact information that would have prevented the collision, even for the one cross-artifact join path that already existed.

Separately, neither ATIF nor ATOF ever asserted a `session_links` parent edge back to the state-db conversational session — the module docstring explicitly called this out as deferred ("NOT materialized into topology_edges/session_links in this pass").

## Solution

- Extracted the profile-key hashing/qualification helpers (previously private, duplicated only in `hermes_state.py`) into a new shared module, `polylogue/sources/parsers/hermes_identity.py` (`profile_key`, `qualified_session_id`, `split_qualified_session_id`). Both `hermes_state.py` and `hermes_spans.py` now compute the join key from this single source of truth.
- `parse_atif_document`/`parse_atof_stream` (`hermes_spans.py`) now accept an optional `profile_root: Path | None` — the directory the raw artifact was acquired from, mirroring `hermes_state.parse_state_db`'s own convention. When known, both the observer session identity (`observer:<raw_id>@profile-<key>`) and a `parent_session_provider_id` (`<raw_id>@profile-<key>`, matching the conversational session's own id) are asserted.
- `polylogue/sources/dispatch.py`: wired `profile_root=Path(source_path).parent` at all three call sites (`_parse_lowered_spec`'s ATOF/ATIF branches and `parse_stream_payload`'s Hermes branch), deriving it from the already-threaded `source_path`/`spec.source_path`.
- Once `parent_session_provider_id` is asserted, the existing, unmodified generic `session_links` write/resolution machinery in `archive_tiers/write.py` picks it up: it resolves the edge once the target conversational session is ingested, or leaves it explicitly `unresolved` (visible debt, never silently dropped) otherwise. No changes were made to that file (owned by another lane).
- **Fail-closed**: when `profile_root` is `None` (unknown at parse time), no parent edge is asserted and the observer session id stays unqualified — a wrong guess would silently correlate evidence with the wrong install, which is worse than an unresolved correlation.
- Added a `parent_session_link` fidelity capability to both ATIF's and ATOF's `import_fidelity_declaration` output, reflecting whether the parent edge was asserted (`inferred`) or withheld (`absent`).

### Scope notes / explicitly not done

- `hermes_verification.py` (the separate verification-ledger artifact family) has the identical unqualified-id / qualifier-stripping pattern, but is not named in this bead's design (`fs1.14` is scoped to "ATIF+ATOF"). Left as a follow-up rather than expanding scope in this PR.
- Subagent-delegation `topology_edges` materialization from `subagent_trajectories` (ATIF) / `hermes.subagent.*` marks (ATOF) remains explicitly deferred/absent, consistent with the existing fidelity declarations. No real `subagent_trajectories` evidence has been observed to verify that shape (per the existing module docstring), and materializing it would require the *child* session to assert its own parent (the pattern every other session_links producer in this archive uses) — the parent-side ATOF/ATIF document only knows the child's id, not the reverse, and asserting on the child's behalf would need new write-path machinery outside `archive_tiers/write.py`'s current generic contract, which is out of this lane's scope.
- "Conflicting parent identifier fails closed or renders an explicit conflict" (bead AC): satisfied by the existing, generic `session_links` resolution machinery, unchanged here — an asserted `dst_native_id` that never resolves stays `unresolved` forever (visible debt), and cycle detection/quarantine is already generic across all parsers. This PR's job was correctly asserting the join key so that machinery can act on it, not duplicating it.

## Verification

- `devtools test tests/unit/sources/parsers/test_hermes_spans.py tests/unit/sources/parsers/test_hermes_state.py tests/unit/sources/parsers/test_hermes_lifecycle.py tests/unit/sources/test_hermes_import_explain.py tests/unit/sources/test_hermes_export_contract.py tests/unit/pipeline/test_hermes_raw_identity.py tests/unit/sources/test_live_watcher.py tests/unit/sources/test_live_watcher_catchup_order.py tests/unit/sources/test_dispatch_payloads.py tests/unit/sources/test_provider_completeness.py tests/unit/sources/test_hermes_source_freshness_integration.py` — 167 passed.
- `mypy --strict` on all touched source files — no issues found.
- `devtools verify --quick` — exit 0 (ran twice: once locally after `render topology-projection`/`render topology-status` regen for the new `hermes_identity.py` module, once again via the pre-push hook).
- Not run: `devtools verify --all` / the full non-integration suite — left to CI per repo policy (per-PR CI intentionally skips the heavy `test` suite; it runs post-merge on master).

Ref polylogue-fs1.14


Sinity:

- **Real identity bug, correctly rooted**: observer-evidence sessions used bare `observer:<raw_id>` while conversational sessions were profile-qualified (`<raw_id>@profile-<key>`); `hermes_observer_session_id_for` even stripped the qualifier before building the observer id. Two Hermes profiles sharing a raw session id collapsed onto one archive session — exactly the bead's named failure.
- **Single source of truth**: new `hermes_identity.py` extracts the qualifier helpers; `hermes_state.py` switched to it (pure dedup); `hermes_spans.py` consumes it for both the observer id and the new `parent_session_provider_id` edge.
- **Fail-closed on unknown profile**: parent edge asserted only when `profile_root` is known (dispatch wires `Path(source_path).parent` at all three ATIF/ATOF call sites); `None` → no edge, tested explicitly. Link resolution rides the existing generic `session_links` path — archive_tiers/write.py untouched, respecting hot-file ownership.
- **Anti-vacuity**: `test_two_profiles_with_the_same_raw_session_id_do_not_collapse` fails if qualification is reverted.
- **Topology projection regenerated** for the new module (render all --check green — the gotcha that usually bites here).
- **Residual flagged, agreed**: `hermes_verification.py` has the same collapse-and-strip pattern, outside this bead's ATIF/ATOF scope — should get a follow-up bead; subagent-topology materialization stays deferred consistent with existing fidelity declarations.

Verification per PR body: 167 focused tests green across 11 files, mypy --strict clean, quick-gate green (2×). Merging.

PR #3227, opened 2026-07-20
fix(sources): profile-qualify Hermes verification-ledger session identity

Sinity:
## Summary

`hermes_verification.py` retained the exact unqualified observer-id collapse
pattern that PR #3224 fixed for ATIF/ATOF (`hermes_spans.py`). This PR
mirrors that fix for the Hermes verification-ledger importer.

## Problem

Two Hermes installs (profiles) can legitimately reuse the same raw session
id. `hermes_verification.observer_session_provider_id` /
`hermes_verification_session_id_for` built a bare `verification:<raw_id>`
identity and stripped any `@profile-<key>` qualifier when mapping back from
a conversational session id, so two installs' verification-ledger evidence
(lint/typecheck/build/test command outcomes) silently collapsed onto one
archive session.

## Solution

- `observer_session_provider_id(hermes_session_id, profile_key=None)` and
  `hermes_verification_session_id_for` now thread an optional profile key
  through `hermes_identity.qualified_session_id` /
  `split_qualified_session_id` -- the exact shared helper `hermes_spans.py`
  and `hermes_state.py` already use. The `verification:` family prefix stays
  distinct from ATIF/ATOF's `observer:` family even when both qualify the
  same raw id against the same profile.
- `parse_verification_evidence_db` / `parse_verification_evidence_db_payload`
  accept `profile_root: Path | None`, compute a profile key when known, and
  assert `parent_session_provider_id` back to the state-db conversational
  session only when the profile root is known (fail-closed: no edge is
  safer than a wrong one).
- Added a `parent_session_link` fidelity capability, aggregated across
  sessions, mirroring `hermes_spans._parent_session_link_capability`.
- Wired `profile_root=Path(spec.source_path).parent` (or `path.parent`) at
  every verification-evidence call site: `dispatch.py`, `import_explain.py`,
  `source_parsing.py`, `live/batch.py`.
- Left `revision_backfill.py`'s existing call untouched per scope (it already
  fails closed with `profile_root` defaulting to `None`).
- Did not touch `hermes_spans.py` / `hermes_identity.py` (consumed
  read-only) or the ATIF/ATOF call sites -- those are owned by another
  active lane.

## Verification

- `devtools test tests/unit/sources/parsers/test_hermes_verification.py tests/unit/sources/test_dispatch_payloads.py tests/unit/sources/test_hermes_import_explain.py tests/unit/sources/test_hermes_source_freshness_integration.py tests/unit/insights/test_hermes_verification_coverage.py tests/unit/sources/test_live_watcher_catchup_order.py tests/unit/core/test_sampling.py tests/unit/sources/test_parsers_local_agent.py` -> `103 passed`
- `mypy --strict` on the 5 touched production modules -> `Success: no issues found in 5 source files`
- `devtools verify --quick` -> `exit_code: 0`
- **Anti-vacuity**: temporarily reverted the profile-key qualification in
  `observer_session_provider_id` (collapsed back to unconditional
  `verification:<raw_id>`) and re-ran the new collapse/family-prefix/
  dispatch-wiring tests -- all 3 failed with the expected unqualified-id
  assertion mismatches. Restored the fix; tests green again.

Ref polylogue-y9zx

### DIFF
--- a/polylogue/sources/dispatch.py
+++ b/polylogue/sources/dispatch.py
@@ -975,7 +975,11 @@ def _parse_lowered_spec(spec: LoweredPayloadSpec) -> list[ParsedSession]:
         if spec.provider is Provider.HERMES and hermes_state.looks_like_state_db_payload(record):
             return hermes_state.parse_state_db_payload(record, spec.fallback_id)
         if spec.provider is Provider.HERMES and hermes_verification.looks_like_verification_evidence_db_payload(record):
-            return hermes_verification.parse_verification_evidence_db_payload(record, spec.fallback_id)
+            return hermes_verification.parse_verification_evidence_db_payload(
+                record,
+                spec.fallback_id,
+                profile_root=Path(spec.source_path).parent if spec.source_path else None,
+            )
         if spec.provider is Provider.HERMES and hermes_spans.looks_like_atif_payload(record):
             return [
                 hermes_spans.parse_atif_document(
--- a/polylogue/sources/import_explain.py
+++ b/polylogue/sources/import_explain.py
@@ -490,7 +490,9 @@ def _explain_hermes_verification_evidence_db(path: Path, *, provider_hint: Provi
     """Inspect the real Hermes verification-ledger SQLite parser path without writing a raw blob."""
 
     try:
-        sessions = hermes_verification.parse_verification_evidence_db(path, fallback_id=path.stem)
+        sessions = hermes_verification.parse_verification_evidence_db(
+            path, fallback_id=path.stem, profile_root=path.parent
+        )
     except (OSError, sqlite3.Error, ValueError) as exc:
         return _skipped_entry(
             path,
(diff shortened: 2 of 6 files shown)


Sinity:

Read the full diff. Mirrors merged #3224 exactly: profile_key threaded through the shared hermes_identity helpers; `verification:` family prefix stays distinguishable from `observer:atif|atof`; `hermes_verification_session_id_for` now preserves (split-and-rethreads) the conversational id's qualifier instead of stripping it — the collapse-and-strip pattern the bead names; parent link asserted fail-closed only when profile_root is known. Extending profile_root to all four real call sites (dispatch, import_explain, source_parsing, live/batch) rather than dispatch alone is the right call — the bug existed at each; revision_backfill.py correctly left untouched (excluded surface) and now fails closed there. Anti-vacuity proven by reverting the qualification (3 tests fail as expected).

Same semantic-reparse note as #3225's P1 applies: already-materialized verification-ledger observer sessions carry old unqualified ids; covered by the coordinator's post-promote targeted Hermes reprocess bead (posted separately). 103 focused tests, mypy --strict, quick-gate green. Merging.

PR #3229, opened 2026-07-20
feat(daemon): add judgment-automation policy sweep actor

Sinity:
## Summary

Adds the missing automated actor for the MCP `judge` dispatcher's candidate
queue: a policy engine plus a daemon periodic-loop trigger surface that
decides which `CANDIDATE`-status assertions are auto-judgeable and calls the
same storage chokepoint the MCP `judge` dispatcher uses, escalating the
residue to an explicit, queryable `handoff` assertion.

## Problem

Per the polylogue-800m design note (PR #3202), the MCP `judge` dispatcher
lets an operator judge assertion candidates in bulk, but nothing calls it
automatically. Every agent-authored candidate (pathology findings, ontology/
transform candidates, findings, ...) defaults to waiting on an explicit
human decision, which the operator has said does not scale — most
agent-authored annotations will never be seen by a human.

## Solution

- **`polylogue/daemon/judgment_automation.py`** (new): a policy engine
  (`evaluate_candidate`) applies per-kind confidence thresholds
  (`auto_accept_min_confidence`/`auto_reject_max_confidence`) decoded from a
  new `[judgment_automation.policies]` config table
  (`parse_judgment_automation_policy`). A candidate is judged
  (`run_judgment_automation_sweep_once`) via the real
  `judge_assertion_candidates` storage chokepoint — the same one the MCP
  `judge` dispatcher calls — when the policy is unambiguous; every other
  candidate (no policy for its kind, no `confidence` value, or a confidence
  inside the undecided band) is escalated: a deterministic `handoff`-kind
  assertion pointing at the candidate is upserted so the residue is a
  queryable review queue, not silent limbo.
- **`polylogue/daemon/cli.py`**: wires `periodic_judgment_automation_sweep`
  into the existing `periodic_loops` list next to `_periodic_db_optimize`/
  `periodic_embedding_backlog_check`, following the established periodic-
  maintenance-loop pattern rather than a `DaemonConverger` stage — judgment
  candidates are not file/session-scoped, and
  `docs/retro/2026-05-24-1498-cascade.md` flags `convergence_stages.py` as
  due for a refactor before a 4th stage, so this keeps the bead's footprint
  out of that hot file entirely.
- **`polylogue/config.py`**: four new layered config keys —
  `judgment_automation_enabled` (fail-closed bool, off by default, following
  the `mcp_judge_enabled` pattern), `judgment_automation_interval_s`,
  `judgment_automation_batch_limit`, and the raw `judgment_automation_policy`
  table (decoded by the owning module, same shape as
  `health_convergence_debt`).
- **Dual capability gate**: the sweep only judges when *both*
  `judgment_automation_enabled` AND `mcp_judge_enabled` (polylogue-800m's
  independent capability boundary) are true — it exercises the same judge
  write authority as the MCP dispatcher, so its own opt-in must not, by
  itself, grant that authority. Both flags are re-read every tick
  (`reload_behavior="daemon-loop"`), so flipping either off in
  `polylogue.toml` takes effect on the next tick without a daemon restart.
- Docs: `docs/configuration.md` (config key table) and `docs/daemon.md`
  (new "Judgment Automation" section with a worked `polylogue.toml` example).

## Verification

- `devtools test tests/unit/daemon/test_judgment_automation.py tests/unit/core/test_config_inventory.py tests/unit/daemon/test_daemon_cli.py` → `183 passed`
- `mypy --strict` on `polylogue/config.py polylogue/daemon/cli.py polylogue/daemon/judgment_automation.py tests/unit/daemon/test_judgment_automation.py` → `Success: no issues found in 4 source files`
- `devtools verify --quick` (ruff format/check, mypy, render-all --check, topology/layering/closure-matrix/manifests/ci-workflows/doc-commands/docs-coverage/test-infra-currency/test-clock-hygiene/pytest-timeout-overrides/degrade-loudly) → `exit_code: 0`
- Not run: full `devtools verify` (testmon-affected) / `--seed-testmon` — per-repo policy this is the default pre-PR baseline, but it was explicitly deferred for this lane in favor of the focused set above plus `--quick`.

## Residual risk

- No real-world confidence-producing detector is wired to a
  `[judgment_automation.policies]` entry yet in this PR — the policy table
  starts empty by default, so the sweep is a bounded no-op until an operator
  configures at least one kind. This is intentional (config-gated OFF,
  fail-closed on missing policy) but means the automation has no live
  exercise beyond the test fixtures until an operator opts a real detector
  kind in.
- The escalation `handoff` assertion is discoverable only via an explicit
  `kind=handoff` filter today (not in `ASSERTION_CLAIM_KINDS`'s default set);
  wiring a dedicated MCP/CLI "review queue" surface for it is left as
  follow-up rather than expanding this PR's scope.

Ref polylogue-6qjc


Sinity:

Read the diff. Assessment: policy engine + periodic daemon loop (deliberately NOT a DaemonConverger stage, with the 1498-cascade retro cited — candidates aren't file/session-scoped and convergence_stages.py is flagged pre-refactor); dual fail-closed gate (judgment_automation_enabled AND mcp_judge_enabled, re-read every tick, ConfigError on typo rather than truthiness coercion); judgments go through the real judge_assertion_candidates chokepoint (anti-vacuity: test asserts the JUDGMENT-kind row with the automation actor_ref that only the shared chokepoint writes); undecidable residue escalates as queryable handoff assertions instead of silent limbo; config inventory entries + docs complete. Deferred items (no live policy wiring yet, filter-only escalation discovery) are honest and inside the bead's scope boundary. 183 focused tests, mypy --strict, verify --quick, quick-gate all green. Merging.

Issue #1873, opened 2026-06-14
DSL compiler + consumer correctness gap (delete truncation, negation, words inversion, /api regressions)


Sinity:
# Query/action correctness stop-line: delete, DSL, strict JSON, and `/api/sessions`

## Purpose

Stop-line correctness issue for the query/action spine. This exists because a fast merge wave missed review findings that included real data-loss and wrong-result bugs. Do not build more command/query surface on top of this behavior until the fixes are merged and reviewed.

## Confirmed bug classes

- `delete --yes --all` can truncate execution to the first default page while cardinality/preview resolved a larger set.
- Cardinality/preview and execution can use different query interpretations.
- `words:<=N` / `words:=N` can map to the wrong min/max field.
- Negation on fields with no exclude axis can silently apply a positive filter.
- Escaped quotes in phrases can misparse.
- JSON spec compile can accept unknown fields unless strict.
- `/api/sessions?contains=` and `id:`/`session_id` behavior regressed when routed through DSL/compiler paths incorrectly.

## Current live state

PR #1876 appears to address these bugs. Treat required-green as insufficient: review the diff, tests, and bot comments before closing. If #1876 is already merged, this issue body remains the regression contract.

## Required PR evidence

- Same matched set for preview/cardinality and execution.
- Destructive delete path operates on concrete resolved IDs or otherwise cannot reapply a default limit.
- Multi-page result-set regression for `delete --yes --all`.
- `words>=`, `words<=`, and `words=` map to min/max/both correctly through spec → plan → execution → API/MCP if applicable.
- Unsupported negation fails typed for fields without exclude semantics.
- Escaped quotes round-trip.
- JSON spec strict mode rejects unknown fields.
- `/api/sessions` preserves literal `contains` behavior and applies `id`/`session_id` filtering.

## Files to inspect

```text
polylogue/archive/query/expression.py
polylogue/archive/query/spec.py
polylogue/archive/query/plan.py
polylogue/archive/query/fields.py
polylogue/cli/query_verbs.py
polylogue/cli/verb_cardinality.py
polylogue/cli/archive_query.py
polylogue/daemon/http.py
polylogue/storage/... query execution/read paths
polylogue/mcp/ query paths
tests/unit/cli/test_query_expression.py
tests/unit/cli/test_verb_cardinality.py
```

## Required tests

- Unit tests for every bug class above.
- CLI destructive-action regression with > default page size.
[...]

PR #3233, opened 2026-07-20
refactor(daemon): query-spec fast path parity + provider-usage degrade

Sinity:
## Summary

Two clustered daemon fixes on one branch:

1. **polylogue-4p1.1** — the split-archive `/api/sessions` fast path
   (`_do_archive_session_list`) now builds a `SessionQuerySpec` via the shared
   `from_params`/`compile_expression_into` pipeline instead of hand-mirroring
   every structured filter param, closing the "filter silently dropped until
   someone edits both sites" gap the bead describes.
2. **polylogue-d07y** — `/api/provider-usage` now degrades gracefully (HTTP
   200, redacted `route_state: degraded` envelope) instead of a raw 500 when
   `index.db` is missing/unopenable, matching sibling read endpoints.

## Problem

**Item 1**: `_do_archive_session_list` carried an explicit in-code comment
("the split-archive fast path does not construct a SessionQuerySpec
directly... it must mirror those public params here") documenting that it
hand-read every HTTP filter param separately from `SessionQuerySpec.from_params`
(`polylogue/archive/query/spec.py:498`). That duplication meant a filter field
added to the spec builder — or to `ArchiveStore`'s SQL filter-kwarg surface —
could silently go unenforced by this fast path. Concretely, before this
change the fast path already dropped `project_refs`, `typed_only`,
`message_type`, `since_session_id`, and `boolean_predicate` (all accepted by
`ArchiveStore.list_summaries`/`search_summaries`/`count_sessions`/
`count_search_sessions`), and only read the *first* occurrence of a repeated
HTTP param (`?origin=a&origin=b`) for `origin`/`tag`/`repo`/`tool`/`action`/etc.

**Item 2**: `tests/unit/daemon/test_web_reader.py::TestReaderAssertionEndpoint::
test_operational_web_payloads_redact_configured_archive_paths` deletes
`index.db` mid-test and hits `/api/provider-usage`, which propagated a raw
`sqlite3.OperationalError` ("unable to open database file") through
`daemon_safe_handler`'s generic 500 fallback instead of the graceful
degraded-payload path sibling endpoints (`/api/archive-debt`, the
split-archive `/api/sessions` fast path) already use. Reproduces
deterministically on unmodified master.

## Solution

**Item 1** (`polylogue/daemon/http.py`):
- `_do_archive_session_list` builds one `SessionQuerySpec` via
  `_build_query_spec_params(params, self)` → `SessionQuerySpec.from_params(...)`
  → `compile_expression_into(query_str, base)` (the same pattern `_do_list`
  already used for the full-backend path).
- New `_archive_filter_kwargs_from_spec(spec, since_ms, until_ms)` module-level
  helper derives every `ArchiveStore` filter kwarg directly from the merged
  spec — no per-field HTTP-param re-read for anything `SessionQuerySpec`
  already models.
- `_build_query_spec_params` (shared by both fast paths) now collects
  CSV/repeated-value params via `_csv_values` instead of only the first
  occurrence, and accepts the `has_tool_use`/`has_thinking`/`has_paste_evidence`
  public HTTP aliases alongside the `filter_has_*` `SessionQuerySpec` field
  names (matching the alias mapping already used by `api/archive.py` and
  `mcp/query_contracts.py`).
- Removed the now-dead `_archive_origin_filter`/`_archive_tag_filter` helpers
  the old manual-mirror code used (no remaining call sites).
- New test `test_archive_filter_kwargs_cover_every_storage_lowerable_spec_field`
  enumerates the `ArchiveStore.list_summaries`/`search_summaries`/
  `count_sessions`/`count_search_sessions` filter-kwarg signatures (proves all
  four are identical) and asserts `_archive_filter_kwargs_from_spec`'s output
  covers exactly that set — the AC-required parity test. Deleting any one line
  from that helper's return dict (e.g. `"project_refs"`) fails this test.
- `test_query_spec_param_builder_uses_canonical_query_fields` updated for the
  new tuple-shaped `origin`/`exclude_origin` builder output; new
  `test_query_spec_param_builder_collects_repeated_csv_fields` proves the
  repeated-param regression fix (fails if `_csv_values` is reverted to
  `handler._get_param`).

**Out of scope for item 1**: `exclude_text_terms`/`similar_text`/
`similar_session_id` are not SQL-lowerable by `ArchiveStore`'s filter layer
(they are CLI-side post-filter / vector-search concerns) and remain
unsupported by this fast path, matching prior behavior — this is an
intentional non-goal, not a dropped requirement.

**Item 2** (`polylogue/daemon/http.py`):
- `_handle_provider_usage` wraps the archive read in
  `try/except (DatabaseError, sqlite3.Error)`, returning HTTP 200 with a typed
  `route_state: degraded` envelope (`RouteReadinessPayload` shape, same as
  other read routes) instead of letting the exception reach
  `daemon_safe_handler`'s generic 500. The degraded payload still passes
  through `_web_privacy_safe_projection` so `archive_root` redacts to
  `"[archive]"` like every other archive-path-bearing response. A
  `logger.warning` call satisfies the `verify-degrade-loudly` gate (broad
  except-handlers need a log call or re-raise).
- No new test: the existing test names the exact regression and goes green.

## AC matrix

**polylogue-4p1.1**
| AC | Status |
| --- | --- |
| Fast path derives all structured filters from a `SessionQuerySpec` built via `from_params` (no per-field re-read for spec-modeled filters) | Satisfied |
| Test enumerates `SessionQuerySpec`/storage filter attributes and fails if the fast path drops any | Satisfied — `test_archive_filter_kwargs_cover_every_storage_lowerable_spec_field` |
| In-code "must mirror those public params here" comment and manual mirroring block removed | Satisfied |
| Render surfaces (openapi/cli-output-schemas) still verify | Satisfied — `devtools verify --quick` includes `render all --check` |

**polylogue-d07y**
| AC | Status |
| --- | --- |
| `test_operational_web_payloads_redact_configured_archive_paths` passes on master | Satisfied |
| `/api/provider-usage` with missing/unopenable `index.db` returns the degraded envelope with redacted paths, not a 500 | Satisfied |

## Verification

- `devtools test tests/unit/daemon/test_web_reader.py` → **177 passed**
  (includes the previously-failing
  `test_operational_web_payloads_redact_configured_archive_paths` and both new
  parity tests)
- `devtools test tests/unit/daemon/` → **1931 passed, 1 pre-existing failure**
  (`test_no_token_in_log_or_print_calls`; verified it reproduces identically
  on unmodified master at a different line number — an unrelated static
  line-number scan, not caused by this change)
- `devtools test tests/unit/storage/test_archive_search_contracts.py
  tests/unit/surfaces/test_search_envelope_contract.py
  tests/unit/cli/test_query_exec_laws.py tests/unit/test_read_surface_coherence.py`
  → **152 passed**
- `python3 -m mypy --strict polylogue/daemon/http.py
  tests/unit/daemon/test_web_reader.py` → `Success: no issues found in 2 source files`
- `devtools verify --quick` → `exit_code: 0` (ruff format/check, mypy, render
  all, topology, layering, closure-matrix, schema roundtrip, manifests,
  ci-workflows, doc-commands, docs-coverage, test-infra-currency,
  test-clock-hygiene, pytest-timeout-overrides, degrade-loudly all green)

Ref polylogue-4p1.1
Ref polylogue-d07y

### DIFF
--- a/polylogue/daemon/http.py
+++ b/polylogue/daemon/http.py
@@ -4005,18 +4005,48 @@ class DaemonAPIHandler(BaseHTTPRequestHandler):
         much worse place to block on that than a CLI call a human can
         interrupt (polylogue-dlmv). Callers that want the full
         diagnostics can still request ``?detail=full`` explicitly.
+
+        A missing/unopenable ``index.db`` used to propagate a raw
+        ``sqlite3.OperationalError`` as an unhandled 500 (polylogue-d07y).
+        Sibling read endpoints (``/api/archive-debt``, the split-archive
+        ``/api/sessions`` fast path) degrade gracefully instead — this route
+        now returns the same typed ``route_state: degraded`` envelope shape,
+        privacy-projected like every other archive-path-bearing payload here.
         """
         origin = self._get_param(params, "origin")
         limit = self._get_int(params, "limit", 25)
         detail = self._get_param(params, "detail", "headline") or "headline"
         from polylogue.paths import archive_root as configured_archive_root
 
+        archive_root = configured_archive_root()
+
         async def _get(poly: Polylogue) -> object:
             report = await poly.provider_usage_report(origin=origin, limit=limit, detail=detail)
             return report.to_dict()
 
-        result = _web_privacy_safe_projection(self._sync_run(_get), configured_archive_root())
-        self._send_json(HTTPStatus.OK, result)
+        try:
+            result = self._sync_run(_get)
+        except (DatabaseError, sqlite3.Error) as exc:
+            logger.warning("provider-usage archive read degraded: %s", exc)
+            reason = f"Provider usage accounting is unavailable: archive index could not be read ({exc})."
+            degraded: dict[str, object] = {
+                "archive_root": str(archive_root),
+                "origin": origin,
+                "detail_level": detail,
+                "origins": [],
+                "caveats": [reason],
+                "route_state": _route_readiness_payload(
+                    "degraded",
+                    "/api/provider-usage",
+                    reason=reason,
+                    component="index_db",
+                    stale_available=False,
+                ),
+            }
+            self._send_json(HTTPStatus.OK, _web_privacy_safe_projection(degraded, archive_root))
+            return
+
+        self._send_json(HTTPStatus.OK, _web_privacy_safe_projection(result, archive_root))
 
     @daemon_safe_handler
     def _handle_query_units(self, params: dict[str, list[str]]) -> None:
--- a/polylogue/daemon/http.py
+++ b/polylogue/daemon/http.py
@@ -477,16 +477,6 @@ def _web_reader_archive_root() -> Path | None:
     return root
 
 
-def _archive_origin_filter(params: dict[str, list[str]]) -> tuple[str | None, tuple[str, ...]]:
-    origin_values = params.get("origin") or []
-    if origin_values and origin_values[0]:
-        origins = tuple(
-            dict.fromkeys(token.strip() for value in origin_values for token in value.split(",") if token.strip())
-        )
-        return (origins[0] if len(origins) == 1 else None), origins
-    return None, ()
-
-
 def _utc_timestamp_json() -> str:
     return datetime.now(UTC).isoformat().replace("+00:00", "Z")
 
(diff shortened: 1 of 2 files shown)


Sinity:
Coordinator stand-in review (CodeRabbit rate-limited): read the full 518-line diff.

- **4p1.1**: single merged `SessionQuerySpec` path (`_build_query_spec_params` → `from_params` → `compile_expression_into`) replaces the hand-mirrored per-field re-reads; verified `contains` is carried by the shared builder's first key loop (no #1873 regression). The signature-introspection parity test pins all four ArchiveStore SQL entry points to one filter surface and fails on any dropped kwarg — this is the right mechanical guard for the bead's 'silently absent' failure mode. CSV/repeated-param collection fix (`_csv_values` over `_get_param`) is a genuine bug fix with its own anti-vacuity test.
- **d07y**: degrade path matches sibling routes' `route_state: degraded` envelope, privacy-projected, with the `logger.warning` the degrade-loudly gate requires.
- Verification receipts in PR body check out (177 daemon web-reader tests, 1931 daemon suite with the one known pre-existing static-scan failure confirmed on master via stash, mypy --strict clean).

Merging on green quick-gate.

PR #3234, opened 2026-07-20
fix(sources): fix cross-material isolation guard + dedup census parse by blob_hash

Sinity:
## Summary
Two fixes in the revision-authority census area, sharing the raw-materialization footprint:
- **polylogue-52l2**: stop an isolated later-discovered raw from silently becoming the permanently accepted content for a logical identity that already has known ambiguous/quarantined siblings.
- **polylogue-869u**: dedup census parse by `blob_hash` across `source_path` for providers whose parse identity is byte-content-only, avoiding redundant full parses of byte-identical duplicate blobs.

### polylogue-52l2
`classify_raw_revision_cohort` classifies a byte-prefix chain from whichever `raw_sessions` rows currently carry `revision_kind='full'` for a `logical_source_key` — not against the complete sibling population for that identity. Retiring an ambiguous sibling to membership governance (`replace_raw_membership_census(..., retire_full_revision_governance=True)`, what the backfill/live-watcher callers do once a cohort is decided ambiguous) nulls that raw's `raw_sessions.logical_source_key`, so it disappears from the query. A THIRD raw for the same identity, discovered afterward (e.g. a re-acquired browser-capture snapshot), is then evaluated completely alone: `classify_historical_full_revision_streams` unconditionally accepts a one-member stream as a byte-proven baseline (no sibling to prove a byte prefix against), so the isolated raw permanently becomes the accepted session content — an outcome that depends on incremental discovery order, not on which content is actually correct.

Reproduced directly against `ArchiveStore` (`tests/unit/storage/test_revision_replay.py`), mirroring the live incremental watcher's own call sequence (`sources/live/batch.py`: `bind_raw_revision` then `classify_raw_revision_cohort`, no census-phase re-derivation in between). The equivalent two-call scenario through `backfill_historical_revision_evidence` does NOT reproduce: its census phase unconditionally re-parses every still-unindexed raw and its connected-component selection expansion reunites retired siblings by shared membership evidence before classification runs — protections the live incremental path lacks. This class of bug is closely related to (but distinct from) the head-collision fixes in #3204/#3205/#3211 that landed in the last 48h; those addressed head-handoff precedence and cohort absorption for an *already-established* head, not this isolated-singleton acceptance path.

### polylogue-869u
Live evidence (2026-07-19, `source.db`): the newest-revision-per-`logical_source_key` population is 87,177 rows / 52.1 GiB, but only 85,066 DISTINCT `blob_hash` values / 43.4 GiB — the same bytes (e.g. one 442MB codex computer-use rollout) recur under up to 8 different `logical_source_key`s / source paths, from re-acquisition or re-export. The existing dedup (#3151, `_parse_retained_raws`) only collapses rows sharing BOTH `blob_hash` AND `source_path`, so this whole cross-path duplicate population — 8.7GB / 17% of newest-only bytes — still paid a full parse per row.

### polylogue-52l2 (`polylogue/storage/sqlite/archive_tiers/archive.py`, `polylogue/sources/revision_backfill.py`, `polylogue/sources/live/batch.py`, `polylogue/archive/revision_authority.py`)
- New `ArchiveStore.raw_membership_retired_full_revision_siblings()`: finds `raw_session_memberships` rows for a `logical_source_key` whose `raw_membership_census.detail` matches the (now shared) `HISTORICAL_NON_PREFIX_GOVERNANCE_DETAIL` marker written at retirement — survives the `raw_sessions.logical_source_key` NULL-out.
- `classify_raw_revision_cohort` refuses the byte-chain path entirely whenever this identity has retired sibling evidence, so the caller's existing "no accepted chain" fallback (`convertible_full_revision_raw_ids`) folds the newly-discovered raw into membership governance instead, where the real prefix-based classifier weighs every known sibling together, rather than silently establishing a wrong permanent head.
- Unified the two `retire_full_revision_governance` call sites' previously-divergent detail strings onto one shared constant so the guard recognizes retirement from either path.
- **Deferred** (named, not hidden): once an identity's retired siblings are recognized, nothing in the LIVE incremental watcher path (unlike offline backfill) currently re-unites them with the new raw for a real membership decision — it fails closed instead (logs a warning, surfaces as failed), correctly never establishing wrong content, but requiring a later offline backfill/repair pass to resolve the identity fully. Also out of scope: PR #3204 already tracks a known, separate limitation (a content-ahead capture yielding to a byte-kind chain head) as polylogue-nfl5.

### polylogue-869u (`polylogue/sources/revision_backfill.py`)
`_parse_retained_raws`' grouping key now drops `source_path` for a new `_PATH_INDEPENDENT_PARSE_PROVIDERS` allowlist (ChatGPT, Claude web/Code, Codex, Gemini/Gemini CLI, Grok, Drive) — those parsers derive session identity purely from payload bytes. Providers whose parse DOES depend on `source_path` keep the original `(blob_hash, source_path)` key: Beads derives workspace-scoped native ids from `source_path`, Antigravity's brain-metadata mode and Hermes' ATOF/ATIF/verification-evidence modes derive `profile_root` from `source_path`. `Provider.UNKNOWN` stays path-scoped out of caution.

## Verification

- `devtools test tests/unit/storage/test_revision_replay.py tests/unit/sources/test_revision_backfill.py tests/unit/storage/test_raw_retention.py tests/property/test_sql_injection_boundary.py` → 160 passed.
  - New regression `test_isolated_later_raw_does_not_override_known_ambiguous_cohort` fails on unmodified master (asserts `accepted_raw_ids == ()` but gets the isolated raw's id) and passes after the 52l2 fix.
- `mypy --strict` clean on all four touched modules.
- `devtools verify --quick` → exit 0.
- **Dedup receipt (869u)**: synthetic corpus of 40 distinct ~300KB Codex payloads, each duplicated across 5 different source paths (200 raw rows, mirroring the live re-acquisition-stampede shape), measured via real `_parse_retained_raw` invocation counts:
  - before (`blob_hash`+`source_path` key): 200 parse calls, 0.516s wall clock
  - after (`blob_hash` key, safe providers): 40 parse calls, 0.104s wall clock
  - 160 avoided parse calls (80% reduction on this corpus).

Note: the task brief also asked to run `tests/unit/storage/test_no_string_interpolated_sql.py` explicitly after any `archive.py` edit. That file does not exist in this repo (grepped for "interpolated" and "hardcoded.*line" repo-wide, no match) — ran `tests/property/test_sql_injection_boundary.py` as the closest existing analog instead (40 passed).

## AC matrix

**polylogue-52l2**
- Isolated-singleton acceptance over a known-ambiguous cohort: **satisfied** — guarded in `classify_raw_revision_cohort`, proven by a failing-then-passing regression test.
- Cross-tick reunification of retired siblings with a new raw in the LIVE incremental path: **deferred** — currently fails closed (no wrong content, but no automatic resolution either); needs a follow-up to fold retired-but-known siblings back into a real membership decision at the live-watcher call site, analogous to what offline backfill already does via `convertible_full_revision_raw_ids`.
- Direct-seed order-reversal for `source_outage_interval_events`/`capture_gap_events` (bead's "related, separately-confirmed finding"): **not addressed** — out of this PR's footprint (`_write_parsed_precedence_result` in the excluded `archive_tiers/write.py`); still needs its own investigation pass as the bead notes.

**polylogue-869u**
- Byte-identical blobs parsed at most once per parser fingerprint, per identity-relevant path scope: **satisfied** for the `_PATH_INDEPENDENT_PARSE_PROVIDERS` allowlist; path-dependent providers (Beads/Antigravity/Hermes/Unknown) intentionally keep path-scoped dedup.
- Measured corpus receipt with parse-call counts and wall-clock before/after: **satisfied** — see Verification above.
- No identity regression for source_path-dependent parsers, test covers the Beads workspace case: **satisfied** — `test_parse_retained_raws_preserves_path_scoped_dedup_for_path_dependent_providers`.

Ref polylogue-52l2, polylogue-869u

## Summary by CodeRabbit

* **New Features**
  * Improved handling of historical revision governance during archive classification and replay.
  * Enhanced retained-data processing to safely deduplicate identical content across paths where supported.
* **Bug Fixes**
  * Prevented later isolated revisions from overriding previously identified ambiguous revision groups.
  * Preserved path-specific handling for providers whose results depend on source paths.
* **Tests**
  * Added coverage for cross-path deduplication and ambiguous revision classification scenarios.

### DIFF
--- a/polylogue/archive/revision_authority.py
+++ b/polylogue/archive/revision_authority.py
@@ -23,6 +23,16 @@ class RawRevisionAuthority(StrEnum):
 
 BYTE_AUTHORITY_CENSUS_DETAIL = "append fragments are governed by byte revision authority"
 
+#: ``raw_membership_census.detail`` marker written when a full-only,
+#: non-prefix-chain cohort is retired from byte-revision governance to
+#: membership governance (``ArchiveStore.replace_raw_membership_census``,
+#: ``retire_full_revision_governance=True``). Shared between the writer and
+#: ``ArchiveStore.classify_raw_revision_cohort``'s polylogue-52l2 guard,
+#: which uses it to detect that a logical identity already has retired,
+#: previously-ambiguous sibling evidence before ever accepting a
+#: later-discovered raw as an unconditional singleton byte-proven baseline.
+HISTORICAL_NON_PREFIX_GOVERNANCE_DETAIL = "historical non-prefix full revision governance"
+
 
 @dataclass(frozen=True)
 class RawRevisionEnvelope:
--- a/polylogue/sources/live/batch.py
+++ b/polylogue/sources/live/batch.py
@@ -24,6 +24,7 @@ from polylogue.archive.ingest_flags import (
     NATIVE_BROWSER_CAPTURE_INGEST_FLAG,
 )
 from polylogue.archive.revision_authority import (
+    HISTORICAL_NON_PREFIX_GOVERNANCE_DETAIL,
     RawRevisionAuthority,
     RawRevisionEnvelope,
     RawRevisionKind,
@@ -1983,7 +1984,7 @@ class LiveBatchProcessor:
                     retained_sessions,
                     parser_fingerprint=self._current_parser_fingerprint(),
                     censused_at_ms=acquired_at_ms,
-                    detail="cross-route full revision governance",
+                    detail=HISTORICAL_NON_PREFIX_GOVERNANCE_DETAIL,
                     retire_full_revision_governance=True,
                 )
             member_sessions: dict[str, Any] = {}
--- a/polylogue/sources/revision_backfill.py
+++ b/polylogue/sources/revision_backfill.py
@@ -1045,6 +1045,33 @@ def _pool_dispatch_amortizes(pool_raw_ids: list[str], payload_sizes: dict[str, i
     return total >= _parse_pool_min_aggregate_bytes()
 
 
+#: Providers whose parsed session identity is derived purely from payload
+#: bytes, never from ``source_path`` -- safe to dedup census parse ACROSS
+#: source paths sharing a ``blob_hash`` (polylogue-869u). Excluded
+#: deliberately: ``Provider.BEADS`` derives workspace-scoped native ids from
+#: ``source_path`` (``sources/parsers/beads.py:_repository_root``);
+#: ``Provider.ANTIGRAVITY``'s brain-metadata mode derives its
+#: ``profile_root``/artifact path from ``source_path``
+#: (``sources/dispatch.py``'s ``antigravity.parse_brain_metadata`` call);
+#: ``Provider.HERMES``'s ATOF/ATIF/verification-evidence modes likewise
+#: derive ``profile_root`` from ``source_path``. Those three keep the
+#: conservative same-path-only dedup below. ``Provider.UNKNOWN`` (browser
+#: capture / unclassified) is also excluded out of caution -- its identity
+#: derivation is not centrally audited here.
+_PATH_INDEPENDENT_PARSE_PROVIDERS: Final[frozenset[Provider]] = frozenset(
+    {
+        Provider.CHATGPT,
+        Provider.CLAUDE_AI,
+        Provider.CLAUDE_CODE,
+        Provider.CODEX,
+        Provider.GEMINI,
+        Provider.GEMINI_CLI,
+        Provider.GROK,
+        Provider.DRIVE,
+    }
+)
+
+
 def _parse_retained_raws(
     archive: ArchiveStore,
     raw_ids: list[str],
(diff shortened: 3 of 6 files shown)


Sinity:
Coordinator stand-in review (CodeRabbit still pending): read the full 415-line diff.

- **52l2**: guard is correctly fail-closed and minimal — refusing the byte-chain path when retired ambiguous siblings exist routes the cohort into membership governance where all siblings are weighed. The repro test mirrors the live watcher's exact call sequence and fails on unmodified master (anti-vacuity confirmed by the lane). Marker extraction to `HISTORICAL_NON_PREFIX_GOVERNANCE_DETAIL` keeps writer and guard in lockstep.
- **Data-compat caveat (accepted, noting for the record)**: durable `raw_membership_census` rows written by the LIVE path before this PR carry the old `"cross-route full revision governance"` detail and won't match the guard query — those identities retain the pre-existing bug, no new wrongness. The v42 rebuild's backfill path already used the exact extracted literal, so rebuild-written rows all match. If a backfill sweep for old live-path rows is warranted it belongs in the 52l2 follow-up bead alongside the deferred cross-tick reunification.
- **869u**: cross-path dedup gated on an explicitly audited provider allowlist with Beads/Antigravity/Hermes/UNKNOWN excluded for documented path-derived identity; measured receipt 200→40 parse calls, 5x wall. Both directions covered by tests.

**Merge is HELD until the v42 walk hits terminal** — this PR changes `classify_raw_revision_cohort` semantics that the walk engine would pick up on any restart mid-rebuild. Will merge immediately after the walk's terminal receipt.

Sinity:
CodeRabbit triage: 1 finding (trivial) — `__all__` export for `HISTORICAL_NON_PREFIX_GOVERNANCE_DETAIL` — applied in 7c5ba6543. No other findings. Walk hit terminal (generation promoted, verification checklist green) — merging.

PR #3235, opened 2026-07-20
feat(storage): add messages_fts_identity ledger for rowid-reuse detection

Sinity:
## Summary

Adds a `messages_fts_identity` shadow ledger that binds each `messages_fts` rowid to its `block_id`, exact reconciliation that joins on rowid AND identity (not rowid alone), bounded ops.db drift-magnitude history, and a Prometheus drift-magnitude gauge. Real-trigger metamorphic property tests (Hypothesis `RuleBasedStateMachine`) plus focused anti-vacuity unit tests prove the check has teeth.

## Problem

`messages_fts` is a **contentless** FTS5 table (`content=''`): its `block_id` UNINDEXED column is write-only and never retrievable by a later `SELECT` (verified empirically). SQLite reuses freed rowids — deleting the highest-rowid block then inserting a new one commonly gets the **same rowid back**, exactly what a full-session-replace does. Count-only reconciliation (`source_rows == indexed_rows`) is blind to this: a stale rowid can silently rebind to a different block while both sides still balance, and the existing `fts_freshness_state` ledger only ever tracked counts (missing/excess/duplicate), never identity.

## Solution

- **`polylogue/storage/fts/sql.py`**: `FTS_MESSAGES_IDENTITY_TABLE_SQL` (`messages_fts_identity(rowid PK, block_id UNIQUE, source_hash, recipe_id)`), `FTS_MESSAGES_IDENTITY_RECIPE_ID` (versioned recipe token), and `message_identity_mismatch_sql()` — an exact join across `blocks`/`messages_fts_docsize`/`messages_fts_identity` that flags a **present-but-wrong** ledger entry (rowid-reuse, changed-text, changed-recipe). A missing entry is deliberately NOT counted (see below).
- Trigger bodies (`messages_fts_ai`/`ad`/`au`) gain a paired identity write in the SAME body as their `messages_fts` write, so the two can never diverge for one event. `source_hash` reuses the existing `blocks.content_hash` evidence hash; `recipe_id` is the versioned constant — the same subject/source/recipe split `storage/derivation_identity.py` formalizes for polylogue-wmsc's `DerivationKey`, applied here as a lightweight per-row ledger (never a shared cross-domain table).
- `storage/fts/fts_lifecycle.py`: every bulk rebuild/repair path (full rebuild, batched missing/excess repair, session-scoped repair) pairs its `messages_fts` write with the matching identity companion SQL; `FtsSurfaceInvariant`/`FtsInvariantSnapshot` gain `identity_mismatch_rows` and fold it into `.ready`.
- `storage/fts/freshness.py`: `fts_freshness_state` gains an `identity_mismatch_rows` column (backward-compatible `ALTER ... DEFAULT 0`), threaded through `freshness_ready_record_trusted` and every read/record path.
- `storage/sqlite/archive_tiers/ops.py` + `ops_write.py`: bounded `fts_drift_samples` table (30d / 5k-row retention, same shape as `cursor_lag_samples`), `record_fts_drift_sample`/`list_fts_drift_samples`. `storage/fts/drift_sampling.py::sample_fts_drift_to_ops_sync` reads the just-recorded `fts_freshness_state` row(s) and appends a sample to the sibling `ops.db`; wired into daemon startup and every `record_fts_invariant_snapshot_sync` call site.
- `daemon/metrics.py`: `polylogue_fts_drift_rows` gauge — labels `surface`, `kind` (missing/excess/duplicate/identity_mismatch) — read O(1) from `fts_freshness_state`, no scan on scrape.
- `INDEX_SCHEMA_VERSION` 42→43 with a declared clone-safe fast-forward (`IndexDeltaDeclaration` v43 in `storage/sqlite/lifecycle.py`, `FTS_REINDEX` class) since every ledgered field derives from already-persisted `blocks` columns — no raw reparse needed.

### Design decision: coverage gap vs. conflict

`message_identity_mismatch_sql()` counts only a **present-but-wrong** ledger entry, never a missing one. Rationale, discovered empirically (see Verification): `storage/sqlite/archive_tiers/write.py`'s non-bulk full-session-replace fast path calls `delete_session_rows_sql`/`insert_session_rows_sql` directly (bypassing the suspended block triggers) and does **not** populate the identity ledger inline — this is the dominant real-world re-ingest path. Counting a missing entry as a mismatch made `messages_ready` permanently false on any ordinarily-written archive, which defeats the point of a readiness signal. A missing entry is provably safe on its own: nothing reads block identity *from* the ledger except this reconciliation query, and the next trigger-fired mutation at that rowid always writes a correct fresh row regardless of prior absence. Only a present-but-wrong entry is the actual danger signature.

### STOP-and-report: `write.py` gap (not fixed here, per task boundary)

`storage/sqlite/archive_tiers/write.py` is on this lane's do-not-touch list. Its non-bulk full-session-replace path (`_full_replace_...`, four call sites) calls `delete_session_rows_sql(1)`/`insert_session_rows_sql(1)` directly and would need one additional paired `delete_session_identity_rows_sql`/`insert_session_identity_rows_sql` call at each site to keep the ledger inline-fresh instead of relying on the next repair pass. Both companion functions already exist in `storage/fts/sql.py`. Until that lands, a freshly-written session is coverage-incomplete (not conflicting) until `dangling_repair.py`'s missing-row repair backfills it — the same eventually-consistent contract `missing_rows`/`excess_rows` already had.

## Verification

- `mypy --strict` on all touched modules: `Success: no issues found in 12 source files`
- `devtools verify --quick`: exit 0 (format, lint, mypy, render-all-check, layering, degrade-loudly, closure-matrix, schema-roundtrip, manifests, ci-workflows, doc-commands, docs-coverage, test-infra-currency, test-clock-hygiene, pytest-timeout-overrides)
- `devtools test tests/unit/storage/test_fts_identity_ledger.py tests/property/test_fts_identity_state_machine.py tests/unit/daemon/test_metrics_endpoint.py tests/unit/daemon/test_daemon_cli.py tests/unit/daemon/test_fts_readiness_fallback.py`: **157 passed**
- Broader FTS-adjacent sweep (`test_dangling_fts_derived_surfaces`, `test_fts_global_repair`, `test_convergence_stages`, `test_bulk_build_lifecycle`, `test_fts_bloat_invariants`, `test_index`, `test_fts_repair_sql`, `test_write_gateway`, `test_write_effects`, `test_ingest_batch_fts_repair`, `test_pl_fold`, `test_schema_safety`): **all green except one pre-existing failure** (`test_reset_and_repair_fts_rows_recovers_excess_message_rows`, confirmed failing identically on baseline `master` before this branch — `no such table: main.derived_refresh_guard`, unrelated to this change)
- `devtools lab policy schema-versioning`: v43 correctly declared; the one reported gap (`missing: [40]`) is confirmed pre-existing on baseline
- `devtools render topology-projection && devtools render topology-status`: regenerated for the new `storage/fts/drift_sampling.py` module

Not run: full `devtools verify` (testmon-seeded pytest) — testmon wasn't seeded in this fresh worktree and seeding is a full-suite-equivalent cost; the targeted sweep above plus mypy --strict cover every touched surface.

## AC matrix (polylogue-1xc.12)

| # | AC | Status |
| --- | --- | --- |
| 1 | `messages_fts_identity(rowid PK, block_id UNIQUE, source_hash, recipe_id)` + exact freshness fields | **Satisfied** |
| 2 | Production triggers + full rebuild maintain FTS, docsize, ledger, O(1) state atomically across empty/text/delete/replace/rollback/recipe-change | **Satisfied** for triggers and bulk rebuild/repair paths in `fts_lifecycle.py`; **deferred** for `write.py`'s non-bulk fast path (see STOP-and-report above) |
| 3 | Exact reconciliation compares identity/source/recipe; rowid-reuse, changed-text, changed-recipe all fail before repair | **Satisfied** (proven by `test_fts_identity_ledger.py::TestIdentityMismatchDetection` and the metamorphic state machine) |
| 4 | Periodic convergence detects drift in FTS or ledger, rebuilds both, records before/after, converges idempotently | **Partially satisfied**: exact reconciliation + bounded repair self-heal on every rebuild/repair call and daemon startup; a dedicated periodic *scheduled* convergence stage (vs. today's startup/rebuild-triggered cadence) is not added — no existing `ConvergenceStage` was extended, out of scope for this lane's effort budget |
| 5 | Metrics/readiness never scan blocks/FTS; ops history is bounded | **Satisfied** (`polylogue_fts_drift_rows` reads O(1) from the ledger; `fts_drift_samples` is time+row-bounded) |
| 6 | Real-trigger Hypothesis state machine covers rowid reuse, full replace, rollback, empty text, source change, recipe change | **Satisfied** — `tests/property/test_fts_identity_state_machine.py` (recipe change proven via the corrupt-then-repair rule since the recipe constant is compiled into trigger SQL, not runtime-configurable) |
| 7 | Removing any trigger arm, block_id/source/recipe check, or exact audit fails | **Satisfied by construction**: tests exercise the real trigger DDL and real reconciliation SQL, never a mock; anti-vacuity docstrings on each test name the exact mutation that would flip it red |
| 8 | FTS consumes shared DerivationKey value semantics without sharing embedding storage/scheduling/lifecycle | **Satisfied** — `source_hash`/`recipe_id` mirror the subject/source/recipe split conceptually; no table/module shared with polylogue-wmsc's embedding freshness work |

Ref polylogue-1xc.12 (do not close — AC #4 and the `write.py` gap in AC #2 remain open follow-up work)


Sinity:
Coordinator stand-in review (CodeRabbit rate-limited): read the full 2561-line diff.

- **Core design is right**: contentless FTS5 makes `block_id` write-only, and rowid reuse under full-session-replace is exactly the drift class count-only reconciliation cannot see. The rowid-keyed identity ledger with source_hash (reusing `blocks.content_hash`) + versioned recipe_id gives exact reconciliation real teeth, and pairing the identity writes inside the SAME trigger bodies eliminates any observable divergence window.
- **The present-but-wrong-only scoping of `message_identity_mismatch_sql` is the correct call** — the design finding writeup (missing-entry counting would poison `ready` permanently via write.py's fast path) is exactly the kind of readiness-signal honesty this repo requires, and it's documented in three places.
- **Schema handling**: v43 with a declared clone-safe `FTS_REINDEX` fast-forward (every ledgered field derivable from persisted blocks columns, no reparse) matches the v35 tokenizer-reindex precedent. Correct classification per the schema regimes doctrine.
- **Bounded ops history**: retention + row-cap shape copied from route_observations; sampling is swallow-on-failure so index.db repair paths never gain an ops.db write dependency. Good.
- **Residuals acknowledged**: write.py identity companions (4 call sites, hot-file restricted) and a scheduled periodic convergence stage (AC4) — will get a named follow-up bead at close time, not silent deferral.

**Merge is HELD until the v42 walk hits terminal** (schema-version bump + trigger DDL changes the rebuild engine would pick up on a mid-generation restart). Merges right after #3234 once the terminal receipt lands.

PR #3238, opened 2026-07-21
feat(storage): execute declared index fast-forward plans on open

Sinity:
## Summary

Wires an executor into the index-tier open path so a declared clone-safe schema-version gap no longer forces a full archive rebuild. Before this change, `IndexDeltaDeclaration`/`IndexFastForwardPlan` (`polylogue/storage/sqlite/lifecycle.py`) had no consumer: `initialize_archive_database` always raised the rebuild-required `RuntimeError` for any `PRAGMA user_version` gap, benign or not.

## Problem

Live incident 2026-07-21: a freshly promoted v42 `index.db` could not be opened by v43 code, even though `#3235` shipped a declared clone-safe fast-forward (`IndexDeltaDeclaration` v43, `FastForwardOperation` name=`v43-messages-fts-identity`, kind=`REBUILD_FTS`). `grep` confirmed nothing outside `lifecycle.py` consumed `IndexFastForwardPlan`/`eligible_for_sql_fast_forward` — the declaration registry existed but no open-path executor applied it, so the declared-benign v43 delta still forced the ~11h full rebuild it exists to avoid.

## Solution

- `polylogue/storage/sqlite/archive_tiers/index_fast_forward_executor.py` (new): `apply_index_fast_forward(conn, plan)` applies a plan's declarations one at a time, each inside its own `BEGIN IMMEDIATE`/commit transaction, bumping `PRAGMA user_version` in the same transaction. A crash mid-chain rolls back to the previous version; the next open rebuilds the plan from whatever version actually landed and re-applies idempotently.
- Dispatch is generic over `FastForwardOperationKind`, not per-version:
  - `REBUILD_FTS` — drops the declared triggers, calls `ensure_fts_index_sync`, then literal `DELETE FROM messages_fts` / `DELETE FROM messages_fts_identity` + `insert_all_message_*_rows_sql()` for whichever surfaces the declaration names (narrow identity-only refresh for v43, not a full content rescan).
  - `REPLACE_TABLE` — copy-forwards columns shared between the existing and canonical shape onto a temp table, then renames it in (handles the declared v33/v36/v38/v41 constraint-widening/column-removal deltas without a raw reparse).
  - `REPLACE_VIEW` / `CREATE_INDEX` / `DROP_TABLE` — straightforward canonical-DDL re-run / drop.
[...]

PR #3239, opened 2026-07-21
fix(storage): write.py identity companions + periodic FTS drift stage

Sinity:
## Summary

Implements the two named residuals from polylogue-1xc.12 (PR #3235), plus fixes a live pre-existing `messages_fts_identity` UNIQUE(block_id) regression reported mid-flight by another lane:

1. Pairs `storage/sqlite/archive_tiers/write.py`'s non-bulk full-session-replace fast path's `delete_session_rows_sql`/`insert_session_rows_sql` calls with the matching `delete_session_identity_rows_sql`/`insert_session_identity_rows_sql` companions (3 delete + 2 insert sites, plus the sync/async twin purge helpers in `storage/session_replacement.py`).
2. Adds `polylogue/daemon/fts_identity_convergence.py`: a periodic 24h-cadence recompute pass that is the recompute authority for `identity_mismatch_rows`, correcting the documented `daemon/fts_startup.py` bounded-STALE-write hazard.
3. Fixes a `messages_fts_identity` block_id UNIQUE-constraint self-heal gap: every identity-ledger write now uses `INSERT OR REPLACE`/multi-target UPSERT so a stale orphaned ledger row can never abort a later colliding write.

## Problem

**AC2 residual (write.py):** `message_identity_mismatch_sql`'s docstring documented that write.py's fast path did not call the identity companions inline, leaving sessions written through the dominant real-world path (`write_parsed_session_to_archive`, `merge_append=False`) coverage-incomplete until the next repair backfilled them.

**AC4 residual (periodic stage):** identity self-heal ran on every rebuild/repair/startup but no *scheduled* pass recomputed the exact snapshot on a quiet cadence, and `fts_startup.py`'s bounded STALE-write path could reset a recorded nonzero `identity_mismatch_rows` to 0 without recomputing it.

**Live regression (mid-flight, coordinator-assigned):** another lane reported 6 pre-existing failures on unmodified master tied to `messages_fts_identity`'s `block_id TEXT NOT NULL UNIQUE` constraint, in `tests/unit/storage/test_revision_replay.py` and `tests/unit/sources/test_live_batch_support.py`. The exact production write-sequence was **not reproducible**: both files pass cleanly and repeatedly (serial, `-n auto` xdist, multiple runs) against this branch and a disposable fresh `origin/master` worktree — the only failure present in either is a pre-existing, unrelated `test_bundle_replay_respects_unconvertible_single_session_head` file-cleanup assertion, confirmed present on unmodified `origin/master` too. However, a synthetic repro built directly against the real trigger DDL (orphan a ledger row via a trigger-suspended delete, then insert a new block computing the same `block_id` at a different rowid) reliably reproduces `sqlite3.IntegrityError: UNIQUE constraint failed: messages_fts_identity.block_id` on current code, confirming a genuine invariant gap independent of the unreproduced call sequence: the ad/au trigger arms and every bulk delete companion only ever clean up ledger rows by `rowid`, never by `block_id`.

## Solution

- write.py: added paired companion calls at all 3 delete + 2 insert call sites (fast-path scoped delete+reinsert, bulk-fts-guard delete+reinsert, trigger-missing purge), plus the sync/async twin in `session_replacement.py`.
- `fts_identity_convergence.py`: standalone module (not a new `DaemonConverger`/`ConvergenceStage` — `docs/retro/2026-05-24-1498-cascade.md`'s explicit verdict is "refactor before adding a fourth stage"), following the `periodic_judgment_automation_sweep` (PR #3229) shape: a sync run-once function + async scheduling loop wired into `daemon/cli.py`'s `periodic_loops`, fixed 24h cadence (matching `_periodic_db_optimize`'s cost-class reasoning), no new config knob.
- `messages_fts_identity` writes now use `INSERT OR REPLACE` (trigger bodies, `insert_all_message_identity_rows_sql`, `insert_session_identity_rows_sql`) and a second `ON CONFLICT(block_id)` clause on `repair_message_identity_rows_range_sql` (SQLite 3.35+ multi-target UPSERT; project already requires 3.43+ for FTS5 `contentless_delete`), so a pre-existing row blocking on either the `rowid` PK or the `block_id` UNIQUE constraint is evicted before the incoming row lands.

## Verification

- `devtools test tests/unit/storage/test_fts_identity_ledger.py` → 19 passed (new `TestWritePathIdentityCompanions` + `TestIdentityLedgerBlockIdCollisionRepro` classes)
- `devtools test tests/unit/daemon/test_fts_identity_convergence.py` → 4 passed
- `devtools test tests/unit/storage/test_revision_replay.py tests/unit/sources/test_live_batch_support.py tests/unit/storage/test_fts_bloat_invariants.py tests/property/test_fts_identity_state_machine.py tests/unit/storage/test_archive_tiers_write.py tests/unit/storage/test_bulk_fts_prefix_reextract.py tests/unit/storage/test_bulk_build_lifecycle.py tests/unit/storage/test_lineage_normalization.py tests/unit/daemon/test_fts_identity_convergence.py` → 205 passed, 2 failed (pre-existing unrelated bundle-replay-cleanup assertion, confirmed present on unmodified `origin/master`)
- `mypy --strict` on all touched/new modules → Success, no issues
- `devtools render topology-projection && devtools render topology-status` (new module) → committed
- `devtools render all --check` → no "out of sync" surfaces
- `devtools verify --quick` → exit 0
- Anti-vacuity (both fixes): manually reverted the paired companion calls / the `INSERT OR REPLACE` fix and reran the corresponding tests → both failed as expected (missing-entry assertion / `sqlite3.IntegrityError`); restored and reran → both passed

## AC matrix

| AC | Status |
| --- | --- |
| write.py identity companions paired at all call sites | Satisfied |
| Test proving zero missing ledger entries after full-session-replace | Satisfied |
| Periodic drift recompute stage, following #3229's pattern (not a new ConvergenceStage) | Satisfied |
| Stage corrects the fts_startup.py bounded-write hazard | Satisfied |
| Test: corrupt → recompute nonzero → repair → recompute zero | Satisfied |
| Live UNIQUE(block_id) regression (coordinator-assigned) | Satisfied — fixed via INSERT OR REPLACE/multi-target UPSERT; original reported call sequence not reproducible after extensive attempts, but the underlying invariant gap is real and independently confirmed |

Ref polylogue-miwv


Sinity:
Coordinator stand-in review (CodeRabbit rate-limited, 0 inline comments): read the full 1076-line diff.

- **write.py companions**: all delete/insert call sites paired (fast path, bulk-fts guard, trigger-missing purge) plus the sync/async twin purges in session_replacement.py — the twin-tree grep the repo requires actually happened.
- **UNIQUE(block_id) fix is the semantically right one**: the ledger has two independent uniqueness constraints and the ad/au cleanup only ever keyed rowid, so a stale (old_rowid, block_id=X) row aborts a later insert of the same block_id at a new rowid. INSERT OR REPLACE makes the incoming write authoritative (self-heal over abort) and the dual ON CONFLICT clause on the repair UPSERT handles the orphan-holder case. Synthetic repro against the real trigger DDL confirms the gap on pre-fix code.
- **Periodic stage placement is correct per the 1498-cascade retro**: standalone asyncio loop in daemon/cli.py (judgment-automation shape), not a fourth ConvergenceStage; gated on catch_up_complete; runs on the write coordinator so it serializes with ingest; explicitly the recompute authority over the fts_startup bounded STALE-write hazard, with a dedicated reproduction test.
- **Trigger-body compat note (accepted)**: CREATE TRIGGER IF NOT EXISTS won't refresh bodies on archives already at v43 — no such production archive exists (live archive is v42 and will fast-forward under post-#3239 code via #3238's executor), so nothing is stranded.
- Verification receipts check out (all touched suites green, mypy strict clean, render + quick gates exit 0).

Merging.

PR #3241, opened 2026-07-21
feat(declarations): typed per-artifact derivation + synthetic-domain proof

Sinity:
## Summary

Completes the storage-free `polylogue/declarations` authoring kernel
(polylogue-o21.1) against its acceptance criteria. The five-file package
(`models.py`, `registry.py`, `derive.py`, `validation.py`) and its MCP pilot
(`polylogue/mcp/declarations`, t46.8.1) were already on `master` (landed via
#3004 as a side effect of the MCP tool-algebra declaration work); this PR
closes the gaps between that existing implementation and the o21.1 AC:
typed per-artifact derivation, a synthetic-domain proof with per-dimension
incompatibility coverage, and a repo-level layering rule.

## Problem

`polylogue/declarations/derive.py` exposed exactly one generic derivation
shape — `DerivationInput.normalized`, a full `asdict()` blob of the whole
`DeclarationSpec`. The o21.1 design explicitly calls for `derive.py` to
"expose typed deriver protocols and stable artifact inputs for names,
contracts, schemas, docs, examples, discovery text, and completeness
projections" — i.e. narrow, source-provenanced, per-artifact-kind slices a
domain renderer can depend on individually, not one undifferentiated blob.

Separately, the AC requires a synthetic proof "asserting the diagnostic
names the differing dimension" per axis; the existing test only exercised
all five `CompatibilityKey` dimensions differing simultaneously, which does
not prove the diagnostic isolates a *single* varying dimension. The
import-hygiene AST test's forbidden-import list also omitted
`polylogue.sources`/`polylogue.maintenance`, which the AC names explicitly
("no import from MCP, sources, storage, insights, maintenance, or other
domain registry packages"). Finally, `docs/plans/layering.yaml` (the
cross-package import gate enforced by `devtools verify layering`) had no
entry for `polylogue/declarations` at all — the boundary was checked only
by an in-package unit test.

## Solution

- `polylogue/declarations/derive.py`: added
  `NameInput`/`ContractInput`/`SchemaDocInput`/`ExampleInput`/
  `DiscoveryInput`/`CompletenessInput` dataclasses, each carrying
  `declaration_id` + `owner_path` (source provenance) plus exactly the
  fields that artifact kind needs. Added one typed `Protocol` deriver per
  kind (`NameDeriver`, `ContractDeriver`, `SchemaDocDeriver`,
  `ExampleDeriver`, `DiscoveryDeriver`, `CompletenessDeriver`) alongside the
  existing generic `DeclarationDeriver`/`DerivationInput`. Added a
  `normalized_*_bytes()` helper per kind for deterministic drift checks,
  reusing the registry's stable `declarations()` ordering so every
  projection is provably independent of registration order. Re-exported
  everything from `polylogue/declarations/__init__.py`.
- **Design decision (underspecified in the bead):** kept the existing
  generic `DeclarationDeriver`/`DerivationInput` pair rather than deleting
  it — a domain renderer that legitimately needs the full declaration (e.g.
  a debug/audit dump) still has that option, while renderers that only need
  one artifact kind now have a narrower typed contract available. This is
  additive, not a replacement of the existing shape, since the existing
  shape is already consumed transitively by `normalized_derivation_bytes`
  used in the pre-existing registry test.
- **Design decision:** `owner_path` alone stands in for "source provenance"
  on every new projection (matches the existing `DeclarationSpec.owner_path`
  field and the `Diagnostic.owner_path` convention in `validation.py`) —
  no new provenance concept was introduced.
- `tests/unit/declarations/test_synthetic_domain.py` (new): a fake `widget`
  domain in the shape a real pilot (MCP) would use it. Proves (a) two
  legitimately compatible declarations unify into one family, (b) five
  parametrized cases — one per `CompatibilityKey` dimension — where a
  declaration differs on exactly that one axis, asserting
  `DeclarationConflictError.differences` names *only* that dimension and
  that the registry state is unchanged after the rejected registration
  (never coerces), and (c) order-independent derivation across every new
  typed artifact kind plus the existing full-declaration blob, under a
  seeded shuffle.
- `tests/unit/declarations/test_registry.py`: widened the import-hygiene
  AST walk's forbidden tuple to include `polylogue.sources` and
  `polylogue.maintenance` (previously covered only implicitly, since no
  test exercised those specific forbidden imports).
- `docs/plans/layering.yaml`: added a `polylogue/declarations` rule
  disallowing imports from every domain/surface package the AC names
  (`api`, `archive`, `cli`, `config`, `daemon`, `insights`, `maintenance`,
  `mcp`, `operations`, `sources`, `storage`, `ui`), enforced by
  `devtools verify layering` (part of `devtools verify --quick`).
- No new module was added under `polylogue/`, so no topology-projection
  regeneration was required; `devtools render all --check` confirmed no
  drift.

## AC matrix (polylogue-o21.1)

1. **Satisfied.** `models.py`/`registry.py`/`derive.py`/`validation.py`
   expose typed, storage-free APIs; the AST import-hygiene test (widened
   this PR) plus the new `docs/plans/layering.yaml` rule both assert no
   import from MCP/sources/storage/insights/maintenance/other domain
   registries.
2. **Satisfied.** `CompatibilityKey` requires equality of all five
   dimensions before declarations share a family (pre-existing
   `registry.py`); the new synthetic-domain test proves the diagnostic
   names exactly one differing dimension in isolation for each of the five
   axes, and that an incompatible declaration is never coerced in.
3. **Satisfied.** `derive.py` now produces typed inputs for
   names/contracts/schema-doc fragments/examples/discovery/completeness,
   each with source provenance (`owner_path`); the synthetic-domain test
   proves shuffled registration order yields byte-equivalent normalized
   output for every one of those six kinds plus the full-declaration blob.
4. **Satisfied** (pre-existing `validation.py`, unchanged this PR — verified,
   not reimplemented). Missing producer/handler/role-gate/schema/example/
   output/consumer-edge each produce one `Diagnostic` naming declaration id,
   owner path, and exact repair command. "Removing the real pilot
   declaration or handler fails anti-vacuously" is covered on the MCP side
   by `tests/unit/mcp/test_tool_declarations.py` (t46.8.1's own scope, out
   of this PR's file set — re-ran to confirm it still passes with the
   widened kernel).
5. **Satisfied, pre-existing.** `polylogue/mcp/declarations/{models,registry}.py`
   already import and build on `polylogue.declarations` (`DeclarationSpec`,
   `DeclarationRegistry`, `CompatibilityKey`, etc.) while owning all
   MCP-specific verb/role/result/registration semantics itself — verified by
   grep and by `mypy --strict` on both packages together.
6. **Satisfied.** Synthetic families, deterministic derivation, actionable
   diagnostics, and the real MCP pilot all pass; `devtools verify --quick`
   is green.

## Verification

```
nix develop --command mypy --strict polylogue/declarations tests/unit/declarations polylogue/mcp/declarations
# Success: no issues found in 11 source files

nix develop --command devtools test tests/unit/declarations tests/unit/mcp/test_tool_declarations.py
# 18 passed in 5.12s

nix develop --command python -m devtools verify layering
# No layering violations found.

nix develop --command devtools verify --quick
# exit_code 0 (ruff format/check, mypy, render all, verify topology/layering/
# closure-matrix, lab schema roundtrip, verify manifests/ci-workflows/
# doc-commands/docs-coverage/test-infra-currency/test-clock-hygiene/
# pytest-timeout-overrides/degrade-loudly — all ok)
```

Not run: `devtools verify --all` (full non-integration suite) — the
`--quick` gate plus focused `devtools test` on every touched/consuming
module is the documented sufficient bar for this size of change; nothing
in this diff touches schema, migrations, or cross-cutting runtime behavior
that would justify the full run.

## Residual risk

- The kernel's `DeclarationDeriver`/`DerivationInput` generic path and the
  six new typed paths are both live; nothing currently forces a domain
  package to prefer the typed path over the generic blob. o21.2/o21.3
  (scaffolds + cross-family adoption, already tracked, out of this bead's
  scope) are the natural place to steer new domain declarations toward the
  typed projections.
- AC 4's "removing the real pilot declaration or handler fails
  anti-vacuously" clause is proven by t46.8.1's own test suite
  (`tests/unit/mcp/test_tool_declarations.py`), not by a new test added in
  this PR — this bead does not own the MCP pilot's test file, and the
  existing coverage there was unaffected by this change (confirmed by
  running it green above).

Ref polylogue-o21.1

## Summary by CodeRabbit

* **New Features**
  * Added category-specific declaration derivation for names, contracts, schemas, examples, discovery, and completeness.
  * Added deterministic normalized outputs for each declaration artifact type.
  * Expanded the public declarations package exports for these capabilities.

* **Tests**
  * Added coverage for declaration family compatibility, isolated validation differences, registry safety, and order-independent derivation.
  * Strengthened import-boundary checks for the shared declarations layer.

* **Refactor**
  * Formalized layering rules to keep the shared declarations kernel independent of domain, runtime, and storage components.


Sinity:
Coordinator stand-in review (CodeRabbit pending, 0 inline comments): read the full 622-line diff.

- The key finding — kernel pre-existed from #3004's MCP tool-algebra work — was handled correctly: gap-closure against the o21.1 AC instead of a duplicate implementation. This is exactly the surgical-renewal read.
- Typed per-artifact projections (Name/Contract/SchemaDoc/Example/Discovery/Completeness) with owner_path provenance and one Protocol per kind match the design text; deterministic-bytes helpers keep drift checkable.
- The five-dimension diagnostic tests (one per CompatibilityKey axis, asserting the diagnostic names ONLY the differing axis + registry unchanged after rejection) and the repo-level layering rule are the right enforcement layers — the import boundary is now checked in verify --quick, not just an in-package test.
- Additive retention of the generic deriver alongside typed ones is the correct call pre-o21.2/o21.3.

Merging. This unblocks polylogue-2qx.1.1 (OriginSpec kernel, next tranche).

PR #3232, opened 2026-07-20
feat(devtools): consume bd sync receipts in backlog-hygiene gate

Sinity:
## Summary

Adds check S1 to `devtools/verify_backlog_hygiene.py` (the standing `devtools lab policy backlog-hygiene` / `devtools verify --lab` gate): it consumes the `SyncReceipt` that `devtools/bd_reimport_guard.py` (polylogue-gxjh.1, merged PR #3220) writes to `.cache/bd-sync-receipts/` after every bd JSONL synchronization, and fails with an actionable finding when the most recent receipt is corrupt, incomplete, or reports a conflicted/unauthorized-downgrade row.

## Problem

Ref polylogue-8jg9.1

gxjh.1 shipped a monotonic per-row bd JSONL sync engine (merge_rows classifies every row new/updated/equal/skipped_downgrade/conflicted/recovered_downgrade, `SyncReceipt.is_clean` reports overall health) precisely because the 2026-07-15 staged-conflict and nine-bead-downgrade incidents proved that a syntactically valid `.beads/issues.jsonl` and a green command exit status are not proof that synchronization happened correctly — a whole-file replacement can silently restore stale per-row state while both existing hygiene lints stay green. gxjh.1's own bead notes explicitly named 8jg9.1 as "repo policy consumption of bd-sync receipts": the guard writes the receipt, but nothing read it back to gate anything until now.

8jg9.1's full scope (active-set admission bands, execution-focus derivation, bounded-page/streaming complete enumeration, program grouping) is much larger than one PR; this PR is the narrow, explicitly-named slice — receipt consumption — assigned to this lane. 8jg9.1's own notes state "the portfolio/frontier policy may be implemented independently, but cannot certify synchronized state until it consumes a complete monotonic sync receipt and rejects downgrade/conflict/incomplete outcomes," and separately forbid implementing a second merge algorithm here.

## Solution

- New check **S1** in `devtools/verify_backlog_hygiene.py`:
  - `_latest_receipt_path` finds the most recent file under `.cache/bd-sync-receipts/` (chronological by filename token, per `bd_reimport_guard.write_receipt`'s naming scheme).
[...]

PR #3242, opened 2026-07-21
feat(devtools): add active-set/execution-focus/program-grouping checks to backlog hygiene

Sinity:
## Summary

Extends `devtools/verify_backlog_hygiene.py` (`devtools lab policy backlog-hygiene`) with the remaining structural checks from polylogue-8jg9.1's three-view policy — active-set validity, execution-focus (in-progress claim) liveness, and program-grouping consistency — plus a non-blocking active-set size diagnostic. AC4's receipt-consumption slice already shipped in PR #3232; this PR is the rest of the scope (ACs 1, 2, 3, 5, 6, 7).

## Problem

polylogue-8jg9.1 required three separately-checked views over the Beads backlog (full ambition / active set / execution focus), but only structural admission metadata existed (`metadata.frontier == "active"`, `frontier_program_ref`) with no invariant enforcement. Concretely, before this PR:

- An epic could carry `frontier=active` and sit in the "active set" as if it were a leaf of work (epics are not work).
- An active leaf's `frontier_program_ref` could be missing, dangling, or point at a bead that never claimed `frontier_program=active` — silently breaking the program grouping.
- `in_progress` claims had no staleness detection (2026-07-12 bead notes record four abandoned session-quota-killed claims discovered only by manual audit).
- A bead could claim `frontier_program=active` with zero admitted leaves actually backing it (the bead's design requires this to be *derived* from member leaves, not asserted independently — AC3).
- The 2026-07-15 bead notes record a `bd list --status open --limit 500` process reaching 7.88 GB RSS / 20.3 GB swap — any new active-set tooling had to avoid unbounded live `bd` queries (AC2).

## Solution

- **F1** (`devtools/verify_backlog_hygiene.py`): an open bead carrying `metadata.frontier == "active"` that is itself an epic is a hard finding — epics are not work.
- **F2**: an active leaf (non-epic) with no `frontier_program_ref`, a dangling ref, or a ref to a bead that isn't itself `frontier_program=active` is a hard finding.
- **F3**: an `in_progress` bead with no `updated_at` activity inside a configurable window (default 7 days, `--stale-claim-days`) is a hard finding — "recent" is defined purely from bd's own `updated_at` field (bd has no separate activity timestamp).
- **F4**: a bead marked `frontier_program=active` with zero open active leaves referencing it via `frontier_program_ref` is a hard finding (program grouping must be derived from real members).
- **`compute_active_set_summary()`**: reports the current active-leaf count against a configurable soft target (30) / warn (50) band as an **informational diagnostic only** — wired into `--json` (`active_set` key) and text output, but never added to `collect_findings()`'s `Finding` list. Exceeding either threshold can never fail the gate, truncate the set, or hide a bead (AC1's "no fixed leaf or program count is a semantic failure or hidden truncation").
- **Bounded enumeration** (AC2): measured first rather than assumed. Every check in this module — old and new — reads the already-exported, finite `.beads/issues.jsonl` (currently ~1,092 issues / ~5 MB) in a single `read_text().splitlines()` pass; none of it shells out to `bd list`/`bd ready` per-issue or re-queries a live, potentially-unbounded source. This is the same bounded shape the 15 pre-existing checks already used, so no rewrite was needed for this slice — documented inline in `_load()`'s docstring against the 2026-07-15 incident for future maintainers.
- Seeded-violation tests: one dedicated test per new check (missing/dangling/mismatched program ref get their own cases), plus a combined fixture exercising F1–F4 together in one `collect_findings()` call, following the bead's original seeded-violation contract. Also covers `compute_active_set_summary()`'s within-target/above-target/above-warn bands and confirms exceeding target never becomes a `Finding`.

Out of scope for this PR (explicitly, per the task instructions): `devtools/frontier_report.py` (the devloop batch-classification tool, a different command with a different purpose) still calls `bd list --status in_progress --limit 0`; it is not part of the portfolio active-set policy surface this bead's remaining scope names, so it was left untouched. Execution-focus *derivation* (ranking by claims/readiness/priority/critical-path/conflicts into a small subset) is not implemented here — this PR only guards the structural invariants backing it (F1/F2/F4 for the active set it draws from, F3 for claim liveness); deriving the ranked subset itself was not in the explicit remaining-scope list handed to this lane.

## Verification

- `devtools test tests/unit/devtools/test_verify_backlog_hygiene.py` — 31 passed (16 pre-existing + 15 new: 9 dedicated F1–F4 cases, 1 combined seeded fixture, 5 `compute_active_set_summary` cases).
- `mypy --strict devtools/verify_backlog_hygiene.py devtools/command_catalog.py tests/unit/devtools/test_verify_backlog_hygiene.py` — no issues.
- `ruff format --check` / `ruff check` — clean.
- `devtools render all --check` — sync OK across all surfaces (`docs/devtools.md` regenerated for the updated check count/description).
- `devtools verify --quick` — all 16 steps green (format, lint, mypy, render, topology, layering, closure-matrix, schema roundtrip, manifests, ci-workflows, doc-commands, docs-coverage, test-infra-currency, test-clock-hygiene, pytest-timeout-overrides, degrade-loudly).

### Real-backlog findings (`devtools lab policy backlog-hygiene --fresh --json` against the live archive)

The gate was already red before this PR (pre-existing debt: A1=53, R1=18, X2=3, H3/H4=1 each, P1=2, E2=1 — unrelated to this change, growth of a known-open debt category per the bead's 2026-07-17 note). This PR's new checks contribute:

- **F4 = 6 findings** (real, not tuned away): `polylogue-a7xr`, `polylogue-4p1`, `polylogue-jlme`, `polylogue-fs1`, `polylogue-4ts`, `polylogue-303r` — each claims `frontier_program=active` with zero open active leaves currently referencing it (their prior admitted members have since closed). This is genuine "program grouping wasn't re-derived after its leaves closed" debt the bead's design anticipated; worth a follow-up bead to either close these programs or re-admit successor leaves.
- **F1 = 0, F2 = 0, F3 = 0** — no active-epic leaves, no invalid program refs, and no in_progress claim older than 7 days (oldest observed was `polylogue-9e5.31` at ~5.8 days) on the current live backlog, so the default 7-day `--stale-claim-days` threshold produces zero noise today.
- **Active-set summary**: 33 active leaves across 13 programs, `band=above-target` (soft target 30) — purely informational, does not affect exit code.

Not fixed in this PR (deliberately, per the "never mutate `.beads/`" constraint on this lane) — left as real backlog signal for a follow-up bead/triage pass.

## AC matrix (polylogue-8jg9.1, remaining scope after AC4/#3232)

| AC | Status |
| --- | --- |
| 1. Three-view policy as one checked concept, soft admission bands (~30 target / ~50 warn), never a hard cap | Satisfied — `compute_active_set_summary()`, band never enters `Finding`/exit code |
| 2. Bounded enumeration via pages/validated streaming | Satisfied — measured, not rewritten: existing single-pass bounded jsonl read already met this; proven and documented rather than replaced |
| 3. Epics-are-not-work; program grouping derived from member leaves | Satisfied — F1 (epic-as-leaf), F4 (program with no real members) |
| 4. (Execution focus fully derived from claims/readiness/priority/critical-path/conflicts/resource policy) | Partially satisfied — F3 guards claim liveness (one input to execution focus); full ranked-subset derivation is not implemented in this PR (see Solution's out-of-scope note) |
| 5. Seeded-violation tests per new check | Satisfied — 9 dedicated + 1 combined fixture |
| 6/7 (repo guidance, canonical commands, parent/child edge integrity) | Not addressed in this PR — outside the explicit remaining-scope list for this lane |

Ref polylogue-8jg9.1


Sinity:
Coordinator stand-in review (CodeRabbit rate-limited): read the full 867-line diff. F1–F4 are correctly hard structural findings while active-set size stays a non-gating soft-band diagnostic (the AC's no-hidden-truncation requirement holds by construction — the summary is not a Finding). The bounded-enumeration proof-not-rewrite call is right at 5MB/1,100 rows, and documenting it against the 7.88GB bd-list incident gives the future trigger condition. Clock injected via param, not mocked. The 6 real F4 findings are coordinator bookkeeping — triaging them now (re-admitting successor leaves / retiring stale program markers). Merging; 8jg9.1 stays open for execution-focus derivation + AC6/AC7.

PR #3245, opened 2026-07-21
fix(browser-capture): stable receiver identity survives token rotation

Sinity:
## Summary

Fixes the polylogue-jlme.5 receiver-pairing architecture defect: the browser-capture receiver's stable pairing identity was derived by hashing the bearer token, so an ordinary token rotation silently changed a paired extension's trusted identity. Also adds a deliberate dev-override pairing state so an operator-pinned dev-loop receiver never silently self-heals to the canonical endpoint.

## Problem

`receiver_identity(config)` (`polylogue/browser_capture/receiver.py`) hashed `config.auth_token` (or, in no-auth mode, the resolved spool path) to derive the `rx-<hash>` identity extensions treat as the pairing trust anchor. This meant:

- Every bearer-token rotation minted a *different* receiver identity, forcing every already-paired browser profile through an unnecessary re-pair -- directly blocking `gnie`'s automatic credential-refresh design (AC2: "token rotation is automatic when stable identity matches").
- The existing unit test suite encoded this as *expected* behavior (`test_receiver_identity_changes_on_token_rotation`), which the bead's 2026-07-16 note explicitly flags as the architecture correction still owed: "stable receiver identity must not be derived from the bearer token."

Separately, the already-shipped bounded canonical-endpoint recovery (`checkReceiverHealth` in `browser-extension/src/background.js`) treats *every* non-canonical, identity-matching pairing the same way: a stale receiver silently swaps the extension over to the canonical endpoint. That is the correct fix for the original 2026-07-15 incident (an extension retaining a dead dev-loop endpoint while the packaged receiver ran at the canonical port), but wrong for a profile an operator has *deliberately* pinned to a dev-loop receiver for active testing -- today it would silently jump away from the dev receiver mid-session.

## Solution

**Python (receiver-side identity):**
- `polylogue/paths/_roots.py`: new `browser_capture_receiver_identity_path()`, archive-root-scoped like the existing token path but a separate file (`browser-capture-receiver-id`).
- `polylogue/browser_capture/receiver.py`: new `load_or_mint_receiver_identity()` mints a non-secret `rx-<20 hex>` value once and persists it via the same `mkstemp` + `os.replace` pattern as the token minter (no `fchmod` needed -- nothing secret). `receiver_identity(config)` now reads/mints from that file instead of hashing `config.auth_token`/`config.spool_path`.
- Rewrote the three affected unit tests in `tests/unit/browser_capture/test_receiver.py` to assert the corrected invariant (identity survives token rotation and spool relocation, both scoped to archive root) instead of the old, now-wrong, expectation.
- `tests/unit/core/test_paths.py`: added the new path symbol to the `polylogue.paths` public-boundary allowlist.

**JS (deliberate dev-override, `browser-extension/src/`):**
- `background.js`: `saveReceiverSettings` now marks `dev_override: true` on the stored pairing whenever an operator explicitly configures a non-canonical `receiverBaseUrl` via the settings UI (and clears it when they repoint at the canonical endpoint -- both are equally explicit acts). `checkReceiverHealth`'s canonical-recovery gate now skips entirely when `pairingBefore.dev_override` is set. `markReceiverPairingUnavailable` marks such a pairing `state: "dev_override_stale"` (distinct from the calm `"offline"` state used for ordinary receiver-asleep framing). `observeReceiverIdentity`'s two branches that previously rebuilt the pairing object from scratch now preserve `dev_override` across every state transition.
- `operator_status.js` / `popup.js`: added a distinct, loud presentation for `dev_override_stale` in `receiverPairingPresentation`, `computeAttention`, `operatorStatusForState`, and `operatorPresentationForState`, at the same priority tier as an identity mismatch.

## Verification

- `devtools test tests/unit/browser_capture/ tests/unit/core/test_paths.py` -- 151 passed
- `mypy --strict polylogue/browser_capture/receiver.py polylogue/paths/_roots.py polylogue/paths/__init__.py tests/unit/browser_capture/test_receiver.py tests/unit/core/test_paths.py` -- clean
- `cd browser-extension && npx eslint src/ tests/` -- clean
- `cd browser-extension && npx vitest run` -- 342 passed, 1 pre-existing failure (`build.test.js` packaged service-worker fixture, a `vi.waitFor` timing flake) confirmed present on `master` before this change via `git stash`; untouched by this PR
- `devtools verify --quick` -- 16/16 steps green (run id `20260721T155909Z-quick-121970-01cf1062`)
- `devtools render all --check` -- exit 0, no `out of sync` surfaces

## Acceptance criteria matrix (polylogue-jlme.5)

1. **Stale-pairing → canonical recovery on identity+schema match** -- already shipped (2026-07-16 partial implementation); covered by pre-existing tests `recovers only at the canonical endpoint when the paired identity matches` / `does not adopt the canonical endpoint when its receiver identity differs`; both still pass unmodified. Not touched by this PR.
2. **Identity persisted independently of the bearer token; token rotation preserves identity; gnie's automatic refresh; different identity/schema never auto-adopts** -- **satisfied by this PR** for the identity/token-rotation half (`test_receiver_identity_survives_token_rotation`, `test_receiver_identity_is_minted_once_and_persisted_on_disk`). The extension-side identity-mismatch refusal was already shipped and is unchanged. `gnie`'s own automatic-bootstrap wiring is that bead's separate scope (noted on jlme.5 as an explicit dependent, not owned here).
3. **Re-pair preserves capture queue, backfill jobs, receipts, extension instance identity** -- already shipped (`polylogue.receiverPairing.reset` only clears the one pairing key); unchanged, covered by pre-existing test `resets only the pairing key and preserves pending work`.
4. **Popup/status show endpoint, identity, schema, last contact, current state without leaking credentials** -- already shipped via `receiverPairingPresentation`; this PR adds the missing `dev_override_stale` presentation branch so a deliberately pinned pairing reports loudly rather than falling into the calm offline framing.
5. **No arbitrary port scan / token disclosure / unauthenticated fallback / browser-profile special case** -- unaffected; this PR's recovery gate only *restricts* the existing bounded canonical probe (adds a `!dev_override` condition), it does not add any new endpoint discovery.
6. **Packaged two-authenticated-profile / dev-isolated-profile / canonical-failover / token-rotation / identity-mismatch proof** -- **not covered by this PR.** This sandboxed lane has no live browser harness (repo-documented constraint); the notes on jlme.5 already flag this as the standing residual requiring a real packaged-extension fixture. New fixture tests here (`test_receiver_identity_survives_token_rotation`, `does not auto-recover a deliberate dev-override pairing even when the canonical identity would match`, `marks a manually configured non-canonical endpoint as a deliberate dev override`) prove the underlying mechanics at the unit level, not the packaged end-to-end proof AC6 asks for.

## Residual risk

- `capture_jobs.py`'s per-lease HMAC proof (`CaptureJobRegistry._proof`) uses `receiver_id` as its HMAC key. This was already effectively non-secret before this PR (the identity is exposed via the authenticated `/v1/status` route to any bearer-token holder), so this change does not weaken that boundary, but it is worth a reader's attention if `receiver_id` is ever repurposed as an actual secret.
- AC6's packaged multi-profile proof remains open, as it was before this PR; no code claim is made against it here.

Ref polylogue-jlme.5


Sinity:
Coordinator stand-in review (CodeRabbit rate-limited): read the full 539-line diff. The root cause (receiver identity hashed from the bearer token, so rotation minted a new rx-id and broke pairing — with the old behavior actually pinned by a test as expected) is a real find; mint-once persisted non-secret identity is the correct fix, and the dev_override flag turns deliberate dev pairings into loud failures instead of silent canonical failover — matching the bead's incident semantics. Receiver-side + extension-side both covered by fixtures; 151 py tests + 342 vitest green, mypy strict + eslint clean. The lease-HMAC key-material note is accurate (receiver_id was already non-secret pre-change — no regression). Merging; AC6 packaged two-profile proof remains the documented open gap.

PR #3243, opened 2026-07-21
fix(config): route direct-env config bypasses through the layered resolver

Sinity:
## Summary

Lands the uu8r-class piece of the polylogue-9gh1 epic (config.py's documented 5-layer precedence vs actual runtime behavior) plus the fixture-driven regression test the epic's design requires across all three named bug shapes. The other two pieces (fd2s split-system delegation, cxlk nested-table deep-merge) were already fixed on master via PR #3079 before this branch was cut; this PR adds the regression coverage that was still missing for them and lands new fixes for the remaining env-bypass class.

## Problem

`investigations/config-resolution.md` (dogfood-2 round-2) found config.py's documented 5-layer precedence (defaults -> site TOML -> user TOML -> POLYLOGUE_* env -> CLI) silently does not hold for the whole inventoried surface, in three structurally distinct ways:

1. **fd2s-class** (split-system delegation) -- already fixed on master (PR #3079, `resolve_runtime_config`/`ResolvedRuntimeConfig`).
2. **cxlk-class** (nested-table full-replace) -- already fixed on master (`_deep_merge_table` in `polylogue/config.py:1749`).
3. **uu8r-class** (direct `os.environ` bypass) -- ~20 files read genuinely POLYLOGUE_*-namespaced, TOML-backed settings straight from `os.environ` instead of `load_polylogue_config()`, so TOML/site/user layering never reached those call sites even though every one of the settings involved was already fully inventoried in `config.py`.

None of the three bug shapes had regression coverage: `grep` for `_deep_merge_table`/`health_convergence_debt`/`ResolvedRuntimeConfig` across `tests/unit/core/test_config.py` before this PR turns up construction/precedence tests for flat scalar keys only, nothing that would fail if the nested-merge or delegation fixes were reverted.

## Solution

[...]

PR #3248, opened 2026-07-21
fix(config): route the un-inventoried env-only readers through config.py

Sinity:
## Summary

Lands the residual polylogue-uu8r-class scope PR #3243 explicitly deferred: 8 genuinely `POLYLOGUE_*`-namespaced settings that had **no config-inventory entry at all** (distinct from an already-inventoried key whose caller merely bypassed the resolver — that class was closed by #3243). Re-derived the file list live against master with `rg 'os\.environ.*POLYLOGUE_' polylogue/` rather than trusting the bead's stale 31-file/7-8-file counts, which had drifted after #3243 and #3079 landed.

## Problem

`investigations/config-resolution.md`'s dogfood-2 round-2 config investigation found `config.py` documents a 5-layer TOML/env precedence system, but several settings bypassed it entirely because they were never inventoried in the first place — so TOML could never reach them regardless of caller migration:

- `POLYLOGUE_HOOK_PROVIDER` (hook-harness detection override)
- `POLYLOGUE_RAW_AUTHORITY_COMMIT_BATCH_SIZE` (raw-materialization repair batching)
- 4 daemon parse-stage prefetch knobs: `POLYLOGUE_DAEMON_PARSE_STAGE_{WORKERS,MAX_INFLIGHT_BYTES,MAX_CACHED_TREE_BYTES,WARM_TIMEOUT_SECONDS}`
- 2 revision-parse dispatch thresholds: `POLYLOGUE_REVISION_PARSE_{DISPATCH_MAX_BYTES,POOL_MIN_BYTES}`

## Solution

- Added 8 `ConfigInventoryEntry` rows to `polylogue/config.py` plus matching `PolylogueConfig` properties, `_default_config_values` defaults, and `_INT_CONFIG_KEYS`/`_FLOAT_CONFIG_KEYS` coercion entries, following existing `owner_class`/`reload_behavior` conventions (mirrors the `ingest_commit_batch_messages`/`daemon_parse_stage_split` siblings).
- Migrated each call site's `os.environ.get(...)` read to `load_polylogue_config().<key>`, preserving each function's existing "explicit override wins, invalid/absent falls back to the module default" semantics exactly — no behavior change for unconfigured deployments.
- Documented all 8 new keys in `docs/configuration.md` (TOML-key table + env-var quick-reference table).

### Per-setting judgment table

| Setting | Decision | Rationale |
| --- | --- | --- |
| `hook_provider` | **Inventoried** (`sources.hook_provider`) | Operator-settable harness-detection override, same shape as the `theme` reference pattern (env wins, falls to TOML otherwise). |
| `raw_authority_commit_batch_size` | **Inventoried** (`pipeline.raw_authority.commit_batch_size`) | Genuine perf tunable, direct sibling of `ingest_commit_batch_messages`. |
| `daemon_parse_stage_workers` / `max_inflight_bytes` / `max_cached_tree_bytes` / `warm_timeout_seconds` | **Inventoried** (`daemon.raw_materialization.parse_stage_*`) | Daemon-owned resource-policy knobs, direct siblings of `daemon_parse_stage_split`; read from `DaemonParseStage.__init__` in-process (ThreadPoolExecutor, never a spawned worker) — no multiprocessing-spawn hazard. |
| `revision_parse_dispatch_max_bytes` / `pool_min_bytes` | **Inventoried** (`pipeline.revision_parse.*`) | Perf-tuning thresholds with measured rationale in their docstrings; read in dispatcher/orchestration code before `ProcessPoolExecutor` creation, never inside a worker — no spawn hazard. |
| `POLYLOGUE_DEV_LOOP_RUN_ID` / `POLYLOGUE_DEV_LOOP_LOG_DIR` | **Intentional env-only exemption** | Launcher-injected correlation metadata for one process invocation, structurally identical to `CODEX_SESSION_ID`/`CLAUDE_SESSION_ID` (already out of the config domain), not an operator TOML preference. Documented at each call site + `docs/configuration.md`. |
| `POLYLOGUE_SESSION_REF` | **Intentional env-only exemption** | Same class as above — grouped with `CODEX_THREAD_ID`/`CODEX_SESSION_ID`/`CLAUDE_SESSION_ID` in `coordination/envelope.py`. |
| `POLYLOGUE_CONFIG` | **Intentional bootstrap-only exemption** | Selects *which* user TOML file the layered resolver loads, so it cannot itself be resolved through that resolver — mirrors `config.py`'s own `_user_config_path`/`_site_config_path` bootstrap reads. |
| `POLYLOGUE_ARCHIVE_ROOT` in `daemon/http.py::_dev_loop_payload` | **Left raw (diagnostic probe)** | `archive_root` itself IS inventoried, but this endpoint reports whether the daemon's *own environment* has the var set (launcher diagnosis), not a config value; routing through the resolver would mask an unset env var behind the TOML/default fallback and defeat the probe's purpose. |

## Verification

- `devtools test tests/unit/core/test_config_resolution_regression.py`: 16 passed (6 pre-existing fd2s/cxlk/uu8r-class + 10 new — one regression class per newly-inventoried setting family, against its real consumer, each with a revert-witness docstring). Live-verified the raw-authority witness by reverting that call site to `os.environ.get` and confirming the test fails (`20 != 4242`) before restoring the fix.
- Touched-consumer files, run individually: `test_config.py` (95), `test_config_inventory.py` (60), `test_hooks.py` (8), `test_repair.py` (55), `test_revision_backfill.py` (44), `test_parse_prefetch.py` (12), `test_envelope.py` (26), `test_embed.py` (18), `test_backup.py` (23) — all passed.
- `devtools verify --quick`: ruff format/check, mypy --strict, `render all --check` (grepped for "out of sync": none), topology/layering/closure-matrix, docs-coverage, and the rest of the quick gate all green (`exit_code: 0`), both standalone and via the pre-push hook.

## AC matrix (polylogue-uu8r)

| AC | Status |
| --- | --- |
| Every direct `os.environ` POLYLOGUE_* read outside `config.py` is classified | Satisfied — re-derived via `rg`, 7 files / 8 settings genuinely un-inventoried; 4 settings classified as intentional env-only/bootstrap exemptions with rationale documented at call sites + docs. |
| All class-(a) (genuine TOML-backed) sites route through `config.py` | Satisfied — 8 settings, 7 files. |
| Config diagnostics output includes the migrated settings | Satisfied — driven automatically from `_CONFIG_INVENTORY` (`config_inventory_payload`, `_toml_section_layout`, `_ENV_CONFIG_KEY_MAP` all derive from the inventory list). |
| `devtools test` on affected files + `config.py` passes | Satisfied — see Verification above. |

Ref polylogue-uu8r; last open child of polylogue-9gh1.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -344,12 +344,20 @@ A few keys not shown in the full example above, with their TOML path:
 | `debug_timing` | `ui.debug_timing` | Emit per-stage timing diagnostics in CLI output. |
 | `hermes_root` | `sources.hermes.root` | Runtime root watched for Hermes state, snapshots, NeMo Relay ATIF/ATOF artifacts, and verification evidence. Defaults to `~/.hermes`. |
 | `hook_sidecar_dir` | `sources.hook_sidecar_dir` | Directory for hook-event sidecar files consumed by the Claude Code/Codex hook harness. Defaults to `<archive_root>/hooks`, so a scratch/test `POLYLOGUE_ARCHIVE_ROOT` genuinely isolates its hook spool from the real one; set explicitly only if you need the spool somewhere other than the archive it belongs to. |
+| `hook_provider` | `sources.hook_provider` | Force hook-event harness detection to `claude-code` or `codex` instead of sniffing the payload shape; unset auto-detects. |
 | `backup_verify_tmpdir` | `maintenance.backup_verify_tmpdir` | Scratch directory for backup-restore verification; defaults to the system temp dir when unset. |
 | `antigravity_language_server` | `sources.antigravity_language_server` | Path to an Antigravity language-server binary, when parsing Antigravity sessions needs it. |
 | `ingest_commit_batch_messages` | `sources.ingest_commit_batch_messages` | Messages per commit batch during ingest (default 8000). |
 | `ingest_parse_workers` | `sources.ingest_parse_workers` | Parallel parse workers during ingest (default 1). |
 | `live_full_ingest_workers` | `sources.live_full_ingest_workers` | Parallel workers for a live full-reingest pass (default 1). |
+| `raw_authority_commit_batch_size` | `pipeline.raw_authority.commit_batch_size` | Census-phase commit batch size for raw-materialization repair (polylogue-amg1); unset uses the built-in default, `<=0` disables batching (per-raw commits). |
+| `revision_parse_dispatch_max_bytes` | `pipeline.revision_parse.dispatch_max_bytes` | Payload-size ceiling (bytes) above which a raw parses sequentially in-process instead of dispatching to the parse pool (polylogue-amg1). |
+| `revision_parse_pool_min_bytes` | `pipeline.revision_parse.pool_min_bytes` | Aggregate pool-eligible payload floor (bytes) below which pool dispatch cannot amortize worker-spawn cost (polylogue-amg1 / polylogue-crd8). |
 | `daemon_parse_stage_split` | `daemon.raw_materialization.parse_stage_split` | Opt-in (polylogue-m6tp phase (a), default off): pre-parse raw-materialization census candidates in a bounded daemon-owned thread pool before the writer hold, instead of parsing inside the writer-held pass. |
+| `daemon_parse_stage_workers` | `daemon.raw_materialization.parse_stage_workers` | Worker cap for the daemon-owned pre-parse thread pool; unset/`<=0` uses the adaptive `cpu_count - 1` default. |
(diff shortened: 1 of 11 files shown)


Sinity:
Coordinator review (CodeRabbit rate-limited): full diff read. Migration patterns match #3243's provenance conventions; the exemption table (dev-loop correlation ids, POLYLOGUE_CONFIG bootstrap, raw diagnostic probe) is correctly reasoned and documented at call sites + docs/configuration.md. NOTE: the lane's repair.py edit was lost to worktree auto-cleanup before commit (its own regression test failed on the pushed branch — caught by coordinator re-run); reconstructed in a766b8a9a with identical env-escape-hatch semantics, 10/10 regression tests + mypy strict green. The process_pool.py leftover bypass for the already-inventoried parse-workers key is noted as a known residual. Merging; this closes the last open child of polylogue-9gh1.

PR #3263, opened 2026-07-22
fix(storage): guard-gated bulk delete_sessions — one-pass derived maintenance instead of per-row trigger detonation

Sinity:
## Summary

`ArchiveStore.delete_sessions` (the product bulk-delete API used by the CLI
`delete` verb and `SessionDeleteActuator`) now wraps its whole batch in the
same `derived_refresh_guard` rows the bulk session-write path already uses,
and performs the derived FTS/trigram/action_pairs/delegation_facts
maintenance explicitly, one pass, instead of relying on unsuppressed
per-row triggers.

## Problem

`delete_sessions` issued a plain per-session `DELETE FROM sessions`.
`blocks_action_pairs_ad` (`storage/sqlite/archive_tiers/index.py`) fires
once per deleted `blocks` row — both the explicit block cascade and the FK
`ON DELETE CASCADE` from `messages`/`sessions` still invoke AFTER DELETE
triggers per row — and each firing deletes+rebuilds the **entire session's**
`action_pairs` via two window-function scans and re-derives
`delegation_facts`. Live incident 2026-07-21: deleting 91 sessions ran 3h,
375GB reads, zero commit, before being killed.

The bulk session-write path already suppresses this class of cost via
`derived_refresh_guard` rows (`'session-write'`, `'fts-bulk-session-write'`),
see `_bulk_fts_session_guard` in `write.py` (bead polylogue-crd8's fix for
the analogous whale-lineage-prefix-delete stall). `delete_sessions` never set
either guard.

## Solution

`ArchiveStore.delete_sessions` (`polylogue/storage/sqlite/archive_tiers/archive.py`):

1. For each session id, explicit session-scoped deletes run first, while the
   `blocks` rows still exist (external-content FTS5 needs the OLD text to
   locate its postings): `messages_fts` / `messages_fts_identity`
   (`delete_session_rows_sql` / `delete_session_identity_rows_sql`) and
   `blocks_command_trigram` (`trigram_delete_session_rows_sql`, landed in
   #3259).
2. Both `derived_refresh_guard` rows are set for the whole batch —
   `'session-write'` (suppresses `blocks_action_pairs_a{i,d,u}` and the
   `delegation_facts` triggers) and `'fts-bulk-session-write'` (suppresses
   the `messages_fts`/`blocks_command_trigram` trigger bodies — redundant
   here since those rows are already gone by step 1, kept for guard-contract
   parity and defense-in-depth).
3. `action_pairs` and `delegation_facts` are explicitly cleared by session
   id — belt-and-suspenders alongside the `ON DELETE CASCADE` FKs those
   tables already carry against `sessions`/`blocks`; this keeps the
   post-condition legible without depending on cascade-ordering.
4. `DELETE FROM sessions` per id relies on indexed FK cascades (not per-row
   trigger logic) for physical `messages`/`blocks` tree removal; guard rows
   are cleared and the transaction commits once for the whole batch.

This mirrors `_bulk_fts_session_guard`'s delete-then-guard-then-mutate shape.

**Epoch triggers audit** (per bead's ask): `query_unit_frame_{sessions,
messages,blocks}_delete` fire per deleted row, ungated by either guard. Left
deliberately ungated — each firing is a single primary-key `UPDATE
query_unit_frame_state SET epoch = epoch + 1 WHERE singleton = 1` against a
one-row table (O(1) per row), not a window-function rebuild, so this was
never implicated by the incident and folding a third guard in for a
zero-measured cost isn't warranted.

**Async path** (`repository/archive/writes/sessions.py` ->
`delete_session_via_backend` -> `queries/sessions_writes.py`'s
`delete_session_sql`): unaffected by this fix and left unchanged. It deletes
**exactly one session per call/transaction** — there is no loop batching N
session ids into a single transaction, so the "91 sessions in one
transaction" incident shape does not reproduce there. (A single very-large
session could still see a bounded version of the same per-block trigger cost
for its own row count — that's the pre-existing, already-tracked crd8
concern for whale *writes*, not a bulk-*delete* regression, and out of this
bead's scope.)

## Verification

- `devtools test tests/unit/storage/test_bulk_delete_guarded.py tests/unit/storage/test_bulk_fts_prefix_reextract.py`
  -> `9 passed`
- **Anti-vacuity proof**: `sqlite3.Connection.set_trace_callback` was tried
  first and rejected — it only reports the outer statement text (`DELETE
  FROM sessions ...`), never SQL executed *inside* a trigger body (verified
  empirically against a minimal repro). Instead, the new
  `test_delete_sessions_bulk_never_fires_unguarded_per_row_canary` installs
  canary triggers on `blocks` carrying the **exact same WHEN-clause guard
  conditions** as the real `blocks_action_pairs_ad` /
  `blocks_command_trigram_ad` triggers. With the fix reverted (`git stash`
  on `archive.py` only, canary test re-run), the canaries fire
  `{'session_write': 14, 'fts_bulk': 6}` for 2 sessions x 3 tool pairs each
  — i.e. the per-row rebuild machinery *would* have detonated. With the fix
  restored, both counts are `0`. What would make this test fail: removing
  or narrowing the scope of either `INSERT OR REPLACE INTO
  derived_refresh_guard(...)` call this PR adds.
- `devtools verify --quick` -> `exit_code: 0` (ruff format/check, mypy
  --strict, `render all --check`, topology/layering/manifests/doc-commands/
  docs-coverage/test-clock-hygiene/pytest-timeout-overrides/degrade-loudly)
- `tests/unit/storage/test_archive_tiers_archive.py::test_exact_session_action_count_bounds_pairing_before_global_ranking`
  fails identically with the fix reverted (`git stash` on `archive.py` only)
  — confirmed pre-existing and unrelated to this change, not touched by this
  PR.
- Not run: full `devtools verify --all` / broad directory pytest (repo
  convention is testmon-affected + focused files, not blanket runs).

Ref polylogue-meoz. Relates to polylogue-crd8 (same `derived_refresh_guard`
mechanism, different write path — whale prefix-delete during ingest vs. bulk
session delete via the product API).

### DIFF
--- a/polylogue/storage/sqlite/archive_tiers/archive.py
+++ b/polylogue/storage/sqlite/archive_tiers/archive.py
@@ -138,6 +138,12 @@ from polylogue.sources.dispatch import merge_parsed_session_chunks
 from polylogue.sources.parsers.base import ParsedSession
 from polylogue.storage.fts.fts_lifecycle import repair_message_fts_index_sync
 from polylogue.storage.fts.session_repair import repair_session_fts_if_needed_sync
+from polylogue.storage.fts.sql import (
+    FTS_BULK_SESSION_WRITE_GUARD,
+    delete_session_identity_rows_sql,
+    delete_session_rows_sql,
+    trigram_delete_session_rows_sql,
+)
 from polylogue.storage.insights.session.records import SessionProfileRecord
 from polylogue.storage.insights.session.runtime import (
     SESSION_INSIGHT_MATERIALIZATION_TYPES,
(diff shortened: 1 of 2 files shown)


Sinity:
Fixed the P1: `delete_sessions` now checks (once, cheaply, via `sqlite_master`) whether the live `blocks_command_trigram_ad` trigger actually references `derived_refresh_guard` before relying on the guard. When it doesn't (an archive last rebuilt before #3259), it skips its own explicit trigram pre-delete and the FTS bulk guard row entirely, letting that same ungated trigger clean up per-block during the cascade — correct, just unoptimized for trigram until the archive's next rebuild. New regression test reproduces the exact "database disk image is malformed" error against the pre-fix code (verified it fails without the fix, passes with it) using a downgraded trigger fixture, no schema-version bump needed since this is a runtime capability check, not a migration. Verified: 3 tests pass, `devtools verify --quick` green.

PR #3264, opened 2026-07-22
fix(cli): unify referenced_path matching through the substrate predicate and delete dead stats aggregators

Sinity:
## Summary

Two related fixes to `polylogue/cli`'s query-stats surface, both under polylogue-t46.6:

1. **Correctness**: `cli/query_semantic.py`'s `referenced_path_matches_slice` used `any(term...)` (OR-of-terms) for a multi-term `--path` filter, while the substrate's `archive/query/runtime_matching.py:matches_referenced_path` used `all(term...)` (AND-of-terms). The CLI semantic-stats surface therefore selected a *different session set* than the actual query filter for the identical `--path X --path Y` query.
2. **Deletion**: `cli/query_stats.py`'s origin/date grouping, semantic action/tool grouping, and profile repo/work-kind grouping re-derive in Python what `ArchiveStore.stats_by` (SQL) already computes, and had zero live callers.

## Problem

**(1)** `query_semantic.referenced_path_matches_slice(action, terms)` checked `any(term in path for term in terms)` against a *single* action's `affected_paths`. `runtime_matching.matches_referenced_path(plan, session)` checked `all(term in some path for term in terms)` against paths aggregated across *every action in the session*. Same input query, two different predicates — the CLI semantic-stats surface (`output_stats_by_semantic_ids` / `analyze --by action|tool`) could report a session as matching a `--path` filter that the actual query filter (`find --path ...`) would exclude, or vice versa.

**(2)** Caller audit (grep across `polylogue/` and `tests/`, not just this bead's cited line numbers, which had drifted from a 2026-07-06 snapshot): `output_stats_by_summaries`, `output_stats_by_grouped_sessions`, `output_semantic_grouped_stats`/`_semantic_grouped_payload`, and `output_stats_by_profile_ids`/`_query`/`_summaries` have **zero callers** outside this module's own re-exports (`query_output.py`) and tests. The live `analyze --by <dimension>` command (`query_verbs.py`'s `stats_by` branch → `_execute_query_verb` → `archive_query.py`'s `stats_by` dispatch) always calls `ArchiveStore.stats_by(group_by, ...)` — a SQL aggregator supporting `origin`, `day`/`month`/`year`, `action`, `tool`, `repo`, `work-kind`, `tag`, `role`, `message_type`, `material_origin` — never this module's Python-side re-derivation.

## Solution

**(1)** Extracted the AND-of-terms path predicate into a single shared function, `paths_match_referenced_terms`, in `polylogue/archive/query/runtime_matching.py`. Both the substrate's `matches_referenced_path` and the CLI's semantic-stats matching now route through it. The CLI's referenced-path check moved from per-action to session-level aggregation (`session_matches_referenced_path`, aggregating affected paths across every action in a session) to match how the substrate aggregates — same operator (AND) *and* same aggregation granularity are both required for the two surfaces to select the same session set. Non-path per-action filters (action/tool/text terms) keep per-action granularity, renamed `action_matches_dimension_filters` to make the split explicit. Deleted `referenced_path_matches_slice`/`action_matches_slice`.

**(2)** Deleted the dead aggregators, their now-orphaned shared helpers (`_emit_grouped_stats_table`, `_count_value`, `_formatted_group_label`, `_sort_group_keys`, `_summary_group_key`, `_session_group_key`) and dimension/column constants. Kept `emit_structured_stats` (JSON/YAML/CSV serialization, still used by `output_stats_sql`) and `output_stats_sql` itself. Updated `query_output.py`'s re-exports/aliases. Deleted the tests that existed solely to pin the deleted shape, and the now-orphaned `SummaryStatsCase`/`summary_stats_case_strategy` Hypothesis generator (zero consumers).

**Out of this PR's audited-but-not-named scope**: `output_stats_sql` also appears to have zero live callers (the live `analyze` stats_only path goes through `archive_query.py`'s `archive.stats(...)` + `_emit_stats`, a different function). Left in place since the bead named only origin/date, semantic, and profile grouping for deletion — flagging for a follow-up rather than expanding this PR.

## Verification

- `devtools test tests/unit/cli/test_query_support_runtime.py tests/unit/cli/test_query_fmt.py tests/unit/archive/test_query_runtime_matching.py tests/unit/archive/test_query_runtime_filters.py` → `71 passed`
- Mutation check: reverting `paths_match_referenced_terms` usage in `query_semantic.py` back to hand-rolled `any(...)` makes the new `test_semantic_stats_referenced_path_matches_substrate_query_filter` fail (`assert True is False`), confirming the regression test actually catches the bug it targets.
- `devtools verify --quick` → `exit_code: 0` (ruff format, ruff check, `mypy --strict`, `render all --check`, topology/layering/closure-matrix/schema-roundtrip/manifests/ci-workflows/doc-commands/docs-coverage/test-infra-currency/test-clock-hygiene/pytest-timeout-overrides/degrade-loudly all `ok`)
- Not run: the full non-quick test suite / `devtools verify --all` (not required for this scope; the targeted+affected-area tests plus mypy --strict cover the changed surface).

Ref polylogue-t46.6


Sinity:
Fixed: re-exported `paths_match_referenced_terms` from `archive.query.runtime` (the existing public boundary) and pointed `query_semantic.py`'s import at it instead of the substrate-internal `runtime_matching` module. Verified: touched-area tests pass, `devtools verify --quick` green.

PR #3265, opened 2026-07-22
fix(sources): persist hook events as session-linked evidence, not standalone sessions

Sinity:
## Summary

Hook events (PreToolUse/PostToolUse/UserPromptSubmit/…) were being ingested as
standalone `codex`/`claude-code` **sessions**, inflating the archive with
content-less session shells. This lands the going-forward fix (hooks persist as
session-linked evidence, no `raw_sessions` row) plus a tested retroactive repair
for the shells already accumulated. The session↔hooks read-model + read surfaces
follow in a fast-follow PR.

## Problem

`sources/hooks.py:_persist_record` drained each spooled hook event through
`write_source_raw_session` with `origin=codex-session|claude-code-session`,
minting a full `raw_sessions` row per hook. The materializer turned each into an
empty standalone index session. On the live archive `/realm/db/polylogue`:

- index sessions = **83,279**, but only **17,553 have any messages**; the other
  **65,727 are empty shells** (codex 35,233 + claude-code 30,488).
- `source_path LIKE '%/hooks/pending/%'` raws = codex 35,216 + claude-code
  29,679 + hermes 1 = **64,896**, exactly the `raw_hook_events` row count.

Each hook was double-recorded: correctly as a `raw_hook_events` row (which
carries `session_native_id`; that table is indexed
`(origin, session_native_id, observed_at_ms)` precisely to attach a hook to its
session), and wrongly as a `raw_sessions` row. `raw_hook_events` has no FK to
`raw_sessions`, so the linkage never needed a session row. Nothing was lost —
every raw blob is retained.

## Solution

- **`write_source_hook_event`** (`source_write.py`): writes the
  `raw_hook_events` row + a durable `raw_payload` blob ref, but **no
  `raw_sessions` row**. Blob GC roots off `blob_refs` independently of
  `raw_sessions`, so the bytes stay retained.
- **`ArchiveStore.write_hook_event`** (`archive.py`): publishes the blob then
  routes to the hook path with a deterministic raw_id.
- **`_persist_record`** uses it; covers codex/claude-code/hermes uniformly.
- **`maintenance/hook_deinflation.py`**: `repair_hook_session_inflation` deletes
  the historical hook `raw_sessions` rows (source, durable) + their empty index
  sessions, while KEEPING the `raw_hook_events` rows and retained blobs.
  Identification is exact (`source_path LIKE '%/hooks/pending/%'`, 1:1 with
  `raw_hook_events`); genuinely-empty real sessions (content quarantined /
  zero-turn) do not match and are never touched (index delete additionally
  guarded by `message_count=0`). The four source FKs into `raw_sessions` are
  `ON DELETE CASCADE` and hook raws never appear in raw-authority
  plans/blockers, so the durable delete self-cleans without orphans and does not
  perturb convergence. `dry_run=True` reports counts without mutating.

The repair runs against the live archive only after this fix is deployed to the
daemon (otherwise the running old daemon re-inflates); backup already taken at
`/realm/staging/polylogue-sqlite/recovery/t93b-preflight-20260722-durable`.

## Verification

- `devtools test tests/unit/sources/test_hook_spool.py tests/unit/storage/test_hook_event_materialization.py tests/unit/maintenance/test_hook_deinflation.py tests/unit/daemon/test_hook_liveness.py` — **38 passed**.
  - New guards: draining a hook asserts 0 raw_sessions / 1 raw_hook_events / 1 blob_ref across codex/claude/hermes; the repair test asserts only hook shells are removed while hook evidence + blobs + real + genuinely-empty-non-hook sessions survive.
- `devtools verify --quick` — exit 0 (format + lint + mypy --strict + render all --check + layering).
- **Live dry-run** against `/realm/db/polylogue`: `hook_raw_sessions=64896, hook_index_sessions=64895, raw_hook_events unchanged=64896, blobs retained=64896`.

Ref polylogue-31r1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hook events no longer create empty standalone sessions, preventing archive inflation.
  * Existing inflated hook sessions can be safely removed while preserving hook evidence and payloads.
  * Added dry-run reporting before applying archive repairs.

* **Tests**
  * Added coverage confirming hook events are retained without creating session records.
  * Added validation for repair behavior, including preservation of real sessions and hook evidence.

* **Documentation**
  * Updated topology and status metrics and refreshed plan formatting.


Sinity:
Addressed both Codex P1 findings in 9ffa4833a:

1. **Verified hook rows** — the durable delete now requires a `raw_hook_events` row written from the same spool file (`source_path IN (SELECT source_path FROM raw_hook_events)`), not just a path-substring match. A legitimate raw whose path merely contained `/hooks/pending/` is no longer deletable without hook evidence. Live dry-run confirms the verified predicate still matches all 64,896 (the 1:1 holds).
2. **Daemon coordination** — `apply` now calls `offline_maintenance_block_reason` and raises if `polylogued` is running, matching every other offline-maintenance mutation; `dry_run` stays read-only and allowed. New test asserts apply raises and mutates nothing while a daemon is reported live.

PR #3266, opened 2026-07-22
fix(maintenance): prune orphaned raw-authority plans in hook-deinflation repair

Sinity:
## Summary

Follow-up to #3265. The hook-deinflation repair deleted hook `raw_sessions` but left their raw-authority frontier plans/blockers/census rows dangling (those tables reference raws by JSON string, not FK, so no cascade cleaned them). On the live archive that orphaned 64,895 plans and the daemon reconciler threw on a dangling plan. This prunes the orphaned authority state as part of the repair, with the performance work needed to do it on the real table sizes.

## Problem

First daemon convergence pass after the live hook de-inflation:
```
RuntimeError("duplicate strategy did not reach its typed terminal postcondition")
```
`raw_authority_plans` (and `raw_authority_blockers`/`census_plans`/`census_post_plans`) reference raws via `input_raw_ids_json`, not a foreign key, so deleting a raw does not cascade-clean its plans. 64,895 frontier plans were left pointing at deleted hook raws; the reconciler chokes on them. (This also confirms the hook raws were flooding the raw-authority machinery — ~65K plans/blockers of noise.)

## Solution

- **Prune purely-orphaned plans** (every input raw gone; verified 0 reference any present raw) + their blocker/census children, children-first (`NO ACTION` FKs to plans), in the same source transaction after the raw delete. New `orphaned_authority_plans` report field.
- **Two performance fixes** proven necessary on the live 84k-plan / 405k-census tables (a naive form hung >1h):
  - Set-based orphan identification (`json_each` expand + `LEFT JOIN` on the `raw_id` PK + `GROUP BY`) replaces per-plan correlated `NOT IN` subqueries — **~26 billion ops → 0.3s**.
  - Temporary full `plan_id` indexes on the three child tables (their existing indexes are partial / non-leading on `plan_id`) so both the `IN`-delete and the FK RESTRICT check on the parent delete are index-driven; dropped before commit (durable schema unchanged). Whole cleanup **~13s**.

## Verification

- `devtools test tests/unit/maintenance/test_hook_deinflation.py` — 4 passed (new `test_repair_prunes_orphaned_raw_authority_plans` seeds a hook plan+blocker and a real plan, asserts only the orphan is pruned; `orphaned_authority_plans == 1`).
- `devtools verify --quick` (pre-push) — passed.
- **Live**: pruned 64,895 orphaned plans (plans 84,042→19,147, blockers 70,887→5,992, census_plans 405,234→275,444, census_post_plans 323,877→258,982); 0 orphans remain; daemon restarted with **0 tracebacks over 8 min**; `raw_hook_events` (64,896) and hook blobs (64,896) untouched.

Ref polylogue-31r1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Maintenance cleanup now removes authority plans that become orphaned after invalid hook session data is deleted.
  * Related blocker and census records are also removed, while valid plans and hook evidence remain intact.
  * Cleanup reports now include the number of orphaned authority plans found or removed.

### DIFF
--- a/polylogue/maintenance/hook_deinflation.py
+++ b/polylogue/maintenance/hook_deinflation.py
@@ -43,6 +43,27 @@ _HOOK_RAW_PREDICATE = (
     "source_path LIKE '%/hooks/pending/%' AND source_path IN (SELECT source_path FROM raw_hook_events)"
 )
 
+# Raw-authority tables reference raws by JSON string (input_raw_ids_json), not by
+# FK, so deleting a raw does NOT cascade-clean its frontier plans/blockers/census
+# rows. A plan whose EVERY input raw is gone is unprocessable and must be pruned,
+# or the daemon reconciler throws ("duplicate strategy did not reach its typed
+# terminal postcondition") on the dangling plan (live incident 2026-07-22, first
+# convergence pass after the hook de-inflation). These are children of
+# raw_authority_plans with NO ACTION (RESTRICT) FKs, so delete children first.
+# Set-based (single pass, LEFT JOIN on the raw_id PK): expand each plan's input
+# raws, GROUP BY plan, keep plans whose inputs are ALL missing. The equivalent
+# correlated-subquery form ran ~26 billion ops (>1h) on the live archive; this
+# form completes in ~0.3s. A plan with empty inputs produces no json_each rows
+# and is correctly excluded (it is not orphaned-by-missing-raw).
+_PURELY_ORPHANED_PLANS_SQL = """
+    SELECT p.plan_id
+    FROM raw_authority_plans p, json_each(p.input_raw_ids_json) j
+    LEFT JOIN raw_sessions r ON r.raw_id = j.value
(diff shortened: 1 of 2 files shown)


Sinity:
Closing: the orphan-plan pruning approach here fixes the dangling-plan "duplicate strategy" error but breaks a different census invariant — the carried-forward/retryable plan persistence postflight (RuntimeError: "raw authority postflight changed a retryable/carried-forward plan"), exactly the L122 review finding. Deleting plans out from under the census ledger is not the right primitive.

Operator decision 2026-07-22: leave the census as-is (daemon defers failed passes to convergence_debt, survives; archive data is correct and queryable), and fold the raw-authority census repair into the t93b/hjpx/lkrc convergence work where the full ledger-invariant model is handled — not more surgical deletes. The live archive already had this pruning applied once; the t93b census rebuild will reconcile it. Ref polylogue-31r1, polylogue-t93b.

PR #3282, opened 2026-07-26
perf(storage): defer FTS repair off the live-ingest write path

Sinity:
## Summary

Continuation of the resumable-bulk-FTS perf line (#3281): moves message FTS/trigram repair off the sole-writer hot path for live ingest, catch-up, and membership replay, and adds daemon catch-up ordering fixes found along the way.

## Problem

A 19 MB live Codex session held the sole writer for over 31 minutes doing targeted FTS repair inline, even though its raw acquisition and canonical session projection were already durable. Membership-governed replay and full-replace paths had the same shape: FTS teardown/rebuild ran synchronously inside the write transaction, and some of it duplicated work the post-write repair pass would redo anyway. Separately, catch-up ordering let historical backlog starve live sources, and unresolved-but-decided membership conflicts were mis-counted as retryable failures, causing repeated reparsing on daemon restart.

## Solution

- Live authoritative replay, membership replay, and full-replace paths now commit canonical state, mark FTS stale, and queue the existing session-scoped FTS convergence debt instead of repairing inline.
- Eliminate a duplicate authoritative-replay FTS rebuild that ran once inline and again in the deferred repair.
- FTS trigram/session cleanup seeks rows by canonical block rowid join instead of a virtual-table `IN` subquery.
- Full-replace FTS teardown timing is now split into distinct stages (message FTS / identity-ledger / trigram deletes) for observability.
- Full ingest convergence is skipped entirely for cursor-only observations that didn't change a session.
- Decided-but-unresolved membership conflicts are preserved as cursor-complete raw authority debt instead of retried as transient failures.
- Daemon catch-up now prioritizes live sources over historical backlog, and defers Drive catch-up until local capture completes.
- `rebuild_index` bulk FTS materialization checkpoints progress (base for #3281, rebased here after that merge).

## Verification

- `devtools verify --quick` — pass (ruff format/check, mypy, render all, topology/layering/closure-matrix/manifests/ci-workflows/doc-commands/docs-coverage/test-infra-currency/test-clock-hygiene/pytest-timeout-overrides/degrade-loudly).
- `devtools test tests/unit/sources/test_live_batch_support.py tests/unit/sources/test_live_catchup_planning.py tests/unit/storage/test_revision_replay.py tests/unit/storage/test_fts_identity_ledger.py tests/unit/storage/test_fts_repair_sql.py tests/unit/storage/test_bulk_fts_prefix_reextract.py tests/unit/daemon/test_daemon_cli.py` — all passing (see individual commit messages for per-commit pass counts).

Ref polylogue-6mvg

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup ordering so background synchronization begins only after watcher catch-up completes.
  * Prioritized recently updated source files during catch-up while continuing to process historical backlog.
  * Reduced unnecessary convergence work when ingestion produces no session changes.
  * Preserved unresolved membership decisions for later handling instead of incorrectly marking ingestion as failed.
  * Improved search index rebuilding and cleanup to prevent stale or missing results during bulk updates and session replacement.

* **Tests**
  * Added coverage for catch-up ordering, deferred convergence, membership outcomes, and search index consistency.

### DIFF
--- a/polylogue/daemon/cli.py
+++ b/polylogue/daemon/cli.py
@@ -518,13 +518,20 @@ async def _run_drive_source_catchup_safely() -> int:
         return 0
 
 
-async def _periodic_drive_source_catchup() -> None:
+async def _periodic_drive_source_catchup(
+    *,
+    catch_up_complete: asyncio.Event | None = None,
+) -> None:
     """Periodically converge remote Drive sources such as AiStudio exports.
 
-    The first pass runs immediately in the background: Drive catch-up is
-    serial network I/O and must never sit between daemon startup and the
-    local convergence loops or the live watcher.
+    The first pass normally runs immediately in the background.  A live
+    watcher supplies ``catch_up_complete`` so fresh local session evidence
+    gets the single archive writer before remote download/index work.  The
+    gate is deliberately absent for maintenance-only callers.
     """
+    if catch_up_complete is not None:
+        await catch_up_complete.wait()
+
     while True:
         coordinator = daemon_write_coordinator()
         changed = await coordinator.run("maintenance.drive_catchup", _run_drive_source_catchup_safely)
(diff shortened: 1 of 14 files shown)


Sinity:
Fixed the critical finding: `repair_message_fts_index_sync` now also repopulates `blocks_command_trigram` per session (commit 950bed0fa), with a regression test proving a deferred full-replace's trigram posting survives catch-up repair. Also fixed the two other CI-relevant findings (frozen_clock instead of monkeypatching `time.time` directly, and the pre-existing-on-master `FakeResolved` config-double gap in `test_bulk_rebuild_routing_resumable_transaction_drives_pass_even_below_threshold`).

The remaining 4 minor/trivial findings in `test_live_batch_support.py` (asserting cursor-complete state more directly rather than only via succeeded/failed lists) are real quality improvements but not bugs — none of the current assertions are wrong. Tracked as polylogue-y9d0 rather than blocking this merge.

PR #3286, opened 2026-07-26
feat(daemon): close two automagic-invariants convergence gaps

Sinity:
## Summary

Closes two verified gaps found while auditing the daemon's "automagic invariants" doctrine against actual behavior: blob-GC had no periodic daemon path at all, and `embedding-orphan-reconcile`'s manual apply path was redundant with (and admittedly demoted-not-deleted next to) its own automatic daemon equivalent.

## Problem

An audit comparing this repo's stated doctrine ("if Polylogue can maintain a condition fully automatically, it generally should... the redundant manual surface is DELETED, not demoted to escape-hatch status", polylogue-gd6v/4jsk) against actual runtime behavior found:

1. `polylogue ops maintenance blob-gc` (reclaiming unreferenced, aged-out content-addressed blobs — a purely mechanical refcount/age decision) had **no daemon-periodic equivalent at all**, unlike every comparable mechanical maintenance operation (FTS merge, WAL checkpoint, embedding orphan reconcile).
2. `embedding-orphan-reconcile`'s `--yes` apply path is redundant: `periodic_embedding_orphan_reconcile_check` has run the exact same reconciliation automatically every 900s since it shipped, and the CLI docstring already said as much — "this command remains the break-glass inspect/apply path" is the precise anti-pattern the 2026-07-19 ruling names and forbids.

Two other candidate gaps from the same audit turned out **not** to be real gaps on closer inspection, and are intentionally NOT touched by this PR:
- `missing_bytes_reacquire` raw-authority obligations are already re-classified every 30s by `_converge_raw_authority_frontier` (part of the existing raw-materialization convergence loop) — no gap.
- `judgment_automation_enabled` (assertion-candidate auto-judgment, default off) is an unrelated subsystem from raw-authority frontier conflicts, and its default is an explicit write-authority security boundary, not something to flip as a side effect of this audit.

## Solution

- New `polylogue/daemon/blob_gc_periodic.py`: `periodic_blob_gc_check` runs one bounded batch (max 200 blobs) every 900s through the daemon write coordinator, structurally mirroring `periodic_embedding_orphan_reconcile_check` (same transient-lock retry handling, same `catch_up_complete` gating). No changes to `run_blob_gc_report`'s existing safety invariants (referenced-blob check, generation-age gate, bounded batch, planning outside the destructive lock) — this only adds a caller.
- `embedding-orphan-reconcile` is now permanently read-only: removed `--yes`, the `RebuildLease`/offline-exclusive apply branch, and the now-dead `dry_run=False` rendering branch.

## Verification

- `devtools test tests/unit/daemon/test_blob_gc_periodic.py` — 4 passed, including a `DaemonWriteCoordinator` production-route proof (bypassing `run_sync` makes the test fail).
- `devtools test tests/unit/cli/test_archive_maintenance_cli.py -k embedding_orphan` — 4 passed, including a new test proving `--yes` is now rejected as an unknown option.
- `devtools verify --quick` — pass.

## Summary by CodeRabbit

* **New Features**
  * Added periodic background garbage collection for unused blobs, including retry handling for temporary archive locks.
  * Added a single-pass blob cleanup operation that removes eligible, unreferenced files.

* **Bug Fixes**
  * Changed embedding-orphan reconciliation to read-only inspection mode.
  * Removed the `--yes` option and manual deletion capability from the reconciliation command.

* **Documentation**
  * Updated topology metadata and status counts to reflect the new maintenance component.


Sinity:
Addressed the Click-wording finding (now asserts `exit_code == 2` + option name in output, not Click's internal error text).

The other two don't hold up:

**"Do not await the synchronous coordinator API"** — this is incorrect for the class actually used here. `DaemonWriteCoordinator.run_sync` (`polylogue/daemon/write_coordinator.py:227`) is `async def`; the suggested non-async call would itself be the bug (returns a coroutine, never actually runs the GC pass). The suggestion appears to be confusing it with a different, similarly-named sync wrapper elsewhere in the same file (used by the HTTP thread bridge, not by tests). Confirmed empirically, not just by reading: the test as written passes (`4 passed`), and it's the exact same `coordinator = DaemonWriteCoordinator(); async def run(): return await coordinator.run_sync(...); asyncio.run(run())` pattern already used in `tests/unit/daemon/test_embedding_orphan_reconcile_daemon.py::test_daemon_coordinator_owns_real_orphan_reconcile_mutation`, which also passes today.

**"Use frozen_clock for the age-gate test"** — not applicable: `_backdate` sets a blob's mtime to a fixed epoch (0), it doesn't read the host clock at all. `frozen_clock` pins what `time.time()`/`datetime.now()` return to code under test; there's no such call here to pin. `devtools verify --quick`'s `verify-test-clock-hygiene` step (which does flag real host-clock reads via AST) already passes on this file.

PR #3291, opened 2026-07-27
feat(storage): add archive-location ownership acquisition for maintenance writers

Sinity:
## Summary

Adds `OwnedArchiveLocation`, a distinct ownership-proof capability that a maintenance/campaign writer must acquire over an `ArchiveLocation` before opening SQLite against its tiers.

## Problem

`ArchiveLocation`/`ArchiveIdentity` (polylogue/storage/archive_identity.py) already distinguish configured root, configured tier, and resolved active generation as separate typed values, with `assert_writable_archive_identity` as a preflight that fails before bootstrap on a durable-tier conflict. But per polylogue-ovme.1 AC1/AC5, none of this says *who* may write to a location: a maintenance/campaign writer (offline index rebuild, devtools benchmark/scale campaign, b5l generation transition) had no typed way to prove exclusive ownership before opening SQLite — an unowned or foreign path could previously only fail after a writer connection already existed, via some later runtime error, not before SQLite opens.

## Solution

- `OwnedArchiveLocation.acquire(location, owner_id=...)`: a pure filesystem preflight that takes an exclusive, non-blocking `flock` on a `.archive-ownership.lock` file under the location's configured root, and never opens a sqlite3 connection. A location already owned by another live process raises `ArchiveOwnershipError` immediately. A lock left behind by a now-dead process is reclaimed (mirrors `index_generation.py`'s existing stale-lease handling for offline rebuilds).
- `assert_owns_archive_location(owned, location)`: rejects reusing an ownership proof against a foreign configured root, and rejects a stale proof whose active generation has since rotated (e.g. a concurrent promotion) — mirroring `assert_writable_archive_identity`'s preflight shape one level up (that function proves one coherent generation; this proves one writer for that generation).
- `OwnedArchiveLocation` is a context manager (`release()` on `__exit__`).

Wiring real maintenance/campaign call sites (storage, status, devtools campaigns) to actually require `OwnedArchiveLocation` before writing is polylogue-ovme.2/.3 scope — not this slice, which is identity/resolver/canaries only (per the ovme parent epic's explicit 3-way split).

## Acceptance criteria (ovme.1)

1. Satisfied for the ownership axis: `OwnedArchiveLocation` is a distinct typed constructor; wrong root or stale generation fails via `assert_owns_archive_location`, and acquiring an already-owned location fails via `OwnedArchiveLocation.acquire` — both before any SQLite tier file is opened. Root/tier/generation typing (the rest of AC1) already existed in `ArchiveLocation`/`TierFileIdentity` prior to this PR.
2/3/4. Unaffected by this change (resolver, production read compatibility, existing split-tier/symlink canaries) — no existing behavior touched.
5. Satisfied — see Verification below.

## Verification

- `devtools test tests/unit/storage/test_archive_identity.py` -> `13 passed` (5 new: concurrent-acquire-before-sqlite-exists, dead-process lock reclaim, foreign-root rejection + accepted match, stale-generation rejection after pointer rotation)
- `devtools test tests/unit/storage/` -> 1 pre-existing unrelated failure (`test_exact_session_action_count_bounds_pairing_before_global_ranking`), reproduced identically on `origin/master` before this change (confirmed via `git stash`)
- `mypy --strict polylogue/storage/archive_identity.py tests/unit/storage/test_archive_identity.py` -> no issues found
- `ruff check` / `ruff format --check` on both files -> clean
- `devtools render all --check` -> no "out of sync" surfaces
- `git push` pre-push quick gate (format/lint/mypy/render/topology/layering/closure-matrix/schema/manifests/ci-workflows/doc-commands/docs-coverage/test-infra-currency/test-clock-hygiene/pytest-timeout-overrides/degrade-loudly) -> all `ok`, `exit_code: 0`

Ref polylogue-ovme.1

### DIFF
--- a/polylogue/storage/archive_identity.py
+++ b/polylogue/storage/archive_identity.py
@@ -7,6 +7,7 @@ import logging
 import os
 import socket
 import sys
+import time
 import uuid
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Sinity:
Self-review finding + fix (CodeRabbit was rate-limited on the initial push, so I reviewed this myself before merging): the original stale-lock reclaim path had a dual-inode race — two processes racing a reclaim at the instant the true owner died could each end up holding a lock on a *different* inode, both believing they own the archive location. Fixed in the second commit by retrying the same flock() call against the same inode instead of ever creating a competing one via rename. Strengthened the dead-process-reclaim test to assert no second lock-file artifact exists after acquisition.

PR #3296, opened 2026-07-27
fix(query): rank near:id: and materialize lineage:id: projection

Sinity:
## Summary

Fixes two pre-existing, execution-layer-only gaps in the generic `find`/`read`
CLI query route: `near:id:<ref>` silently fell back to an unfiltered session
list instead of real vector-similarity ranking, and `lineage:id:<ref>` never
materialized its declared `parent_refs`/`child_refs`/`continuation`
recursive-graph projection columns.

## Problem

`polylogue-z9gh.3`'s 2026-07-20 note narrowed the bead's remaining scope to
two "size S" residuals, confirmed via a spot-check against a seeded demo
archive: `near:id:` executes without error via the generic find/read CLI
route but does not perform real similarity ranking (falls back to an
unfiltered session list), and the declared recursive-page projection columns
are not materialized by that same route for `lineage:id:` queries.

I independently re-verified both against a live `polylogue demo seed`
archive before writing any fix:

- `polylogue find 'near:id:<seed>' --format json` returned the exact same
  19-session list, in the same order, as a query with no filter at all.
- `polylogue find 'lineage:id:<parent>' --format json` correctly narrowed to
  the 2-session lineage family (the SQL-level filter already worked), but
  neither row carried `parent_refs`, `child_refs`, or `continuation`.

Root cause for `near:id:`: `polylogue/cli/archive_query.py`'s hand-rolled
search-execution path (`_query_hits`, `_daemon_session_page_supported`,
`_try_emit_daemon_session_page`) only ever threaded `similar_text` through;
`SessionQuerySpec.similar_session_id` was never read anywhere in that file,
so a bare `near:id:` predicate (no FTS terms, no `similar_text`) fell through
every branch straight to the plain `archive.list_summaries(**filter_kwargs)`
call. (The separate `SessionFilter`/`archive_execution.py` route already had
correct vector-seeded ranking since PR #3018 — this bug was isolated to the
CLI's own parallel execution path.)

Root cause for `lineage:id:`: the SQL-level `QueryLineagePredicate` filter
(via `boolean_predicate`) was correct, but no code anywhere translated the
declared `RECURSIVE_COLUMNS` (`session_id`, `parent_refs`, `child_refs`,
`continuation`) from `archive/query/discovery.py` into actual payload fields.

## Solution

- `polylogue/cli/archive_query.py`: thread `similar_session_id` through
  `_query_hits` the same way `similar_text` already is. When set, it calls
  the existing `VectorProvider.query_by_session` mechanism (the same
  mechanism `archive_execution.py`'s `_session_seed_scored` already uses for
  the `SessionFilter` route) rather than building new vector-search code.
  Raises a typed `click.UsageError` when no vector backend is configured or
  the seed session has no stored embeddings, matching the existing
  `near:"text"` error contract. Also excludes `similar_session_id` from the
  daemon session-page fast path (`_daemon_session_page_supported`), since
  that fast path's `/api/sessions` split-archive route has the identical gap
  and forcing local execution is the minimal fix within this file's scope.
- `polylogue/storage/sqlite/archive_tiers/archive.py`: adds
  `ArchiveStore.session_lineage_edges`, a small bounded query over the
  existing `sessions.parent_session_id` column (the same column
  `lineage:id:` already filters on via `root_session_id`) returning direct
  one-hop parent/child ids for a page of session ids — no new recursive
  graph traversal.
- `polylogue/cli/archive_query.py` + `polylogue/surfaces/payloads.py`: when a
  list page's `boolean_predicate` carries a `QueryLineagePredicate`
  (`lineage:id:`), look up per-page lineage edges and populate
  `SessionListRowPayload.parent_refs`/`child_refs`/`continuation`.
  `continuation` mirrors the page's own pagination cursor; edges are
  qualified per page, consistent with the "recursive-page" result-semantics
  contract already documented in `archive/query/discovery.py`
  (`RESULT_SEMANTICS_TEACHING`).
- Regenerated `docs/schemas/cli-output/session-list-row.schema.json`,
  `docs/openapi/search.yaml`, and `webui/src/api/generated.ts` for the two
  new payload fields.

No grammar changes: `polylogue/archive/query/expression.py` is untouched.

## Verification

- Manual, against a seeded demo archive: `near:id:<ref>` now raises a typed
  error instead of returning the full session list; `lineage:id:<ref>` now
  emits `parent_refs`/`child_refs` per row.
- `devtools test tests/unit/cli/test_query_exec_laws.py` — 88 passed (adds 3
  new tests: real vector-ranked `near:id:` through the CLI route with an
  assertion that `list_summaries` is never reached, a typed `UsageError`
  with no vector backend configured, and a real-archive `lineage:id:` test
  via `SessionBuilder`/`write_parsed_session_to_archive` proving
  `parent_refs`/`child_refs` materialize correctly against production
  `ArchiveStore` SQL, not a mock).
- `devtools test tests/unit/cli/test_archive_query.py tests/unit/cli/test_cli_output_schemas.py` — 135 passed.
- `devtools test tests/unit/archive/test_near_session_seed_execution.py tests/unit/storage/test_vec_session_seed.py` — 9 passed.
- `devtools test tests/unit/mcp/test_query_request_contracts.py tests/unit/cli/test_query_expression.py tests/unit/cli/test_select.py` — 449 passed, 1 skipped.
- `mypy` on all three changed source files — no issues.
- `ruff check` + `ruff format --check` — clean.
- `devtools render all --check` — sync OK, no "out of sync" lines.
- `devtools verify --quick` — exit 0 (also ran automatically via the pre-push hook).

Not run: `devtools verify --all`/full suite — seeding `pytest-testmon` from
scratch triggered a large, apparently pre-existing failure spread across the
whole suite baseline unrelated to these three files; the targeted
`devtools test` runs above cover every file this change touches or could
plausibly affect.

Ref polylogue-z9gh.3

### DIFF
--- a/docs/openapi/search.yaml
+++ b/docs/openapi/search.yaml
@@ -1578,6 +1578,28 @@ components:
           - $ref: '#/components/schemas/SessionFlagsPayload'
           - type: 'null'
           default: null
+        parent_refs:
+          anyOf:
+          - items:
+              type: string
+            type: array
+          - type: 'null'
+          default: null
+          title: Parent Refs
+        child_refs:
+          anyOf:
+          - items:
+              type: string
+            type: array
+          - type: 'null'
+          default: null
+          title: Child Refs
+        continuation:
+          anyOf:
+          - type: string
+          - type: 'null'
+          default: null
+          title: Continuation
       required:
       - id
       - origin
--- a/docs/schemas/cli-output/session-list-row.schema.json
+++ b/docs/schemas/cli-output/session-list-row.schema.json
@@ -183,6 +183,33 @@
       "default": null,
       "title": "Anchor"
     },
+    "child_refs": {
+      "anyOf": [
+        {
+          "items": {
+            "type": "string"
+          },
+          "type": "array"
+        },
+        {
+          "type": "null"
+        }
+      ],
+      "default": null,
+      "title": "Child Refs"
+    },
+    "continuation": {
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "null"
+        }
+      ],
+      "default": null,
+      "title": "Continuation"
+    },
     "created_at": {
       "anyOf": [
         {
@@ -230,6 +257,21 @@
     "origin": {
       "title": "Origin",
       "type": "string"
+    },
+    "parent_refs": {
+      "anyOf": [
+        {
+          "items": {
+            "type": "string"
+          },
+          "type": "array"
+        },
+        {
+          "type": "null"
+        }
+      ],
+      "default": null,
+      "title": "Parent Refs"
     },
     "repo": {
       "anyOf": [
--- a/polylogue/cli/archive_query.py
+++ b/polylogue/cli/archive_query.py
@@ -839,12 +839,19 @@ def _lineage_seed_from_predicate(predicate: QueryPredicate | None) -> str | None
     session rows to one shared-root lineage family. This walks the same
     boolean-predicate tree to detect that shape so the list route can
     materialize the declared recursive-graph projection columns for it (#z9gh.3).
+
+    Only descends into ``and`` nodes. A ``lineage:id:X or repo:foo`` result
+    set is NOT purely lineage X's family -- rows matched only via the ``or``
+    branch would get X's parent_refs/child_refs/continuation stamped on them,
+    which is wrong, not just imprecise. An ``or`` node (or a ``not`` wrapping
+    the predicate, which isn't a ``QueryBoolPredicate``/``QueryLineagePredicate``
+    at all) correctly yields no seed here.
     """
     if predicate is None:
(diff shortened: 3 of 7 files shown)


Sinity:
Self-review finding + fix (CodeRabbit rate-limited on this PR's pushes, so I reviewed it myself): \`_lineage_seed_from_predicate\` recursed into every \`QueryBoolPredicate\` child regardless of the boolean operator. A query like \`lineage:id:X or repo:foo\` would still detect X as the seed and stamp X's parent_refs/child_refs onto every row in the result — including rows that only matched via the unrelated \`or\` branch and were never part of X's lineage family at all. Fixed by only recursing on \`op == "and"\` (AND only narrows further, so it's still correct there). Added a direct regression test against synthetic and/or-combined predicate trees.

PR #3297, opened 2026-07-27
feat(daemon): surface GIL/free-threaded runtime mode in status

Sinity:
## Summary

Adds a `gil_enabled` field to the daemon status payload so operators can see whether `polylogued` is running under a standard GIL-enabled interpreter or a free-threaded (PEP 703, Python 3.13+) build such as 3.14t.

## Problem

`polylogue-dcz5` ("Deploy polylogued on Python 3.14t (free-threaded) and gate runtime") requires the daemon status surface to show the interpreter's runtime mode before a production deploy decision can be verified end-to-end. No status/observability field existed anywhere for this.

## Solution

- `polylogue/daemon/status.py`: new `_gil_enabled()` helper — calls `sys._is_gil_enabled()` when present (Python 3.13+), falls back to `True` on older interpreters that never had a disableable GIL. New `DaemonStatus.gil_enabled: bool = True` field, wired into `build_daemon_status()`'s return.
- `polylogue/daemon/status_snapshot.py`: `_minimal_status_payload()` (the pre-refresh/refreshing fallback dict) also reports `gil_enabled` — it's a cheap, no-I/O fact matching that function's cost profile, reusing the module's existing deferred-import pattern from `polylogue.daemon.status`.
- `polylogue/cli/commands/status.py`: `_show_daemon_status()` renders a `Runtime: ...` line; `gil_enabled` added to `_compact_status_payload()`'s key allowlist so it survives `polylogue ops status --json` without `--full`.
- Daemon HTTP `/api/status` reads from the same snapshot builders, so no separate wiring was needed there. Grepped for MCP consumers of `DaemonStatus` — none found, so no MCP surface change needed.

## Verification

- `mypy --strict polylogue/daemon/status.py polylogue/daemon/status_snapshot.py polylogue/cli/commands/status.py` → clean.
- `ruff check` + `ruff format --check` on touched files → clean.
- `devtools test tests/unit/daemon/test_daemon_status.py` → 61 passed, including 4 new tests: GIL-disabled (free-threaded), GIL-enabled (standard), `sys._is_gil_enabled` attribute-absent fallback, and minimal-snapshot payload inclusion.
- `devtools test tests/unit/cli/test_status.py` → 1 pre-existing failure (`test_archive_facade_route_catalog_covers_public_async_facade`, a `hermes_integration_health` route-catalog mismatch), confirmed identical with this diff stashed out — unrelated to this change.
- `devtools render all --check` → no drift.
- Pre-push hook (`devtools verify --quick`: format + lint + mypy + render-all-check) → exit 0.

## Self-review note

CodeRabbit is rate-limited on every PR this session, so I self-reviewed the diff directly: 4 files touched, matches the plan exactly, no stray edits, no unrelated `.beads/*` churn included (those pre-existing dirty files were stashed separately and are not part of this branch).

## Scope

This covers AC item 2 of 4 on `polylogue-dcz5` ("daemon status surface shows free-threaded runtime"). Explicitly out of scope for this repo/PR: the sinnix-side package pin bump + `switch`/`test-vm`, recorded 7mtf-style benchmark numbers proving parallel fan-out, and rollback documentation — all real deploy coordination outside this repo.

Ref polylogue-dcz5

## Summary by CodeRabbit

* **New Features**
  * Daemon status now displays whether the runtime uses the GIL or free-threaded execution.
  * Compact status output includes the runtime mode for machine-readable monitoring.
  * Status reporting remains compatible with standard Python runtimes when detailed runtime detection is unavailable.

* **Tests**
  * Added coverage for free-threaded, standard, and fallback runtime detection scenarios.

### DIFF
--- a/polylogue/cli/commands/status.py
+++ b/polylogue/cli/commands/status.py
@@ -1482,6 +1482,13 @@ def _show_daemon_status(env: AppEnv, status: dict[str, Any], *, compact: bool = 
                 desc = comp.get("description", "")
                 env.ui.console.print(f"  [{state_color}]●[/{state_color}] {name}: {desc}")
 
+    # Runtime (polylogue-dcz5): free-threaded (3.14t+) vs GIL-enabled.
+    gil_enabled = status.get("gil_enabled")
+    if gil_enabled is not None:
+        mode_text = "free-threaded (GIL disabled)" if gil_enabled is False else "GIL enabled"
+        mode_color = "green" if gil_enabled is False else "dim"
+        env.ui.console.print(f"  Runtime: [{mode_color}]{mode_text}[/{mode_color}]")
+
     # Live ingest. The status payload carries LiveIngestAttemptSummary, which
     # exposes running_count + per-attempt worker progress in `recent` — it has
     # no top-level completed_count/total_count, so the previous keys always read
@@ -1578,6 +1585,7 @@ def _compact_status_payload(status: dict[str, Any], *, source: str) -> dict[str,
         "messages",
         "raw_records",
         "next_action",
+        "gil_enabled",
     ):
         if key in status:
             payload[key] = status[key]
--- a/polylogue/daemon/status.py
+++ b/polylogue/daemon/status.py
@@ -7,6 +7,7 @@ import json
 import os
 import re
 import sqlite3
+import sys
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any, Literal, cast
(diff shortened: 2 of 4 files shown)


Sinity:
Checked CodeRabbit's finding about `_SNAPSHOT` pollution across tests — it's a false positive. `tests/conftest.py` already has an autouse fixture (lines ~406-480) that calls `reset_status_snapshot()` after every test specifically to guard against this. Verified empirically: ran `test_status_snapshot_serves_cached_payload_without_rebuilding_status` immediately followed by `test_minimal_status_payload_reports_gil_enabled` in a single process with randomization disabled (the exact order CodeRabbit's concern depends on) — passes identically whether or not the new test resets `_SNAPSHOT` itself. Not merging a redundant defensive change that would misleadingly imply a vulnerability that the existing fixture already closes.

PR #3299, opened 2026-07-27
feat(usage): reconcile per-session token/cost evidence for one case

Sinity:
## Summary

First slice of polylogue-f2qv.6 ("Reconcile profiles and costs to exact
provider usage"): adds `build_session_usage_reconciliation()` and
`SessionUsageReconciliation` to `polylogue/storage/usage.py`, reconciling
one session's three disagreeing usage/cost sources into a single canonical
EvidenceValue-backed snapshot.

## Problem

polylogue-f2qv.6 reports that a single live Codex session can show three
incompatible answers:

- `session_model_usage` (exact rollup): 64,561 uncached input + 723,456
  cache read + 7,776 output tokens (795,793 total)
- `session_profiles`: a stale 4,031-token estimate
- the per-session cost insight: zero / unavailable

The cost insight reads its number off the `session_profiles` row
(`_session_cost_insight_from_archive_row` in
`polylogue/storage/sqlite/archive_tiers/archive.py`), so a stale profile
silently produces a stale-and-wrong cost too. Across the operator's full
Codex census, zero profiles matched their session's exact rollup at the
time this bead was filed.

## Solution

- Two new `FactFamilySpec` declarations, `SESSION_USAGE_RECONCILED_TOKENS_FAMILY`
  and `SESSION_USAGE_RECONCILED_COST_FAMILY`, at session grain, following
  the same EvidenceValue pattern this module already uses for archive-level
  pricing-lane evidence (`USAGE_LANE_EXACT_TOKENS_FAMILY` /
  `USAGE_LANE_CATALOG_COST_FAMILY`, landed under polylogue-cuxz.2).
- `build_session_usage_reconciliation(...)` is a pure function over
  already-loaded values (no storage reads). It builds an `EvidenceValue`
  for the `session_model_usage` rollup (authority `provider-reported`) and
  one for the `session_profiles` estimate (authority `structural` or
  `model-derived`, driven by the profile's own `cost_provenance` label),
  then reconciles them with `refine_evidence_value` -- the shared
  cuxz.2 primitive that picks the strongest-authority value on
  disagreement (never an average, never insertion-order) while keeping
  every input as a labeled `contributions` entry.
- The money lane is reconciled independently, per the bead's own design
  note ("disjoint token-lane authority separate from monetary price
  authority"): a fresh catalog reprice of the *winning* token evidence
  (authority `catalog-derived`) is reconciled against the legacy persisted
  cost (authority derived the same way as the token side).
- `SessionUsageReconciliation.superseded_token_observations()` /
  `superseded_cost_observations()` expose exactly which lower-authority
  inputs lost and why, rather than burying them.

## Scope (first slice, not the full bead)

This satisfies **one** reconciliation case with a real test, not the bead's
full acceptance criteria. Explicitly **not** done here (see the bead note
added alongside this PR):

- No storage-layer wiring: nothing in `storage/insights/session/rebuild.py`,
  `storage/sqlite/archive_tiers/archive.py`, or the insight registry calls
  this function yet -- `session_model_usage`, `session_profiles`, and the
  cost insight still disagree in the live archive today.
- No daemon convergence / contradiction-debt integration.
- No corpus-wide census proving "zero unexplained profile contradictions."
- Broader EvidenceValue family/surface migration remains polylogue-cuxz.3.

## Verification

```
devtools test polylogue/storage/usage.py \
  tests/unit/storage/test_session_usage_reconciliation.py \
  tests/unit/storage/test_provider_usage_report.py \
  tests/unit/storage/test_session_profile_model_usage_consistency.py
# -> 25 passed

mypy --strict polylogue/storage/usage.py tests/unit/storage/test_session_usage_reconciliation.py
# -> Success: no issues found in 2 source files

devtools verify --quick
# -> exit_code: 0
```

Ref polylogue-f2qv.6

### DIFF
--- a/polylogue/storage/usage.py
+++ b/polylogue/storage/usage.py
@@ -10,12 +10,13 @@ from __future__ import annotations
 import json
 import sqlite3
 from collections import Counter, defaultdict
-from collections.abc import Iterable
+from collections.abc import Iterable, Sequence
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Final, Literal, cast
 
+from polylogue.archive.semantic.cost_records import ModelUsageTotals
 from polylogue.archive.semantic.pricing import (
     CATALOG_EFFECTIVE_DATE,
     CATALOG_PROVENANCE,
@@ -32,11 +33,14 @@ from polylogue.archive.semantic.subscription_pricing import (
 from polylogue.core.enums import Origin, Provider
 from polylogue.core.evidence_value import (
     CoverageExclusion,
+    EvidenceObservation,
     EvidenceValue,
     FactFamilySpec,
     FrameCoverage,
     FreshnessProvenance,
+    MeasurementAuthority,
     TemporalProvenance,
+    refine_evidence_value,
     sum_evidence_values,
 )
 from polylogue.core.refs import ObjectRef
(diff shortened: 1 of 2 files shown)


Sinity:
Self-review finding + fix (CodeRabbit rate-limited): \`_catalog_cost_evidence\` priced the ENTIRE reconciled token total (input+output+cache_read+cache_write summed into one int) as if every token were an input token. Input/output/cache tokens have very different catalog rates - cache reads especially are typically priced far below input. Verified on the bead's own reproduction numbers (64,561 input + 723,456 cache read + 7,776 output): the bug computed **$0.994741** vs the correct **$0.248893** - a ~4x overstatement, since 723K cache-read tokens got priced at full input rate instead of their real discounted rate.

Fixed by having \`build_session_usage_reconciliation\` determine which source won the token reconciliation and pass THAT source's real per-category breakdown into \`_catalog_cost_evidence\`, which now calls \`estimate_cost\` with each category separately. Strengthened the existing cost test to assert the exact expected value (with a guard proving the fixture's model has genuinely non-uniform rates, so the assertion can't pass by coincidence) - confirmed it fails without the fix.

PR #3317, opened 2026-07-27
fix: add aggregate zip-size cap to import --explain preview

Sinity:
## Summary
Brings the `import --explain` dry-run zip preview into agreement with the real `process_zip` decode path on aggregate uncompressed-size rejections.

## Problem
PR #3314 added `MAX_AGGREGATE_UNCOMPRESSED_SIZE` (64 GiB) to `ZipEntryValidator.filter_entries` in `polylogue/sources/decoder_zip.py`, tracking a running total of declared uncompressed sizes across all entries in a zip and rejecting entries whose admission would push that total over the cap — closing a zip-bomb-by-many-entries gap that the pre-existing per-entry ratio/size cap didn't cover.

`polylogue/sources/import_explain.py`'s `_zip_entry_skip_reason` (the `import --explain` dry-run preview path) duplicates the old per-entry ratio/size checks but was never updated with the aggregate check. This is a preview/reality drift, not a security hole (the real decode path is already protected): the preview could report an entry "will import" for a case a real `import` run now rejects on aggregate grounds.

## Solution
- `import_explain.py` now imports `MAX_AGGREGATE_UNCOMPRESSED_SIZE` from `decoder_zip.py` (no redefinition).
- `_zip_entry_skip_reason` takes/returns a running `aggregate_total`, threaded through `_explain_zip`'s entry loop in the same iteration order the real `process_zip`/`ZipEntryValidator.filter_entries` path uses.
- Check order mirrors the real validator exactly: extension/ratio/per-entry-size checks first — entries that fail those never contribute to the running total (matching `ZipEntryValidator`) — then the aggregate check against central-directory metadata alone, before any entry bytes are read.

## Verification
- `devtools test tests/unit/cli/test_import_explain.py tests/unit/sources/test_decoders.py` → 34 passed
- `mypy` (via pre-push quick gate) → no issues
- `ruff check` / `ruff format --check` (via pre-push quick gate) → clean
- New tests in `tests/unit/cli/test_import_explain.py`:
  - `test_import_explain_zip_rejects_aggregate_over_cap_before_read` — builds a zip with 3 entries each individually under the per-entry cap whose sum exceeds a monkeypatched aggregate cap, confirms the preview reports the correct entries as skipped for the aggregate reason, and cross-checks the accepted/rejected split against a real `ZipEntryValidator.filter_entries` run over the same entries to prove they agree exactly.
  - `test_import_explain_zip_allows_archive_comfortably_under_aggregate_cap` — a small archive well under the cap reports no aggregate-cap skips (regression guard).

Ref polylogue-it3u

## Summary by CodeRabbit

* **Bug Fixes**
  * ZIP import previews now enforce the archive-wide uncompressed size limit.
  * Entries exceeding the combined size limit are clearly identified as skipped.
  * ZIP entry filtering now more closely matches import decoding behavior.

* **Tests**
  * Added coverage for archives exceeding the aggregate size limit.
  * Added coverage confirming valid small archives remain unaffected.

### DIFF
--- a/polylogue/sources/import_explain.py
+++ b/polylogue/sources/import_explain.py
@@ -552,7 +552,12 @@ def _explain_zip(
     try:
         with zipfile.ZipFile(path) as archive:
             for info in archive.infolist():
-                skip_reason, aggregate_total = _zip_entry_skip_reason(info, aggregate_total=aggregate_total)
+                skip_reason, aggregate_total = _zip_entry_skip_reason(
+                    info,
+                    aggregate_total=aggregate_total,
+                    zip_path=path,
+                    provider_hint=provider_hint,
+                )
                 if skip_reason is not None:
                     skipped.append(
                         ImportSkippedRowPayload(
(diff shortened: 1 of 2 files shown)


Sinity:
Fixed in 65e41ea34 — threaded `zip_path`/`provider_hint` into `_zip_entry_skip_reason` and applied the identical `classify_artifact_path`/`parse_as_session` exclusion `ZipEntryValidator.filter_entries` uses, in the same order, before advancing the running aggregate total.

While verifying this, found a deeper pre-existing issue: every current `OriginArtifactRule.path_pattern` requires a slash-anchored match (`(?:^|/)workflows/...`), but zip-embedded entry paths are colon-joined (`f"{zip_path}:{name}"`), so `classify_artifact_path` appears to never actually return a non-session classification for any zip-embedded entry today — the `session_only` exclusion looks like dead code in the real decoder too, not just something this preview fix needed to catch up to. Filed as `polylogue-dc1k` rather than silently noting it; the new test here isolates and proves this fix's logic via a monkeypatched classification, independent of that separate gap.

PR #3354, opened 2026-07-27
fix(sources): defer stale full-revision retirement instead of quarantining new raw

Sinity:
## Summary

Fixes a permanent-quarantine bug found while investigating live `polylogue ops debt list` output: a stale full-revision sibling's retirement guard tripping (a legitimate, transient invariant check) was propagating up and permanently poisoning an unrelated, healthy raw currently being ingested.

## Problem

The live archive's `polylogue ops debt list` surfaced `debt:raw-materialization:claude-code-session:parse-failed`, including 11 of 73 raws carrying `parse_error = "RuntimeError: an active byte-revision chain cannot move to membership governance"` (raised in `ArchiveStore.replace_raw_membership_census`, `storage/sqlite/archive_tiers/archive.py` ~line 2653).

Root cause: `_apply_membership_sessions` (`sources/live/batch.py`) iterates `archive.convertible_full_revision_raw_ids(logical_source_key)` — every raw sharing a session's logical identity with `revision_kind='full'` — and tries to retire each into membership governance as a side effect of processing a brand new, otherwise-unrelated raw (`source_raw_id`). That coarse query doesn't check whether a candidate still has an active byte-revision-chain dependent (another raw's `predecessor_raw_id`/`baseline_raw_id` pointing at it); that invariant is enforced deeper, inside `replace_raw_membership_census`, which raises `RuntimeError` when it finds one.

Before this fix, that `RuntimeError` propagated out of the retire-siblings loop into the caller's blanket `except Exception` (`sources/live/batch.py`), which called `archive.mark_raw_parse_failed` on `source_raw_id` — the CURRENT raw being ingested — permanently quarantining healthy, unrelated content because of a stale sibling's bookkeeping conflict. Since `validation_status` never gets reached in this failure path, the row is stuck (shown as `unknown` in the debt summary) forever with no retry mechanism resolving it.

This is the same class of incremental-discovery-ordering race documented and partly fixed for polylogue-52l2/polylogue-hm2f (byte-revision cohort acceptance), but on the retire-sibling leg rather than the accept-cohort leg, and not covered by either fix.

## Solution

Wrap the `archive.replace_raw_membership_census(..., retire_full_revision_governance=True)` call in `_apply_membership_sessions` in a narrow `try/except RuntimeError` that logs a warning and defers (continues to the next candidate) instead of letting the failure propagate and poison the current record's own write. The sibling's own census write already rolled back in its own transaction (each `replace_raw_membership_census` call manages its own transaction), so nothing is left inconsistent; a later daemon tick can retry the retirement once the dependent chain has resolved.

A second, unrelated commit (`docs: regenerate README docs-surface section`) regenerates a pre-existing generated-surface drift in README.md that was blocking the pre-push verification gate on unmodified `origin/master` — verified via `git stash` that the drift predates this branch.

## Verification

- `devtools test tests/unit/sources/test_live_batch_support.py` — 72 passed, including new regression test `test_full_revision_retirement_conflict_defers_without_quarantining_current_raw`, which mocks `replace_raw_membership_census` to raise the exact production `RuntimeError` for a retire call and asserts the current raw's own membership census still gets written and no exception propagates.
- **Anti-vacuity**: with the `try/except` reverted (`git stash` on `polylogue/sources/live/batch.py`), the same test fails with the `RuntimeError` propagating and the assertion on the current raw's persisted membership census failing — confirmed live.
- `devtools verify --quick` — green (format, lint, mypy --strict, render all --check, topology, layering, closure-matrix, schema roundtrip, manifests, ci-workflows, doc-commands, docs-coverage, test-infra-currency, test-clock-hygiene, pytest-timeout-overrides, degrade-loudly, hash-boundary-census).

## Scope notes

Investigated but intentionally NOT touched in this PR (documented in polylogue-lpen):

- 59 of the same 73 claude-code-session parse-failed raws: `"captured JSONL payload ends before a complete record boundary"` for now-inert `.pre-enrich` snapshot files that will never be appended to again — genuinely stuck source data, not a code bug.
- 25 of 26 codex-session parse-failed raws: `"raw revision CAS rejected an older/conflicting accepted frontier"` — these look like the no-shrink CAS invariant correctly rejecting superseded duplicate captures of an actively-growing session file acquired at multiple points in time (working as designed, per polylogue-yla8's no-shrink invariant).
- `hermes-session` (2 raws) and `unknown-export` (22 raws) "produced no materializable sessions" failures — these are recognizable non-session sidecar/manifest files (Hermes skill templates, Claude/ChatGPT export-zip `projects.json`/`export_manifest.json`), but `archive/raw_materialization.py::parsed_non_session_artifact_reason` has no branch for those origins, AND even if it did, `_raw_materialization_category` only consults it for already-parsed rows — rows that error via `parse_error` in `ingest_worker.py::_materialize_parsed_sessions` never reach it. Fixing this needs a broader, multi-origin ingest WRITE-path change, not a narrow patch.
- `debt:convergence:fts:fts_surface:messages_fts` — traced to the daemon's periodic convergence-debt-retry loop being gated behind `catch_up_complete`, which the watcher doesn't signal until its ENTIRE initial catch-up backlog finishes (observed live: a single catch-up chunk took ~2.6 hours against the current backlog size). Working as designed, not a bug, though worth a design discussion about whether FTS-debt retry should be gated behind the full catch-up.

Ref polylogue-lpen (filed with full investigation detail for all 5 debt categories from the original ask).

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ normal path that sends archive text outside the machine. Run
 
 Live site: <https://sinity.github.io/polylogue/>.
 
-Start with the task-oriented guides below; [docs/README.md](docs/README.md) separates guides, reference, internals, operations, evidence, design, and historical records. Current sequencing and active workstreams live in the Beads backlog (`bd ready`, `bd list --status open`).
+Start with the task-oriented guides below. The complete documentation map is in [docs/README.md](docs/README.md).
 
 | Document | Description |
 |----------|-------------|
--- a/polylogue/sources/live/batch.py
+++ b/polylogue/sources/live/batch.py
@@ -2177,14 +2177,44 @@ class LiveBatchProcessor:
                     raise RuntimeError(
                         f"full revision {revision_raw_id}:{logical_source_key} no longer parses uniquely"
                     )
-                archive.replace_raw_membership_census(
-                    revision_raw_id,
-                    retained_sessions,
-                    parser_fingerprint=self._current_parser_fingerprint(),
-                    censused_at_ms=acquired_at_ms,
-                    detail=HISTORICAL_NON_PREFIX_GOVERNANCE_DETAIL,
-                    retire_full_revision_governance=True,
-                )
+                try:
+                    archive.replace_raw_membership_census(
+                        revision_raw_id,
+                        retained_sessions,
+                        parser_fingerprint=self._current_parser_fingerprint(),
+                        censused_at_ms=acquired_at_ms,
+                        detail=HISTORICAL_NON_PREFIX_GOVERNANCE_DETAIL,
+                        retire_full_revision_governance=True,
+                    )
+                except RuntimeError as exc:
+                    # polylogue-lpen: retiring a stale full-revision sibling
+                    # to membership governance is best-effort cleanup, not a
+                    # precondition for accepting THIS record's own content.
+                    # `replace_raw_membership_census`'s own guards (archive.py
+                    # ``retire_full_revision_governance``) can legitimately
+                    # refuse a sibling that still has an active byte-revision
+                    # chain (another raw's predecessor/baseline still points
+                    # at it) or isn't self-contained-full at this exact
+                    # instant -- the same class of incremental-discovery
+                    # ordering race documented for polylogue-52l2/hm2f, just
+                    # on the retire-sibling leg instead of the accept-cohort
+                    # leg. Before this guard, that failure propagated through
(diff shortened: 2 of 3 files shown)


Sinity:
Superseded by PR #3357 (merged) — a different agent independently fixed the same polylogue-lpen bug in parallel with a better-typed exception (ActiveByteRevisionChainError vs bare RuntimeError). Closing without merge.

PR #3386, opened 2026-07-28
refactor(paths): migrate sibling_index_db call sites to ArchiveLocation

Sinity:
## Summary

Second resolver fully migrated for polylogue-l2cd (following resolve_active_index_db_path in #3385): sibling_index_db() (16 call sites, 14 files) is fully retired.

## Solution

Every production call site now resolves through ArchiveLocation.resolve(root).active_index_path instead of the literal sibling-derivation-from-anchor-parent anti-pattern, or -- where the anchor was provably already the resolved active index path (several daemon/status.py, provenance.py, similarity.py sites where the caller's dbf/path always names "index.db" already, since resolve_active_index_path raises otherwise) -- the now-dead derivation was simply dropped in favor of the value itself.

Two call sites (embed.py, embeddings/preflight.py) preserved a genuine two-candidate fallback contract (db_path's own directory first, then the globally configured archive root) rather than collapsing to one, since db_path can be a caller override distinct from the configured root -- and improved it to try each candidate's real connection rather than just the first existing path.

Function deleted from polylogue/paths/_roots.py and its export from polylogue/paths/__init__.py. devtools/verify_archive_resolver_completeness.py's baseline updated: sibling_index_db removed entirely (0 remaining); active_index_db_path and archive_file_set_root_for_paths remain, both still fully baselined.

## Verification

- mypy --strict polylogue/: clean (1091 files)
- ruff check/format --check: clean
- devtools test across all affected daemon/cli/storage/devtools test files: 2163 passed, 6 failed -- all 6 confirmed pre-existing/flaky by reproducing identically against the unmodified base commit (git stash A/B): a paths-export drift from an unrelated pre-existing symbol (browser_capture_pairing_state_path), a schema_drift health-tier doc gap, an asyncio task-cancellation flake (passes in isolation), and two facade/route-contract gaps already present on master
- python3 devtools/verify_archive_resolver_completeness.py: 0 unbaselined call sites
- devtools verify --quick: exit 0 (after regenerating docs/devtools.md for the command_catalog.py description change)

Ref polylogue-l2cd

## Summary by CodeRabbit

* **Bug Fixes**
  * Archive and status tools now consistently locate the active index database, improving behavior across migrations and archive layouts.
  * Database inspection, metrics, provenance, embedding, and cleanup workflows now follow the active archive index reliably.
  * Archive enrichment safely skips processing when the active index is unavailable.

* **Refactor**
  * Removed the legacy sibling-index path resolution approach and updated public path exports accordingly.

* **Documentation**
  * Updated archive resolver migration guidance and completeness reporting to reflect current coverage.


Sinity:
Triaged all 6 findings:

**Fixed** (real bugs, addressed in the follow-up commit):
- `embed.py`/`metrics.py` P1/P2 tier-mismatch: fixed by threading the resolved `ArchiveLocation`/configured root through explicitly instead of deriving embeddings/ops/source paths from the (possibly-external-generation) active index path.
- `blob_publication.py` resource leak: moved all pre-try setup into the protected block.
- `embeddings/preflight.py` candidate validation: now validates each candidate via the existing `_is_archive_index` helper before accepting it, so a stale/wrong index no longer wins unconditionally.

**Investigated, false positive** (not changed):
- The two "keep archive resolution out of the CLI adapter" / "keep surface packages behind a public API" findings. Checked `docs/plans/layering.yaml` directly: the enforced rule for `polylogue/cli` reads "imports substrate/insights as needed; no inward dependency is permitted" -- a one-way constraint (substrate must not import surfaces back), not a ban on surfaces importing substrate. `devtools verify --quick`'s layering check passes clean on this exact import both before and after this PR.

PR #3393, opened 2026-07-30
fix(storage): anchor the index pointer on index.db, not the promoted generation

Sinity:
## Summary

`IndexGenerationStore` poisoned its own canonical index pointer on first use, so any archive that had been promoted at least once could never be opened for an offline index rebuild again. This is the fix that unblocked the 2026-07-30 full rebuild of the live archive; it was verified in production but never made it into a PR, so master still carries the bug.

## Problem

Opening the live archive for a rebuild failed with:

```
invalid canonical index pointer anchor:
/realm/db/polylogue/.index-generations/gen-1784807190100-34534407/index.db
```

The canonical pointer is the path `promote()` replaces with a symlink — the archive's own `index.db` — never the generation that symlink currently targets. `IndexGenerationStore.__init__` followed the symlink, took the resolved generation path as `active_pointer`, wrote *that* into `.index-active-pointer`, and then rejected exactly that value on the next construction. It also derived `generations_root` as the nested `.index-generations/gen-*/.index-generations`.

So the store wrote a value it would later refuse, and the failure was permanent rather than transient: every subsequent run read the poisoned anchor and raised.

## Solution

`polylogue/storage/index_generation.py`: recompute and rewrite the anchor when it is absent **or poisoned**, instead of hard-failing on a poisoned value. The canonical pointer resolves from the configured `index` tier path rather than from the followed symlink. A symlink is still followed when it leaves the archive root, because an archive root that is a symlink farm pointing at the real location genuinely does have its canonical pointer elsewhere — that case is distinguished by checking for `.index-generations` in the resolved parts.

This is a recovery path, not a migration: an archive already carrying a poisoned anchor heals on next open.

## Verification

```
devtools test tests/unit/storage/test_index_pointer_anchor.py
4 passed in 4.88s
```

`tests/unit/storage/test_index_pointer_anchor.py` is new and covers both directions: a poisoned generation-nested anchor is recomputed rather than rejected, and an out-of-archive symlink target is still honored. Beyond the unit tests, this change is what let the real 41,363-raw / 99 GB rebuild open the live archive and run to promotion.

Also carries one `chore(beads)` commit filing `polylogue-bu1i` (aistudio-drive ambiguous cohorts are misclassified non-branches) and a correction note on `polylogue-7ilr`. Bead bookkeeping is folded into a code branch deliberately, per this repo's convention, rather than opened as a sibling branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when index pointer metadata becomes invalid after a generation promotion.
  * Repeatedly opening promoted archives now works reliably without errors.
  * Automatically repairs incorrect pointer metadata during archive access.
  * Preserves correct index resolution when archives use symbolic links, including links outside the archive root.

* **Tests**
  * Added coverage for promoted archives, pointer healing, repeatable access, and symbolic-link resolution.


Sinity:
Both review findings triaged. One actionable, one false positive.

**@coderabbitai — actionable, fixed in 6e41a108c.** You were right, and the code was broader than the comment sitting next to it: the check rejected any pointer whose path contained a `.index-generations` component, while the comment claimed only a link into this archive's own generations was refused. The extra breadth is actively harmful, because a misclassified pointer is not rejected — it is silently rewritten back to a stale local `index.db`.

Fixed by keying on what actually breaks the invariant. `generations_root` and `transactions_root` derive from the pointer's *parent*, so the failure mode is being **inside a generation**, not sharing a directory name. `_is_generation_member` now tests depth relative to the generations root: a path one or more directories below it is refused, a path sitting directly in it derives self-consistent roots and is allowed.

I kept it keyed on depth rather than on locality, which I think is stronger than the "compare against this archive's own generation root" suggestion. A farm pointing into some *other* archive's generation produces the identical nested `<gen>/.index-generations` breakage, so scoping the check to the local archive would have opened a second hole while closing the first. Two tests cover the boundary in both directions (`..._directly_inside_a_similarly_named_directory_is_followed`, `..._inside_another_archives_generation_is_refused`). Reverting the predicate to the broad name match fails the first of them.

**@chatgpt-codex-connector — false positive, no change.** The scenario requires `.index-active-pointer` to name a promoted generation while the conventional `index.db` is a divergent stale file, produced by an interrupted promotion. That state is unreachable: `promote()` never writes the anchor. It reads `self.active_pointer` to know which path to replace with a symlink, and the only write to `.index-active-pointer` in the codebase is in `__init__` (one site, confirmed by grep over the module).

So a generation-valued anchor has exactly one provenance — the bug this PR fixes, where `__init__` followed `index.db`'s symlink and recorded the resolved target. In that state the anchor and `index.db` **agree** by construction, because the anchor's value was derived from following that very symlink. There is no divergence to lose, and rewriting the anchor to `index.db` is lossless: `index.db` still symlinks to the same generation.

Worth adding that the pre-existing behaviour was to `raise` on that anchor, which did not preserve the promoted generation either — it just refused to open the archive at all, permanently. That is what blocked the real rebuild.

If you were pointing at a *latent* risk — that a future `promote()` which writes the anchor before swapping the symlink would make this repair lossy — I agree that would be a genuine hazard, but it would be a hazard in that change, and the ordering constraint belongs with it rather than as a guard here against a state nothing can currently produce.

PR #3399, opened 2026-07-30
chore(audit): add the corpus acceptance gate for absences and fidelity

Sinity:
## Summary

Adds a read-only corpus audit that measures the two things the operator's bar actually requires — **no absences** and **maximum fidelity** — and exits non-zero on failure so it can gate the pending full index rebuild.

## Problem

Neither half was measured by anything.

`verify-archive`'s `source-index-coverage` counts superseded revisions as missing work (`polylogue-ey3r`), so it cannot reach zero on any archive that has ever ingested a conversation twice — which is every real archive — and therefore cannot serve as a rebuild acceptance criterion. And nothing measured fidelity at all: an archive can report perfect coverage while every attachment whose bytes it holds is recorded `unfetched`. That is not hypothetical; it is the live state measured on 2026-07-30.

## Solution

`.agent/scripts/corpus-fidelity-audit.py`, `mode=ro` on every connection, three measures:

1. **Absences** — logical documents (`origin` + `provider_session_id`) the archive holds evidence for but does not surface, bucketed by cause (`ambiguous-only`, `mixed-ambiguous`, `settled-yet-absent`) so a fix's effect is attributable instead of a single number moving for unknown reasons.
2. **Attachment fidelity** — acquired vs not-acquired refs split by origin and `upload_origin`, because a Drive-hosted reference never fetched is actionable while a byte-less attachment kind is not.
3. **Revision fidelity** — documents whose indexed evidence is smaller than the largest revision recorded for them.

## Baseline (live archive, frozen with the daemon stopped)

```
ABSENCES  1,009 of 18,248 known documents
     587  claude-ai-export/ambiguous-only
     184  claude-code-session/ambiguous-only
     135  chatgpt-export/ambiguous-only
      71  aistudio-drive/settled-yet-absent
      20  unknown-export/settled-yet-absent

ATTACHMENT FIDELITY  acquired=2,118  not-acquired=7,655
   3,684  chatgpt-export/oauth/unfetched
   1,975  aistudio-drive/drive/acquired
   1,119  aistudio-drive/drive/unfetched

REVISION FIDELITY  94 documents below best recorded evidence
      76  hermes-session
      16  claude-code-session

VERDICT: FAIL
```

The `settled-yet-absent` buckets (71 drive, 20 unknown-export, 1 codex) are not explained by any currently-tracked cause — documents with no ambiguous decision anywhere that are nonetheless missing. They want their own investigation.

## A false positive this already caught

The first version compared indexed *messages* against `raw_session_memberships.message_count` and reported **474** shortfalls, 294 of them `codex-session`. All false. `message_count` was recorded by whichever parser censused that raw, and index v46 deliberately reclassified a large share of Codex/Claude Code rows from chat turns into typed `session_events`. One codex session read as "15 indexed vs 68,553 recorded" when it in fact holds 15 messages plus 84,612 events.

Counting `messages + session_events` drops it to 94. The docstring records the trap, because cross-generation counts are only approximately comparable and any metric built on them needs checking against a real sample before its number is believed — I only caught this by inspecting five worst-offenders rather than trusting the aggregate.

## Verification

```
ruff check .agent/scripts/corpus-fidelity-audit.py     All checks passed!
python3 .agent/scripts/corpus-fidelity-audit.py        VERDICT: FAIL (baseline above)
```

Run against the live archive read-only; the daemon is stopped and the archive is frozen, and the script opens every tier `mode=ro`, so this cannot have mutated it.

## Scope note

Deliberately a script under `.agent/scripts/` rather than a `devtools` command with a `CommandSpec`. The fixes it measures are still in flight and its thresholds will move as they land; `polylogue-f1vg` tracks promoting it to a first-class enforced gate afterwards, wired into the post-rebuild acceptance path alongside `verify-archive`. Committed with `-f` because `.gitignore` excludes `.agent/*`.

Ref polylogue-f1vg

## Summary by CodeRabbit

* **New Features**
  * Added a read-only corpus fidelity audit that checks missing documents, attachment acquisition, and revision completeness.
  * Reports detailed `PASS` or `FAIL` results in the console, with optional JSON output.
  * Provides diagnostic details for missing records, unavailable attachments, and revision discrepancies.

* **Documentation**
  * Added a tracked corpus acceptance issue defining fidelity requirements and baseline results.


Sinity:
All three findings accepted and fixed in `8a541956a`. Each one was letting the gate under-report, which is the worst failure mode for an acceptance gate — a green light is more dangerous than no light.

**P1, singleton raws (`:55`)** — correct, and the omission was larger than the comment implies. The census started from `raw_session_memberships` only, so every byte-revision-governed document was invisible. Enumerating `raw_sessions.logical_source_key` for raws with no membership row found **18 more absences immediately** (17 `aistudio-drive`, 1 `hermes-session`). It also exposed **561 raws carrying no stored identity to key on at all** — those are now their own failing measure rather than being silently dropped, since an unattributable raw is exactly the kind of gap a gate should surface rather than round away.

**P1, verdict ignoring attachments (`:201`)** — correct. The verdict could have printed PASS over 7,655 references whose bytes were never read, which is precisely the half of "maximum fidelity" this exists to enforce. `refs_not_acquired` now gates. I deliberately did **not** add an allowlist yet: a not-acquired reference clears by being acquired or by being shown genuinely unfetchable (deleted upstream, over the size cap, a byte-less attachment kind), and that judgement should live in an explicit written justification rather than in a default that quietly passes. Better to have it fail loudly now and earn each exemption.

**P2, mismatched units (`:159`)** — correct, and this one caught me over-correcting. My first version compared indexed messages against `message_count` and produced 474 false positives from the v46 message→event reclassification. Switching to messages+events fixed those but, exactly as you say, introduced false *negatives* — one message plus nine events satisfying a recorded ten while nine messages are genuinely missing.

Now compared message-to-message, the unit `message_count` is actually written in (`len(projection.message_hashes)`), with the events surplus recorded as an **explanation** for a shortfall rather than a substitute for it. Both numbers are reported so neither hides the other: **94 unexplained, 369 explained by reclassification**. That is strictly more informative than either single number I had before.

Resulting baseline on the frozen live archive:

```
absent_documents                     1,027   (was 1,009)
unexplained_revision_shortfall          94
unacquired_attachment_refs           7,655   (previously not gated at all)
raws_without_attributable_identity     561   (previously invisible)
VERDICT: FAIL
```

Worth stating plainly: this review moved the gate from one that would have passed a materially broken archive to one that fails it for four independent reasons. That is a real save, not a nit.

PR #3407, opened 2026-07-30
refactor(tests): make host-clock reads unreachable instead of allowlisted

Sinity:
## Summary

Replaces `docs/plans/test-clock-allowlist.yaml` (a hand-kept registry of 34
files exempted from a separate `devtools verify test-clock-hygiene` AST scan)
with an autouse pytest fixture that makes direct host-clock reads from test
code raise immediately, instead of merely detecting them after the fact.

## Problem

The allowlist was a rung-1 governance tell (see the parallel effort tracked
under polylogue-ganm/polylogue-2jga): the capability to call
`datetime.now()`/`time.time()` directly from test code stayed fully
available, and the only thing standing between a new violation and a clean
`devtools verify` run was a developer remembering the separate lint existed.

## Solution

- `tests/infra/clock_guard.py`: a new autouse fixture patches
  `time.time`/`time.monotonic`/`time.monotonic_ns`/`time.time_ns` with a
  wrapper that inspects the immediate caller's stack frame and raises only
  when called from guarded test code (production code exercised by the test
  is untouched — its frame lives under `polylogue/`, not `tests/`).
- For `datetime.now`/`.utcnow()`: `datetime.datetime` is an immutable C type
  (`datetime.datetime.now = ...` raises `TypeError`), so it can never be
  monkeypatched directly. The fixture instead patches the guarded test
  module's own `datetime` symbol to a raising subclass — the same
  per-module-symbol technique `frozen_clock` already uses to pin
  `datetime.now` in production modules, applied here to test modules.
- `tests/infra/` and `conftest.py` stay exempt (they're the harness, not the
  tests being guarded), matching the old lint's scope.
- Tests that genuinely need the real clock opt out inline, next to the code
  that needs it, instead of via an external YAML row:
  `pytestmark = pytest.mark.uses_real_clock("reason")`. All 34 previously
  allowlisted files got this marker with their original rationale carried
  over verbatim (one dead YAML entry —
  `tests/unit/storage/test_raw_ingest_artifacts.py`, which no longer exists
  in this checkout — was dropped rather than carried forward).
- Deleted `devtools/verify_test_clock_hygiene.py` and the YAML allowlist, and
  removed the `verify test-clock-hygiene` step from `devtools verify`
  (`devtools/verify.py`, `devtools/command_catalog.py`). The property is now
  enforced on every pytest invocation instead of only when someone runs a
  separate verify step.
- `tests/unit/infra/test_clock_guard.py` proves the guard actually blocks the
  capability: it calls `time.time()`, `time.monotonic()`,
  `time.monotonic_ns()`, `time.time_ns()`, `datetime.now()`, and
  `datetime.utcnow()` directly and asserts each raises `RuntimeError`
  pointing at `frozen_clock`, and proves both the `frozen_clock` fixture and
  the `uses_real_clock` marker correctly bypass the guard.

## Verification

- `devtools verify --quick` — all 18 steps green (ruff format/check, mypy,
  render all --check, topology/layering/closure-matrix/manifests/
  ci-workflows/doc-commands/docs-coverage/test-infra-currency/
  pytest-timeout-overrides/degrade-loudly/hash-boundary-census,
  schema-versioning policy, schema promotion audit).
- `devtools test tests/unit/infra/test_clock_guard.py
  tests/unit/infra/test_frozen_clock.py tests/unit/devtools/test_verify.py`
  — 97 passed.
- `devtools test tests/unit/storage/test_scale.py
  tests/unit/security/test_path_sanitization.py
  tests/unit/pipeline/test_resilience.py
  tests/unit/daemon/test_daemon_lifecycle.py
  tests/unit/cli/test_query_exec_laws.py tests/fuzz/fuzz_timestamp.py` — 276
  passed (representative sample of the 34 opted-out files, each exercising a
  different bypass reason: timing benchmarks, real-thread waits, opaque
  metadata timestamps, fuzz harnesses).
- `python -m pytest tests/unit -q -m "not scale_medium and not scale_large"`
  — 682 passed, stopped at the one pre-existing, unrelated
  `test_check_output_snapshot` failure (documented pre-existing failure, not
  caused by this change).

Not run: the full `tests/integration` suite (slow, protected) and
`tests/benchmarks` — the guard's opt-out marker was applied identically to
those files and spot-checked by reading the diff, but not executed in this
session.


Sinity:
Both review findings triaged. One fixed, one accepted and beaded rather than papered over.

**Keep guarding clocks not frozen by `frozen_clock` — fixed in `dcc82f5be`.** Correct and worth catching: `freeze_clock` patches only `time.time`, `time.monotonic`, and the `datetime` symbol in named modules, so bailing out entirely left `time.time_ns` and `time.monotonic_ns` reachable — in exactly the tests most likely to read a clock. The guard now *narrows* instead of disabling: the nanosecond clocks stay guarded unconditionally, and only the callables `freeze_clock` actually manages are skipped.

Verified with `devtools test tests/unit/infra` (38 passed) plus a real `frozen_clock` consumer, `tests/unit/archive/test_session_revision_membership.py` (39 passed), which proves the narrowed guard doesn't break tests that legitimately manage the clock.

**Guard clock reads before test collection — accepted, filed as `polylogue-plfr`, not fixed here.** Also correct, and it means this PR's framing overstates what it delivers. The guard installs as an autouse fixture, so it arms after pytest has imported the test module; a module-level clock read — a constant, a decorator argument, a parametrize value — executes during collection and escapes. The static AST lint this replaces *did* catch that shape, so on this one axis the runtime guard is weaker than its predecessor.

I'm not fixing it in this PR because the fix is a restructure, not a patch: `pytest_configure` runs before collection and the existing `_time_raiser` caller-frame check would scope it correctly, but the per-module `datetime` rebinding can't work at configure time (the module set isn't known yet) and would need to become a guarded patch of `datetime.datetime` itself. The `tests/infra`/`conftest.py` exemptions have to survive that move too. That's worth doing properly rather than rushing at the end of a long session.

So the honest claim for this PR is "unreachable from inside a test function", not "unreachable". The bead records the gap, the direction, and the acceptance criteria — including that the word "unreachable" should only be used where it's true.

PR #3411, opened 2026-07-31
fix(browser-extension): restore live capture, delete DOM fallback

Sinity:
## Summary

Investigates and fixes the live browser-capture pipeline: why capture stopped landing anything, why the DOM-scrape fallback ever fired for real (and why it no longer needs to), why ChatGPT temporary chats never captured despite the archive side being ready for them, and a silent structured-block drop at the envelope boundary.

**Merge-coordination note:** an earlier revision of this PR also fixed the ChatGPT backfill bridge's content/metadata field-dropping in `page_transport.js`/`providers.js`. PR #3412 fixed the same defect independently with a strictly stronger solution (verbatim forwarding + byte-bounded chunking, removing the 24 MiB cap entirely rather than working under it) and also covers Claude's equivalent gap. Per coordinator instruction, those changes were dropped from this PR (see the `revert:` commit) — #3412 owns `page_transport.js`. This PR will need a rebase onto master once #3412 merges. Separately, PR #3410 found and fixed a related allowlist-drop bug in `common.js`'s `buildEnvelope` (turns silently lost their `blocks` field); since this PR already owns the content-script files, the fix plus a regression test were folded in here at the coordinator's request.

## Problem

1. **Live capture was dead.** The daemon's journal shows zero `batch ingested browser-capture` events from 2026-07-19 through 2026-07-31, and the live archive spool (`/realm/db/polylogue/browser-capture`) had exactly zero files before this session. Root cause (confirmed from the live Chrome profile's `Preferences`): the unpacked dev extension was loaded from `/realm/worktrees/polylogue-browser-live-proof/browser-extension`, a git worktree later removed as routine post-merge cleanup. Once the directory was gone, Chrome silently stopped running the extension's service worker — no error anywhere. This is the browser-extension analog of the "worktree removal must check live processes" hazard.

2. **`chatgpt-dom-v1`/`claude-ai-dom-v1` were dominated by native, but real.** `claude-ai-dom-v1` had zero captures ever. `chatgpt-dom-v1` had 17 captures, all inside one 4-minute native outage on 2026-07-17, producing degraded 3-turn shadows of conversations the export holds in full — a brand-new conversation whose URL hadn't been published yet by ChatGPT's SPA router, so native had no id to fetch or match by, even though the page's own fetch had already been intercepted.

3. **Zero ChatGPT temporary chats had ever landed**, despite the archive side being fully ready (`SessionKind.TEMPORARY`, ingest flag, parser support). Root cause: an asymmetry between `common.js`'s URL parser (maps a temporary-chat URL to a sentinel) and two independent gates that returned `null` for the same condition — `background.js`'s own `conversationIdForUrl` (silently skipped ever sending the tab a capture message) and `chatgpt.js`'s `parseNativeCapture` (required a URL-derived id to accept an already-intercepted capture, which a temporary chat never gets).

4. **`common.js`'s `buildEnvelope` silently dropped every turn's structured `blocks`** (found by PR #3410) via a fixed field allowlist with no `blocks` entry — the same allowlist-drift class as #3412's bug, one layer down the pipeline.

## Solution

- **Live remediation (not part of the diff):** reloaded the extension from the stable main checkout, re-paired it against the running receiver, confirmed capture resumed — two fresh `chatgpt-native-v1` captures landed within seconds of activating a conversation tab. `README.md` now documents the hazard.
- **`chatgpt.js`:** `fetchNativePayloadOnDemand` now waits (bounded 6s poll) for a fresh conversation's id before giving up. `parseNativeCapture` accepts an already-intercepted capture with no URL-derived id when the page is a temporary chat and the payload agrees (`is_temporary === true`), using the provider's own ephemeral id. `background.js`'s `conversationIdForUrl` now returns the same sentinel `common.js` already used instead of `null`.
- **Deleted** `chatgpt-dom-v1`/`claude-ai-dom-v1` entirely from `chatgpt.js`/`claude.js`. A capture with no native payload now returns `{ok:false, error:"native_capture_unavailable"}`. `grok.js` is untouched.
- **`common.js`:** added the missing `blocks: Array.isArray(turn.blocks) ? turn.blocks : []` pass-through, plus `tests/common_envelope_boundary.test.js` — a maximal turn/block round-trip through the real source (not a copy) asserting every `BrowserCaptureTurn`/`BrowserCaptureBlock` field (`polylogue/browser_capture/models.py`) survives. Field lists are hardcoded rather than derived from the Python models at test time (no cross-language plumbing exists in this repo and `npx vitest run` is a Python-free command); noted as a follow-up.
- **Ported `tests/content/chatgpt.test.js` and `tests/content/claude.test.js`** off hand-copied function bodies onto the real-source-loading pattern the sibling test files already use. `claude.test.js`'s old copy had already silently drifted from reality (it tested a `conversationApiUrlFromResources`/`organizationIdFromStorageKeys` pair that doesn't exist in `claude.js` at all — the real functions live in `claude_bridge.js` under different names/signatures), passing against fiction the whole time.
- Deleted the now-dead DOM-fallback tests rather than updating them.

## Verification

```
npx vitest run
```
334 passed, 1 pre-existing failure (`tests/build.test.js`'s foreground-tab-activation case — confirmed failing at HEAD before any change in this branch, unrelated).

```
npx eslint src/ tests/
```
clean.

Live end-to-end: activated an existing ChatGPT conversation tab, a fresh capture landed in `/realm/db/polylogue/browser-capture/chatgpt/` within ~10s.

Not run: Claude.ai live end-to-end (no open Claude conversation tab available without sending a live message); Grok (out of scope, untouched).

**Pending:** rebase onto master once PR #3412 merges (expected conflicts in `background.js`/`providers.js`/`tests/backfill.test.js` per merge-coordination guidance — merge order is #3412, then this PR, then #3410).

### DIFF
--- a/browser-extension/README.md
+++ b/browser-extension/README.md
@@ -28,6 +28,20 @@ For active development against the in-tree source:
 3. Click **Load unpacked** and select this directory (`browser-extension/`)
 4. Pin the extension to the toolbar so the badge is always visible
 
+**Load from a stable, permanent checkout only** (e.g. the main
+`/realm/project/polylogue` clone) — never from a throwaway agent worktree.
+An unpacked extension is just a filesystem reference: if the directory it
+points at is deleted (routine `git worktree remove` after a merged PR is
+explicitly authorized as non-destructive cleanup elsewhere in this repo's
+conventions), Chrome silently stops running its service worker with no
+error surfaced anywhere — not in the extension's own UI, not in the
+receiver's logs, nothing. This has happened for real: an unpacked install
+pointed at an agent worktree that was later cleaned up, and live-capture
+was silently dead from that moment until someone thought to check
+`chrome://extensions` and noticed the entry had no running service worker.
+If you must develop from a worktree, copy or symlink the built extension
+into a stable location and load unpacked from there instead.
+
 ### 2b. Chrome / Chromium — packed `.zip` from a release
 
 For users who want a stable artifact rather than a working tree:
--- a/browser-extension/src/background.js
+++ b/browser-extension/src/background.js
@@ -1285,6 +1285,13 @@ async function setStateForTab(tabId, state, expectedTabUrl = null) {
     expectedProvider
     && expectedSessionId
     && activeSessionId
+    // TEMPORARY_CHAT_SENTINEL is a "this tab has a real, capturable
+    // conversation" signal, not the conversation's real identity -- it can
+    // never equal a captured state's true provider_session_id (the
+    // ephemeral id from the native payload), so this staleness guard would
+    // reject every state write for a temporary chat tab, including the
+    // capture that just succeeded on it.
+    && activeSessionId !== TEMPORARY_CHAT_SENTINEL
     && (
       activeProvider !== expectedProvider
       || activeSessionId !== expectedSessionId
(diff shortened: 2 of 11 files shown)


Sinity:
Addressed the P1 finding: `setStateForTab`'s staleness guard had the identical sentinel-vs-real-id mismatch as the freshness-hint handler (and was actually masking each other's bugs — `captureTab` was previously passing the sentinel too, which "matched" only because both sides were equally wrong). Fixed all three sites (freshness-hint handler, `setStateForTab` guard, `captureTab`'s `pageSessionId` derivation) in 793fb0f74, with regression tests for both the exemption and that it doesn't weaken the check for ordinary sessions.

Filed polylogue-upbv for the remaining, lower-severity part (archive-state UI staying inaccurate for temporary chats — not a data-loss bug, since content-hash dedup makes the redundant re-captures cheap).

PR #3412, opened 2026-07-31
fix(browser-extension): stop dropping ChatGPT thoughts/citations in the backfill bridge

Sinity:
## Summary

`page_transport.js`'s ChatGPT conversation projection (`compactChatGptContent`/metadata allowlists) silently collapsed every `thoughts` (reasoning) node and dropped `content_references`/`citations`/`finish_details`/`search_result_groups`/`aggregate_result`/attachment `size`+`extracted_content` to fit under a hard 24 MiB per-call bridge budget. Fidelity now never trades off against conversation size: the bridge forwards content/metadata verbatim and chunks the transfer across multiple bounded `chrome.scripting.executeScript` calls instead of dropping fields.

## Problem

Measured on conversation `6a330cc0-1abc-83eb-baae-59e5395ad725`: the browser-captured session had 0 `thoughts` blocks; the GDPR export for the same conversation has 828. The message-id overlap between the two captures was exactly 232 — exactly the export's plain-`text` message count, i.e. every non-text content type was silently dropped. 465 of ~570 spooled captures used this path. A second allowlist on `message.metadata` discarded the entire citation graph (16,842 URLs in one export) plus tool-sandbox output evidence.

The 24 MiB bridge cap exists because Chrome's `chrome.scripting.executeScript` result serialization has a hard ceiling around 32 MiB; the previous code used field-dropping as a cheap way to stay under it.

## Solution

- `compactChatGptContent`/`compactChatGptMetadata` (`browser-extension/src/backfill/page_transport.js`) now forward the content/metadata objects as-is — no per-content-type modeling. `polylogue/sources/parsers/chatgpt.py` already reads `thoughts`, `content_references`, `citations`, `search_result_groups`, `aggregate_result`, etc. directly off message metadata for GDPR-export parsing, so this also drops the old `"chatgpt-native-compact-v1"` projection tag, which routes browser captures through that same already-correct/tested native parser path instead of the lossier turn-array fallback (`_has_chatgpt_native_payload` in `browser_capture.py` excludes only that exact tag string — untouched, no Python change needed).
- `buildChatGptChunks()` packs the full-fidelity mapping into ordered, byte-bounded, **node-aligned** chunks (a node's JSON is never split). The parsed raw source is cached on a MAIN-world page global (`window.__polylogueChatGptChunkCache`) keyed by native id with a 5-minute TTL, so pulling chunk *N > 0* re-serves from cache instead of re-fetching the conversation from ChatGPT's API.
- `background.js`'s `providerPageFetch` (via new `reassembleChunkedChatGptConversation`) loops `chrome.scripting.executeScript` once per chunk and merges each chunk's `mapping` slice before the capture reaches the receiver. Each individual call stays under Chrome's serialization limit; the reassembled conversation in the extension process has no such cap.
- `providers.js` drops the now-dead "compact" capture tier (`captureFidelity` is unconditionally `native_full` for ChatGPT backfill captures now) and — a smaller sibling gap found during the audit — `chatGptText()`/`claudeText()` (used for the session-turns *summary*, not the archival payload) didn't know how to read `content.thoughts`/`content.thinking` blocks either, so a reasoning-only turn produced empty text and could get silently dropped by the `no_turns` skip check even though the archival record was fine.

### Options considered (per the task's evidence-first instruction)

Evaluated: (a) chunk/stream across multiple bridge calls, (b) POST directly from page context to the local receiver, (c) ArrayBuffer/port streaming, (d) generic full-fidelity projection. Went with **(a) + (d) together**: full-fidelity projection removes the *reason* to drop fields, and chunking (not a page-context network POST) keeps the change inside the existing `chrome.scripting.executeScript` request/response shape the rest of the bridge already uses — no new page-context network surface, no new CSP/CORS considerations, no receiver-endpoint knowledge leaking into MAIN-world page code. The real conversation-size ceiling observed (~996 messages / 720 tool nodes) is well within the 64 MiB raw-source cap and comfortably chunks into single digits of ~12 MiB packing-target chunks.

## Repo-wide silent-drop audit

Reviewed all 19 `return {}`/`return []` seed sites in `browser-extension/src/`. Two were genuine silent data-loss bugs (fixed above, both in `src/backfill/`, this PR's owned surface). The rest are legitimate: per-part filtering inside an already-safe array (e.g. skipping a non-text-bearing `parts` entry), UI-popup fallbacks on a failed runtime message (display degradation, not archival loss), and hostname-routing defaults (no matching provider = no scripts to inject). `src/content/*.js` and `src/background.js`'s capture-liveness logic are owned by a parallel lane per the task's coordination note; I made one minimal, necessary edit to `background.js` (the chunk-reassembly loop) and left everything else in that file untouched.

On the Python side, Polylogue already has an established, enforced "degrade-loudly" doctrine (`devtools verify degrade-loudly`, `docs/plans/degrade-loudly-allowlist.yaml`, bead `polylogue-cpf.4`, closed) — nothing in this change touches Python, and no new parallel mechanism was invented; the JS fixes are direct (make the bad shape unrepresentable) rather than a new lint.

## Verification

```
cd browser-extension && npx vitest run
```
368 passed, 1 pre-existing failure (`tests/build.test.js`'s packaged-service-worker fixture — confirmed identical on `origin/master` via `git stash`; not touched by this change).

New tests:
- `page_transport.test.js`: real `thoughts` node fidelity, `content_references`/citations/attachments-with-`size`+`extracted_content` metadata fidelity, multi-chunk build + MAIN-world cache reuse (proves no re-fetch on chunk N>0), fail-loud on a stale/expired chunk request (`backfill_bridge_chunk_cache_miss`, not a truncated reassembly).
- `background.test.js`: end-to-end integration test proving `providerPageFetch`'s chunk loop drives a real second `chrome.scripting.executeScript` call and the merged mapping reaches the envelope submitted to the receiver.
- `backfill.test.js`: `chatGptText`/`claudeText` no longer misclassify a reasoning-only turn as `no_turns`.

`npx eslint src/ tests/` clean on all touched files.

Ref polylogue-thoughts-fidelity (no tracking bead was filed — self-contained fix, PR body is the record per the "skip issues for self-contained fixes" convention).

### DIFF
--- a/browser-extension/src/backfill/page_transport.js
+++ b/browser-extension/src/backfill/page_transport.js
@@ -137,6 +137,16 @@ export async function executeProviderPageRequest(request) {
       create_time: typeof source.create_time === "string" || typeof source.create_time === "number" ? source.create_time : null,
       update_time: typeof source.update_time === "string" || typeof source.update_time === "number" ? source.update_time : null,
       current_node: typeof source.current_node === "string" ? source.current_node : null,
+      // polylogue.sources.parsers.chatgpt.parse() reads these three
+      // top-level fields to set session_kind (is_temporary) and
+      // provider_project_ref (conversation_template_id/gizmo_id) -- omitting
+      // them silently downgraded every temporary/project-scoped ChatGPT
+      // conversation to a standard/unscoped session the moment this
+      // projection started advertising itself as native_full instead of the
+      // old degraded/compact signal.
+      is_temporary: typeof source.is_temporary === "boolean" ? source.is_temporary : null,
+      conversation_template_id: typeof source.conversation_template_id === "string" ? source.conversation_template_id : null,
+      gizmo_id: typeof source.gizmo_id === "string" ? source.gizmo_id : null,
     };
   }
 
@@ -238,7 +248,19 @@ export async function executeProviderPageRequest(request) {
       header = chatGptConversationHeader(source);
       projectedChunks = buildChatGptChunks(source);
       if (projectedChunks.length > 1) {
-        chunkCache.set(nativeId, { chunks: projectedChunks, header, createdAtMs: Date.now() });
+        const entry = { chunks: projectedChunks, header, createdAtMs: Date.now() };
+        chunkCache.set(nativeId, entry);
+        // Scheduled at insertion time, not left to a future bridge call: a
+        // tab that goes idle right after the first chunk (no further
+        // identity/inventory/conversation traffic at all) would otherwise
+        // never trigger the sweep above, holding the entry indefinitely.
+        // The `chunkCache.get(nativeId) === entry` guard matters -- without
+        // it, a stale timer firing after this nativeId's entry was already
+        // replaced by a newer fetch (same conversation re-pulled) would
+        // delete that unrelated newer entry instead of a no-op.
+        window.setTimeout(() => {
+          if (chunkCache.get(nativeId) === entry) chunkCache.delete(nativeId);
+        }, chunkCacheTtlMs);
       } else {
         chunkCache.delete(nativeId);
       }
--- a/browser-extension/src/backfill/page_transport.js
+++ b/browser-extension/src/backfill/page_transport.js
@@ -2,13 +2,31 @@ export async function executeProviderPageRequest(request) {
   const currentOrigin = window.location.origin;
   const requestTimeoutMs = 55000;
   const absoluteMaxResponseBytes = 32 * 1024 * 1024;
-  // A ChatGPT conversation is fetched in MAIN world, then reduced before the
-  // scripting-result bridge. Both stages stay bounded independently: raw
-  // provider bloat may use 64 MiB in page memory, but no more than 24 MiB can
-  // cross into extension storage/worker code. 24 MiB leaves 8 MiB of headroom
-  // beneath Chrome's established 32 MiB scripting-result limit.
+  // A ChatGPT conversation is fetched in MAIN world, then projected into the
+  // canonical native mapping shape (every content-type payload key and every
+  // metadata key preserved -- no field allowlist) before crossing the
+  // scripting-result bridge. Raw provider bloat may use up to 64 MiB in page
+  // memory. No single scripting-result call may exceed 24 MiB (8 MiB of
+  // headroom beneath Chrome's established 32 MiB scripting-result limit), but
+  // fidelity never trades off against size: a projection that would exceed
+  // the per-call bridge budget is split into ordered, node-aligned chunks
+  // (never splitting a single node's JSON) and the caller re-invokes this
+  // function once per chunk, reassembling the full mapping on the extension
+  // side. See chunkedChatGptProjection() below.
   const compactChatGptSourceMaxBytes = 64 * 1024 * 1024;
   const compactChatGptBridgeMaxBytes = 24 * 1024 * 1024;
+  // Target size when packing nodes into a chunk, well under the hard bridge
+  // cap so JSON-escaping overhead measured by assertScriptingResultBound
+  // (quotes, control characters, the extension-side wrapper) has headroom.
+  const chatGptChunkPackTargetBytes = compactChatGptBridgeMaxBytes / 2;
+  // A chunked conversation fetch does exactly one network round-trip: the
+  // raw parsed source is cached in this MAIN-world page global (persists
+  // across the sequential executeScript calls that pull each chunk) so
+  // chunk N>0 does not re-fetch. Keyed by native conversation id.
+  const chunkCacheTtlMs = 5 * 60 * 1000;
+  const maxChatGptChunks = 64;
+  window.__polylogueChatGptChunkCache = window.__polylogueChatGptChunkCache || new Map();
+  const chunkCache = window.__polylogueChatGptChunkCache;
   const originalFetch = window.fetch;
   const uuidPattern = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
   const maxResponseBytes = Number.isInteger(request?.maxResponseBytes)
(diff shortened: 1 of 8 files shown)


Sinity:
Addressed all three findings (2 commits: f89fc557f for the browser-extension pair, 268a6be6e for the parser fix):

**P1 (children ordering, page_transport.js:135)** — `projectChatGptNode()` now forwards each node's `children` array verbatim (array of node-id strings, same shape the provider sends). `polylogue.sources.parsers.chatgpt.extract_messages_from_mapping` derives `branch_index` from a parent's `children` position, so this was silently losing branch/variant ordering the moment this projection became the durable raw capture — correct call that it's the same defect class one level up. New test: "preserves each node's `children` ordering across the projection".

**P1 (thoughts text still empty end-to-end, page_transport.js:224)** — confirmed and fixed in `polylogue/sources/parsers/chatgpt.py`: `_extract_content_text()` only read `parts`/`text`/`result`, so a `thoughts` node produced a THINKING block with empty text even with the raw bytes preserved. Added a `thoughts`-array fallback (prefer each step's `content`, fall back to `summary`). Verified **behaviorally**, not structurally, per your instruction: new test `test_chatgpt_thoughts_node_produces_nonempty_thinking_text_end_to_end` parses a thoughts-only node, asserts the message/block text is non-empty, then writes it through the real archive writer (`write_session_sync`) and queries the live `messages_fts` index (via the rowid→block_id identity ledger a contentless FTS5 table requires) to prove the content is actually findable via search — not just stored. Reverted the fix locally and confirmed the test fails (`assert '' == '...'`) without it. `devtools test tests/unit/sources/test_parsers_chatgpt.py`: 110 passed. `devtools verify --quick`: exit 0. PR #3408 is already merged into master and this branch is current with it; no conflict, `_extract_content_text` was untouched by that PR.

**P2 (chunk cache never evicts, page_transport.js:218)** — fixed two ways: (1) the cache entry for a fully-served conversation is now deleted the instant its last chunk is returned, not left for a future TTL check that may never come; (2) a sweep at the top of `executeProviderPageRequest` (runs on every bridge call — identity checks happen far more often than conversation fetches) evicts any entry past its TTL even if that native id is never looked up again, covering the abandoned-midway case you called out. New test "evicts an abandoned-midway chunk cache entry..." inspects `window.__polylogueChatGptChunkCache.has(nativeId)` directly rather than relying on a later same-id lookup (which the pre-existing lazy check would already satisfy without a real fix) — reverted locally and confirmed it fails without the sweep/delete code.

Full verification: `npx vitest run` in `browser-extension/` — 370 passed, 1 pre-existing failure (`tests/build.test.js`'s packaged-service-worker fixture, confirmed identical on `origin/master` via `git stash`, unrelated to this PR). `npx eslint` clean. `devtools verify --quick` (pre-push hook) exit 0 both commits.

Will wait for CI before merging per policy.

Sinity:
@codex review

Sinity:
Addressed the second review pass (3 more findings, commit fbafc7706) — all direct consequences of the chunking fix removing the old bridge's implicit ~24 MiB ceiling, same pattern as before.

**P1 (background.js:1727, oversized payload through the exact-capture channel)** — confirmed: `chrome.tabs.sendMessage` (used to hand the reassembled body to the content script as `nativePayload`) is a second, independent IPC channel with its own size ceiling, previously safe only because the old single-shot bridge implicitly capped `nativePayload` at ~24 MiB. `captureOverride` now measures the reassembled body and only forwards it when it fits a conservative 24 MiB budget; over that, it forwards nothing (the content-script side falls back to its own pre-existing native-fetch/DOM path) and background.js attaches the full `nativePayload` to `raw_provider_payload` in-process afterward, so archival fidelity never actually degrades even for a conversation too large for this second channel. New test drives a real ~28 MiB / 4-chunk conversation through the full backfill flow and asserts both that `sendMessage`'s `nativePayload` was `null` and that the receiver still got the full merged mapping.

**P2 (page_transport.js:139, missing is_temporary/conversation_template_id/gizmo_id)** — confirmed and fixed: `chatGptConversationHeader()` now forwards these three fields, which `polylogue.sources.parsers.chatgpt.parse()` reads for `session_kind`/`provider_project_ref`.

**P2 (page_transport.js:41, sweep needs another request)** — confirmed and fixed: a `window.setTimeout` scheduled at cache-insertion time (guarded against a stale timer clobbering a newer entry for the same nativeId) now evicts an abandoned entry even if the tab never makes another bridge call at all.

Each fix verified with anti-vacuity (reverted locally, confirmed the new/existing test fails without it) — details in the commit message. `npx vitest run`: 373 passed, 1 pre-existing failure (unrelated, confirmed on `origin/master`). `npx eslint` clean. `devtools verify --quick`: exit 0.

@codex review

Sinity:
Addressed round 3 (2 more findings, commit 579f0bcf8) — thank you for catching that the round-2 fix for the oversized-payload case was itself incomplete.

**P1 (background.js:335, bypass exact capture entirely for oversized payloads)** — confirmed: passing `nativePayload: null` doesn't select a bounded fallback in content/chatgpt.js — it independently re-fetches the same conversation and returns it in the `sendMessage` response direction, and explicitly disables its DOM fallback when a conversation id is pinned. Rather than patch content/chatgpt.js (owned by a parallel lane on this branch), `captureOverride` now **skips `captureProviderConversation` entirely** for an oversized conversation and calls `adapters.chatgpt.normalizeCapture()` directly — zero `chrome.tabs.sendMessage` round trips in either direction. This also simplifies away the previous fix's post-hoc envelope-patching.

**P2 (providers.js:67, thoughts handling in the exact-capture normalizer)** — confirmed as a real production gap: `captureOverride` always tries the exact-capture path first for chatgpt, and that path's own normalizer doesn't read `thoughts` — a thoughts-only conversation got zero turns from it and hit the `no_turns` retry loop forever, never actually reaching the already-thoughts-aware adapter path. Fixed the same way, staying in-lane: `captureOverride` now tries the exact-capture path first (preserving its DOM/live-generation enrichment for the normal case) but falls back to `adapters.chatgpt.normalizeCapture()` whenever it comes back with zero turns, instead of returning an empty capture.

Anti-vacuity: reverted `captureOverride` to the prior shape locally and confirmed both tests fail — the item never reaches the receiver in either case (stuck retrying, matching the actual reported defects), not just a structural mismatch. `npx vitest run`: 374 passed, 1 pre-existing failure (unrelated). `npx eslint` clean. `devtools verify --quick`: exit 0.

@codex review

PR #3425, opened 2026-07-31
fix(insights): prefer typed PR/session evidence over regex correlation

Sinity:
## Summary

`analyze correlation` (the `read --view correlation` CLI path and
`Polylogue.session_correlation_payload`) regex-scans message text for
GitHub PR/issue refs and scores git commits by time-window/file-overlap
heuristics, while the archive already persists the same facts as typed
evidence with zero readers. This PR makes the correlation path consume
the typed evidence first and treats the heuristics as a fallback, per
polylogue-l9su.

## Problem

Bead polylogue-l9su (P1, filed after an independent audit) found:

- `session_refs` (kind=`pull_request`, from the Claude Code `pr-link`
  sidecar record) has 18,932 rows across 167 sessions — never read by
  `build_correlation_result`.
- `claude_bridge_session` session_events (12,154 rows) carry a
  `bridge_session_id` (`cse_<token>`) — never read either.
- This repo's own commits carry a `Claude-Session:
  https://claude.ai/code/session_<token>` trailer by convention (116
  commits in this repo's history). Verified live: the `<token>` is the
  *same* id as `claude_bridge_session`'s `bridge_session_id` minus its
  `cse_` prefix (`cse_0182HDxDpJpsbn2qcKWK6Fsf` in `session_events`
  matches this repo's own `session_0182HDxDpJpsbn2qcKWK6Fsf` trailer
  byte-for-byte). Nothing parsed it.
- `session_commits.detection_type` has a live CHECK-constraint value
  `origin_reported` with zero rows using it.

## Solution

`polylogue/insights/session_commit.py`:

- `detect_session_commits` gains `bridge_session_ids`. When a commit's
  `Claude-Session:` trailer token matches one of the session's own
  bridge-session ids, the edge is `detection_method="origin_reported"`,
  `confidence=1.0`, and supersedes file_overlap/time_window/explicit_ref
  for that commit sha. When a trailer names a *different* session than
  the one being scored, the heuristic edge is kept (not silently dropped)
  but flagged with `disagreement_note` — "derive rather than compare, fail
  loud rather than guess quietly."
- `build_correlation_result` gains `typed_pr_refs`/`typed_issue_refs`
  (`GitHubRef` rows built from `session_refs`). When supplied, they are
  authoritative for that ref kind; the regex scan of message text still
  runs (it's the only source for `file_paths`) but is used only as a
  fallback for sessions with no typed evidence, or to detect a
  disagreement against typed evidence — never to override it silently.
- New `CorrelationDisagreement` / `SessionCorrelationResult.disagreements`
  and `GitHubRef.source` (`typed_session_ref` vs `heuristic_regex`) make
  which mechanism resolved each linkage visible in the JSON/plaintext
  payload, satisfying AC3/AC4 (record which mechanism won; surface
  disagreements instead of silently preferring one).
- New pure helpers `typed_refs_from_session_refs` /
  `bridge_session_ids_from_events` convert `SessionRefRecord`/
  `SessionEvent` rows into the typed inputs above (duck-typed, so the
  module stays DB-free and unit-testable).

Wiring (point 5 — read surface): `read --view correlation` and
`session_correlation_payload` already existed as the CLI/API surface for
this data (`polylogue/cli/read_views/correlation.py`,
`polylogue/insights/correlation_view.py`); both were updated to fetch
`session_refs` via `repository.get_session_refs` and bridge ids from the
session's own `session_events`, and pass them through. No new surface was
needed — the blocker was purely that the existing surface ignored typed
evidence it already had access to.

**Incidental bug fix required for correctness**: while adding the
trailer scan I found `_parse_git_log_blocks` had a latent bug — splitting
`git log --format=%H%n%ai%n%s%n---` output on the literal `\n---\n` token
left every commit's changed-file set **empty**, because `--name-only`
appends the file list *after* the format text, landing it in the next
split segment past where the parser had already stopped consuming header
lines. This meant `file_overlap` detection never actually returned a
file-overlap edge against a real repository (only ever exercised in tests
against nonexistent paths, which never hit this code). Fixed by switching
to `%x1e`/`%x1f` ASCII field/record separators that cannot collide with
commit text.

**Explicitly out of scope**: `session_commits` (the durable batch-ingest
table in `storage/sqlite/archive_tiers/write.py`, excluded from this
lane's surface) is a separate mechanism populated from
`session.git_commit_hash` at ingest time — a narrower fact (repo HEAD at
session-capture time, not "commits this session produced") that the bead's
self-correction note already distinguishes from the on-demand correlation
path this PR fixes. It is unaffected and unchanged by this PR.

## Verification

- `devtools test tests/unit/insights/test_session_commit.py
  tests/unit/cli/test_correlate_view.py` → 45 passed. New fixtures build a
  real git repo in `tmp_path` (subprocess `git init`/`commit`) and cover:
  trailer match takes priority over file_overlap (`origin_reported`,
  confidence 1.0); a foreign trailer is flagged as a disagreement but the
  heuristic edge is kept; no `bridge_session_ids` falls back to plain
  heuristics unflagged; typed PR refs are preferred over regex and tagged
  `source=typed_session_ref`; a disagreement is recorded when regex finds
  PR numbers not in typed evidence; the `typed_refs_from_session_refs` /
  `bridge_session_ids_from_events` converters.
- `devtools verify --quick` → exit 0 (ruff format/check, `mypy --strict`,
  `render all --check`, layering/closure-matrix/manifest/doc-command
  gates) — run twice, once pre-commit and once on `git push` via the hook.
- `mypy polylogue/insights/session_commit.py
  polylogue/insights/correlation_view.py polylogue/api/archive.py` → no
  issues found.
- **Live read-only measurement** against `/realm/db/polylogue/index.db`
  (`sqlite3 ...?mode=ro`, no writes, no daemon):
  - 167 sessions carry typed `pull_request` session_refs (1,690 typed
    PR-number rows total).
  - Running the **old** regex-only `extract_github_refs()` over each of
    those sessions' own block text finds 1,934 PR-number mentions: 102
    sessions match typed evidence exactly, **65 sessions would have
    surfaced extra/different PR numbers from the regex scan alone** — this
    is exactly the silent-disagreement class the fix now makes visible
    instead of guessing.
  - Trailer side: of 9 distinct `Claude-Session` trailer tokens found in
    this repo's own git history, **8 resolve to a real archived session**
    via `claude_bridge_session` (one token maps to 2 sessions → 9 sessions
    total). Once a caller threads that session's `bridge_session_ids`
    through, those commits get `detection_method="origin_reported"`
    instead of file_overlap/time_window/nothing.
  - `session_commits` unaffected: still 2,990 rows, all
    `detection_type=explicit_ref` (out of this lane's declared surface,
    see above).

## Follow-ups

- `session_commits`' `origin_reported` CHECK slot is still unused at the
  batch-ingest layer; wiring the same trailer-parse logic into
  `archive_tiers/write.py` (explicitly excluded from this lane) is a
  separate follow-up if the operator wants the durable table to carry
  this fact too, rather than only the on-demand correlation payload.
- `typed_issue_refs` has the same priority wiring as PR refs, but no
  producer currently writes `session_refs` rows with `kind=issue` (only
  `pull_request` is populated today) — so the issue-ref path is exercised
  by unit tests but has no live typed data yet.

Ref polylogue-l9su


Sinity:
Triaged the 3 CodeRabbit P2 findings (PR #0 coercion at L785, cross-repo number collision at L739, foreign-trailer false-disagreement at L500) as real but non-blocking: they are edge cases of the new typed-evidence path (Codex Cloud refs lacking ref_number; repo-qualified identity comparison; sessions with no bridge evidence), not regressions in existing behavior, and the 45 new/updated tests cover the primary trailer-match/typed-ref/fallback paths. Filed polylogue-2vor with per-item acceptance criteria and merging. Ref polylogue-2vor.

PR #3515, opened 2026-08-01
fix(demo): self-heal stale demo schema and fix tour budget attribution

Sinity:
## Summary

Fixes two failures in the documented, private-data-free onboarding path that CLAUDE.md and TESTING.md point a cold agent at instead of the live archive: `polylogue demo seed` failed hard on a stale-schema demo archive with no self-heal, and `polylogue demo tour` overran its own 30s "first result" latency budget due to a timer-attribution bug.

## Problem

Two related failures reported in polylogue-3ycw:

1. `polylogue demo seed` against a demo archive left over from an older schema version raised:
   `RuntimeError: index.db schema version 46 is not the current index tier version 53; move it aside and rebuild the archive root`
   with no path named and no self-heal, even though a demo archive is entirely synthetic, disposable fixture data that this same command regenerates on every run.

2. `polylogue demo tour` failed its own latency budget: `First result: 70.934s` against a 30s budget, `Full tour: 86.336s` against a 420s budget (which passed). The tour's own `report.md` showed each of the 4 narrated CLI steps individually took only 4.6-5.3s — the 70.9s "first result" number was dominated by one-time archive seed+verify setup happening *before* the first narrated step, not by query latency.

I reproduced both against a fresh scratch archive (`POLYLOGUE_ARCHIVE_ROOT=/realm/tmp/... polylogue demo seed`) — a genuinely fresh root seeds cleanly, confirming issue 1 is not a fundamental demo-seed bug but a missing self-heal for a *pre-existing* stale-schema demo archive at the resolved root. Reading `polylogue/demo/tour.py` confirmed issue 2 directly: `first_result_s = time.perf_counter() - start`, where `start` is set before the seed/verify calls, not before the narrated command loop.

## Solution

- `polylogue/demo/seed.py`: added `_archive_root_is_demo_owned()`, which treats the presence of the demo fixture source directory (`demo-fixture-world-source`, written only by `materialize_demo_source`, never by the live daemon) as strong, retroactive-compatible evidence that a root's tier databases hold only synthetic demo content. `_self_heal_stale_demo_archive_tiers()` then moves aside (`<file>.stale-<timestamp>`) and rebuilds any tier whose on-disk schema version predates the current code's — but **only** when that ownership check passes. A stale-schema *foreign* root (e.g. an operator forgot `--root` and the default `archive_root()` resolved to the configured production archive) is never auto-rebuilt; it still raises loudly, protecting the live/durable archive from any auto-mutation.
- `polylogue/storage/sqlite/archive_tiers/bootstrap.py`: the (still-raised, for non-demo-owned roots) `RuntimeError` now names the full path and an explicit rebuild command (`polylogue ops reset --index && polylogued run` for `index.db`, `mv <path> <path>.stale` otherwise) instead of just the bare filename with no remediation.
- `polylogue/demo/tour.py`: rebased `first_result_s`'s clock (`query_phase_start`) to start *after* the one-time seed+verify setup completes, so the budget measures actual query latency instead of cold-start archive construction. `total_duration_s` is unchanged (it already comfortably includes setup and wasn't flagged as wrong).
- `polylogue/demo/models.py`, `polylogue/cli/commands/demo.py`: surface `healed_tiers` on `DemoSeedResult` and in the CLI `seed` command's plain-text output.

## Verification

Reproduced fresh (scratch `POLYLOGUE_ARCHIVE_ROOT`, no live archive touched):

```
$ POLYLOGUE_ARCHIVE_ROOT=/realm/tmp/demo-seed-clean/archive polylogue demo seed
Demo archive ready
  ...
$ # rolled index.db's PRAGMA user_version back to 46 on that same demo-owned archive
$ POLYLOGUE_ARCHIVE_ROOT=/realm/tmp/demo-seed-clean/archive polylogue demo seed
Demo archive ready
  ...
  Self-healed: index.db (stale schema moved aside and rebuilt)
  Verify:       polylogue demo verify
$ polylogue demo tour --out-dir /realm/tmp/demo-tour-clean
Polylogue demo tour: passed
  First result: 2.680s
  Full tour:    19.651s
```

- `devtools test tests/unit/demo/ tests/unit/cli/test_demo_command.py tests/unit/storage/test_index_fast_forward_executor.py` — 59 passed.
- New tests:
  - `test_seed_demo_archive_self_heals_a_stale_schema_on_a_demo_owned_root` — rolls an already-seeded demo archive's `index.db` back to an old `PRAGMA user_version` and asserts a second `seed_demo_archive()` call self-heals (`healed_tiers == ("index.db",)`) instead of raising.
  - `test_seed_demo_archive_never_self_heals_a_non_demo_owned_root` — a stale-schema root *without* the demo fixture source directory still raises the original `RuntimeError`; nothing is moved aside. This is the control proving self-heal never touches a foreign/live archive.
  - `test_first_result_budget_excludes_one_time_archive_setup` (new file `tests/unit/demo/test_demo_tour_timing.py`) — injects a real 6s delay around the actual `seed_demo_archive` call and shrinks the budget to 6s; asserts the injected delay does not leak into `first_result_s`.
- **Anti-vacuity confirmed by hand**: temporarily reverting the `tour.py` timer rebase (`first_result_s = time.perf_counter() - start`) made the new tour test fail exactly as expected — `first_result_s=9.5s` against the shrunk 6s budget, with `problems=('first result exceeded 6s budget: 9.505s',)`. Restored the fix before committing.
- `devtools verify --quick`: exit 0 (ruff format/check, mypy --strict, render all --check, layering/closure-matrix/schema-versioning lanes, pre-push hook run automatically on `git push`).

## Acceptance criteria (from polylogue-3ycw)

- (a) "demo seed should detect+auto-rebuild a stale schema version rather than erroring with a manual, underspecified remediation instruction" — **satisfied**, scoped to demo-owned roots only (see Solution above for why an unconditional auto-rebuild would be unsafe).
- (b) "demo tour's budget should exclude one-time fixture/archive setup from the 'first result' timer" — **satisfied**.
- The `demo verify` "0/expected" cascade reported in the bead was a downstream symptom of (a) (verifying a never-actually-seeded archive); no separate fix needed once (a) lands.

Ref polylogue-3ycw

### DIFF
--- a/polylogue/cli/commands/demo.py
+++ b/polylogue/cli/commands/demo.py
@@ -64,6 +64,11 @@ def seed_command(
     if output_format == "json":
         click.echo(json.dumps(payload, sort_keys=True))
         return
+    healed_note = (
+        f"  Self-healed: {', '.join(result.healed_tiers)} (stale schema moved aside and rebuilt)\n"
+        if result.healed_tiers
+        else ""
+    )
     env.ui.console.print(
         "[bold green]Demo archive ready[/bold green]\n"
         f"  Archive root: {result.archive_root}\n"
@@ -71,6 +76,7 @@ def seed_command(
         f"  Sessions:     {result.session_count}\n"
         f"  Messages:     {result.message_count}\n"
         f"  Overlays:     {'yes' if result.overlays_seeded else 'no'}\n"
+        f"{healed_note}"
         "  Verify:       polylogue demo verify"
     )
 
--- a/polylogue/demo/models.py
+++ b/polylogue/demo/models.py
@@ -20,6 +20,7 @@ class DemoSeedResult:
     overlays_seeded: bool
     assertion_count: int
     construct_coverage: tuple[DemoConstructCoverage, ...] = ()
+    healed_tiers: tuple[str, ...] = ()
 
     def to_payload(self) -> dict[str, object]:
         """Serialize the seed result for CLI JSON output and tests."""
@@ -33,6 +34,7 @@ class DemoSeedResult:
             "overlays_seeded": self.overlays_seeded,
             "assertion_count": self.assertion_count,
             "construct_coverage": [row.to_payload() for row in self.construct_coverage],
+            "healed_tiers": list(self.healed_tiers),
         }
 
 
--- a/polylogue/demo/seed.py
+++ b/polylogue/demo/seed.py
@@ -39,7 +39,7 @@ from polylogue.storage.embeddings.identity import (
 )
 from polylogue.storage.embeddings.materialization import archive_embeddable_message_where, message_prose_sql
 from polylogue.storage.insights.session.rebuild import rebuild_session_insights_sync
-from polylogue.storage.sqlite.archive_tiers.bootstrap import initialize_archive_database
+from polylogue.storage.sqlite.archive_tiers.bootstrap import ARCHIVE_TIER_SPECS, initialize_archive_database
 from polylogue.storage.sqlite.archive_tiers.embedding_write import ArchiveEmbeddingWrite, upsert_message_embeddings
 from polylogue.storage.sqlite.archive_tiers.embeddings import EMBEDDING_DIMENSION
 from polylogue.storage.sqlite.archive_tiers.types import ArchiveTier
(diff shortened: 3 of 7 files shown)


Sinity:
Superseded by #3516, which cherry-picks this PR's self-heal commit and adds the archive-root collision guard (polylogue-o3a1t) plus the ownership-manifest hardening (polylogue-wyvio) in one coherent branch. Closing to avoid parallel unmerged branches touching the same functions.

PR #3525, opened 2026-08-01
fix(sources): classify Claude Code message_type from TEXT blocks only

Sinity:
## Summary

Fixes a genuine ingest-time/backfill-time classification divergence in the Claude Code parser that was silently producing new `message_type` drift on every ingest, discovered via bd polylogue-c831's live-archive scan (1,919 candidate rows).

## Problem

`storage/message_type_backfill.py`'s `_message_text_by_id_sql()` reconstructs a message's `classify_text_message_type` input from **only** persisted TEXT-type `blocks` rows (`message_prose_sql(block_types=("text",))`), deliberately excluding thinking/tool_use/tool_result content.

The Claude Code ingest path built a **different, wider** input for the same classifier call: `extract_message_text`/`extract_text_from_segments` (`polylogue/sources/parsers/claude/common.py`) folds THINKING (wrapped `<thinking>...</thinking>`) and TOOL_USE/TOOL_RESULT (JSON-dumped) segment content into one combined string — even though those segments are split into their own separate `ParsedContentBlock` rows before persistence via `content_blocks_from_segments`.

Whenever a THINKING/TOOL_USE segment happened to contain a classifier marker (e.g. a `<file path=` line) that the message's own TEXT block did not carry, ingest-time classification and a later backfill re-run over the persisted TEXT-only blocks silently disagreed. This is a real code-path divergence, not a dated classifier-semantics change — I checked `git log` on `polylogue/archive/message/artifacts.py` and there's no post-materialization semantic change that would explain the drift; the two paths have simply always used different reconstructions of "text". Codex (`extract_codex_text` pulls only `text`/`input_text`/`output_text` fields) and ChatGPT parsers were checked and do not exhibit this pattern.

## Solution

- Added `text_blocks_prose()` (`polylogue/sources/parsers/base_support.py`, re-exported via `base.py`) — the parse-time twin of `message_prose_sql(block_types=("text",))`: joins only TEXT-type block text, in position order, with `\n`.
- Changed `code_parser.py`'s `_message_type_from_code_record` call site to classify from `text_blocks_prose(content_blocks)` (the message's own already-split TEXT blocks) instead of the combined `extract_message_text` string.
- Added a regression test (`tests/unit/sources/test_parsers_claude_code_artifacts.py::test_parse_code_classifies_message_type_from_text_blocks_only`) with a THINKING block carrying a `<file path=` marker plus a plain TEXT reply block. Verified by temporarily reverting the fix: the test fails (`MessageType.CONTEXT != MessageType.MESSAGE`) without it, and passes with it.

**Explicitly not in scope:**
- `classify_material_origin`'s `text=text` argument in `code_parser.py` still receives the combined multi-segment string, which has the same latent-divergence shape for `OPERATOR_COMMAND`/`GENERATED_*_PACK` markers. Left unchanged since the bead's 1,919-row evidence scan was `message_type`-specific, not `material_origin`-specific.
- Retiring the manual `message_type_backfill` maintenance target. The 1,919 already-materialized rows still need one operational backfill run (or the pending `polylogue ops reset --index && polylogued run` reindex) to converge; this PR only stops new drift from accumulating on future Claude Code ingests.

## Verification

- `devtools test tests/unit/sources/test_parsers_claude_code_artifacts.py` — 30 passed (including the new regression test).
- `devtools test tests/unit/sources/ tests/unit/pipeline/test_message_type_backfill.py` — 2222 passed; 7 failures reproduced identically via `git stash` on the pre-change tree (pre-existing/unrelated, none touching `code_parser.py`/`base_support.py`/`base.py`).
- `devtools verify --quick` — format + lint + mypy + render-all-check + layering + lab policies, `exit_code: 0` (also ran again automatically via the pre-push hook).

Ref polylogue-c831

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude Code message classification by evaluating visible text content only, excluding thinking and tool-related content.
  * Prevented messages from being incorrectly categorized based on hidden processing content.

* **Tests**
  * Added coverage to verify classification remains accurate when thinking content includes misleading context markers.


Sinity:
Fixed in d1c08a30f: `message_prose_sql` now runs its own `GROUP_CONCAT` inside a self-contained scalar subquery over a pre-sorted derived table, instead of wrapping `GROUP_CONCAT` around a correlated scalar subquery that every real caller's `LEFT JOIN blocks` fanned out into one row per block (so the scalar only ever returned the first TEXT block). New regression test confirmed to fail on the unmodified code and pass with the fix.

PR #3526, opened 2026-08-01
fix: scope the daemon UDS socket path to the archive root

Sinity:
## Summary

The daemon's Unix domain socket path was derived from `$XDG_RUNTIME_DIR/polylogue/daemon.sock` with no archive-root component, so two `polylogued` instances serving different archives on the same machine collided on the same socket path. This PR keys the socket path off the resolved archive root so that never happens.

## Problem

`polylogue.daemon.uds.daemon_socket_path()` (before this PR) took only an optional `runtime_dir` and always returned `$XDG_RUNTIME_DIR/polylogue/daemon.sock`. `DaemonAPIUnixHTTPServer.__init__` unconditionally unlinks whatever sits at its target path before binding, so a second daemon instance (e.g. a test/dev daemon pointed at a scratch archive) silently steals the production daemon's socket file out from under it. A CLI invocation then reaches whichever daemon most recently bound the shared path, regardless of `POLYLOGUE_ARCHIVE_ROOT`/`--archive-root`.

This was discovered live during PR #3517 perf work (bead polylogue-kadx3): a lane's measurement harness leaked real archive content into its shell output twice before switching to an in-process/mocked-daemon harness.

The bead also asked to investigate whether `--no-daemon` local execution ignores `POLYLOGUE_ARCHIVE_ROOT`. I live-reproduced this against the actual CLI entry point (both with and without `--no-daemon`, using `python3` with the worktree's own source on `PYTHONPATH` per the shared-venv editable-hijack hazard documented in CLAUDE.md) and confirmed the direct/no-daemon path correctly resolves `POLYLOGUE_ARCHIVE_ROOT` in every case tried — it fails looking for the scratch archive's own `index.db`, never falls through to a real archive. This matches the existing extensive archive-root scoping in `polylogue/paths/_roots.py` (polylogue-4ma3, polylogue-o7hx) and the hundreds of existing tests that already exercise `--no-daemon` against isolated per-test archive roots. I found no separate code defect here — the observed leak is fully explained by the socket collision this PR fixes: an agent harness reaching the daemon-probe fallback path (with or without an argv-position mistake around `--no-daemon`) would, pre-fix, have talked to whichever daemon last stole the shared socket path.

## Solution

- New `polylogue/daemon/socket_path.py` — stdlib-only, no archive/storage imports (mirrors the cheap-import constraint already documented on `cli/daemon_client.py`). `archive_scope_key()` computes a stable 12-hex-char SHA-256 key from the resolved (`expanduser().resolve()`) archive root; `daemon_socket_path(archive_root, *, runtime_dir=None)` folds it into `$XDG_RUNTIME_DIR/polylogue/<key>/daemon.sock`.
- `polylogue/daemon/uds.py` now imports and re-exports `daemon_socket_path` from the new module instead of defining its own unscoped version.
- Updated every call site to pass the same archive-root value already used for `/api/health` probe matching, so the path a CLI probe computes always matches the path the corresponding daemon actually bound:
  - `polylogue/daemon/cli.py` — daemon startup, `archive_root_path`.
  - `polylogue/cli/archive_query.py`, `polylogue/cli/click_app.py`, `polylogue/cli/commands/facets.py` — `config.archive_root`, replacing their previous inline `XDG_RUNTIME_DIR`-only path construction (and dropping now-unused `os`/`Path` imports in `facets.py`/`archive_query.py`).
- Registered the new hash call site in `docs/plans/hash-boundary-registry.yaml` (`identifier` classification — the digest is never compared for drift, only used to derive a filesystem path) and regenerated `docs/plans/topology-target.yaml` for the new module.

## Verification

- New `tests/unit/daemon/test_uds_socket_scoping.py` (5 tests): pure `archive_scope_key`/`daemon_socket_path` invariants (distinct per root, stable per root, spelling-insensitive to relative/trailing-slash differences) plus an integration test that starts two real `DaemonAPIUnixHTTPServer` instances scoped to two different archive roots under one shared runtime dir and asserts neither server's socket file or listener is disturbed when the other starts — this reproduces the exact collision the bead reported and confirms it's closed.
- Updated `tests/unit/cli/test_daemon_golden_parity.py` and `tests/benchmarks/test_daemon_uds.py` call sites for the new required `archive_root` parameter.
- `devtools test tests/unit/daemon/test_uds_socket_scoping.py tests/unit/cli/test_daemon_golden_parity.py tests/unit/daemon/test_daemon_cli.py tests/unit/archive/test_facets.py tests/unit/cli/test_click_app.py tests/unit/cli/test_click_app_main.py tests/unit/cli/test_archive_query.py` — 326 passed.
- `devtools verify --quick` — exit 0 (format, lint, `mypy --strict`, `render all --check`, layering, hash-boundary-census, schema-versioning, degrade-loudly, etc.); ran again automatically via the pre-push hook, also exit 0.
- `devtools render topology-projection` — regenerated for the new module.

Ref polylogue-kadx3


Sinity:
Fixed in 7e43be76a: `daemon_socket_path` now falls back to a short, still archive-scoped path (`/tmp/polylogue-<uid>/<key>.sock`) whenever the ordinary `$XDG_RUNTIME_DIR/polylogue/<key>/daemon.sock` path would exceed AF_UNIX's 107-byte usable `sun_path` limit. New regression test reproduces the exact 73-character runtime-dir case from the review comment and asserts the resulting path stays ≤107 bytes and distinct per archive root.

PR #3527, opened 2026-08-01
fix(storage): let a raw rewrite its own accepted revision head

Sinity:
## Summary

Fixes a genuine, confirmed write-precedence defect that silently freezes zero-message `codex-session` rows in the archive, closing polylogue-buq8/i415/lkos. All three beads described the same live-archive symptom (real, substantial raw content but `message_count=0`) with three different, mutually-inconsistent root-cause theories; direct reproduction against the live raw bytes falsifies all three theories and locates the actual defect one layer downstream, in `revision_authority_refuses_write`.

## Problem

**buq8** theorized `revision_authority='quarantined'` blocks materialization entirely. **i415** theorized an old-format rollout envelope the codex parser can't handle. **lkos** theorized a distinct "message-extraction defect" residual from PR #3497. None of these hold up:

- Live-archive query: of 1,757 `codex-session` rows with `revision_authority='quarantined'`, only 15 have `message_count=0` — 1,742 have real content. Quarantine status is orthogonal to whether materialization happened.
- Direct reproduction: running the current `parse_stream_payload` against the exact live raw bytes of buq8/i415's flagship sample (`native_id=0199fada-d8bd-7fc0-997b-d23d3a6849c7`, 3.3 MB) extracts **1,496 real messages**. The current codex parser is not defective for this shape.
- Direct DB inspection: this session's `raw_revision_heads.accepted_raw_id` equals its own `sessions.raw_id`. `raw_revision_heads.decided_at_ms` is **months after** `sessions.updated_at_ms` — a bookkeeping-only backfill retroactively recorded this raw as authoritative long after the original (defective, for reasons this checkout can no longer observe) write happened, without re-running message extraction.

The actual bug: `revision_authority_refuses_write`'s "governed" check refused **every** future write for a session_id once any `raw_revision_heads` row existed for it, without ever comparing which raw the row names as accepted. This meant the very raw a backfill had just declared authoritative could never pass its own gate to correct itself — a session whose content was defectively written before governance existed stays permanently frozen, because no future re-ingest tick can ever get past this check.

## Solution

`polylogue/storage/sqlite/archive_tiers/ingest_precedence.py`: `revision_authority_refuses_write` now compares the incoming `raw_id` against `raw_revision_heads.accepted_raw_id` and refuses only when they differ (a losing/competing raw trying to overwrite the winner). A write for the accepted raw itself — the winner re-asserting its own content — is no longer refused. This is the single shared gate used by both write paths (`ingest_batch/_core.py`'s `_write_session` and `revision_governance.py`'s `_write_parsed_precedence_result`), so the fix applies to both without duplication.

`tests/unit/pipeline/test_ingest_batch.py` adds two regression tests:
- `test_write_session_allows_rewrite_of_its_own_accepted_revision_head` reproduces the exact defect shape (governed session, zero existing messages, same raw_id, real incoming content) and asserts the write now succeeds. Verified to fail (`assert False is True`) against the unmodified production code, then pass after the fix.
- `test_write_session_still_refuses_a_different_raw_than_the_accepted_head` pins that the existing invariant (a genuinely different, losing raw for a governed session_id) stays refused — the fix doesn't reopen last-writer-wins.

## Verification

```
devtools test tests/unit/pipeline/test_ingest_batch.py tests/unit/storage/test_revision_replay.py \
  tests/unit/storage/test_revision_application.py tests/unit/storage/test_raw_revision_authority.py \
  tests/unit/sources/test_live_batch_support.py tests/unit/sources/test_revision_backfill.py \
  tests/unit/storage/test_repair.py tests/unit/storage/test_raw_retention.py \
  tests/unit/storage/test_raw_authority_ledger.py tests/unit/storage/test_quarantined_accepted_raw_repair.py \
  tests/unit/storage/test_duplicate_raw_identity_repair.py tests/unit/storage/test_incremental_rebuild_equivalence.py \
  tests/unit/daemon/test_raw_materialization_parse_stage_equivalence.py \
  tests/unit/storage/test_browser_capture_origin_repair.py tests/unit/cli/test_status.py \
  tests/unit/daemon/test_daemon_status.py tests/unit/maintenance/test_raw_authority_reset.py \
  tests/integration/test_append_cohort_memory.py tests/unit/sources/test_dispatch_payloads.py \
  tests/unit/sources/test_dispatch_ordering.py tests/unit/sources/test_source_laws.py
```
2 new tests pass. Every other failure observed (24 across the raw-authority-ledger/repair/append-cohort-memory cluster, plus 1 pre-existing count-drift in `test_dispatch_payloads.py`'s repeated-`session_meta` fixture) reproduces identically against the unmodified baseline (`git stash` + rerun), confirmed test-by-test — this change adds zero regressions.

```
devtools verify --quick
```
-> exit 0 (ruff format/check, mypy --strict, render all --check, topology/layering/closure-matrix/schema-policy lab checks all pass).

## Residual / follow-ups

This fix unblocks *future* re-ingest of affected sessions but does not retroactively repair already-materialized rows — the live archive's existing zero-message sessions were written before this fix existed and need an ordinary session-scoped reparse (a normal daemon reprocess tick touching those raw_ids, or `polylogue ops reset --index && polylogued run` for a full rebuild) to pick up corrected content. This is **not** a schema/index-structure change, so no `lifecycle.py` `SEMANTIC_REPARSE` declaration applies — it's an application-logic fix in the write-precedence layer, and the repair is an ordinary reprocess, not an index generation bump.

buq8's "materialization never runs" and lkos's "message-extraction defect distinct from the gate" framings do not match what the live data shows; i415's "old rollout envelope" parser theory is also not supported. All three describe the same underlying write-gate defect fixed here, not three independent problems — and this is explicitly **not** the systemic raw-authority absorbing-state issue tracked separately (and still open, unrelated scope) in polylogue-u19l.

Ref polylogue-buq8, polylogue-i415, polylogue-lkos


Sinity:
P2 (arbitrary-row selection under `LIMIT 1`) fixed in ad8d74d96: the refusal check now looks for the existence of any `raw_revision_heads` row whose `accepted_raw_id` differs from the incoming raw, instead of comparing against one arbitrarily selected row. New regression test builds two parallel heads and confirms the write is still refused regardless of row order; verified to fail on the unmodified code and pass with the fix.

P1 (stale `raw_revision_applications` receipts on an accepted-raw rewrite) is a real gap, but it's bigger than a same-lane fix belongs to be — it touches both ordinary write call sites (`revision_governance.py`, `ingest_batch/_core.py`) and the receipt-synthesis contract in `revision_application.py`, and needs its own design for what `decision`/`decided_at_ms` to record for an unrequested-but-legitimate reparse. Tracked as `polylogue-2tfug` rather than rushed into this PR. Not a live-path bug today — `validate_raw_replay_application_receipt` only runs during a formal replay/restart-proof check, not on the ordinary ingest write path — but a genuine latent gap in the authoritative-replay ledger that should get fixed before that check is relied on more broadly.

PR #3560, opened 2026-08-02
fix(daemon): requeue raw parse/validate backlog after interrupted ingest

Sinity:
## Summary

Interrupted ingest attempts now register durable, restart-surviving convergence debt so validation/parse resumes automatically after a daemon restart, instead of waiting for an accidental future touch of the same source path.

## Problem

2026-07-31 acquisition-completeness forensic audit (report: `/realm/inbox/polylogue-audits-2026-07-31/acquisition-completeness.html`, findings F-07/F-08) found 365 claude-code-session union-find groups (2.18 GB) and 588/1004 claude-ai-export conversations acquired but never validated/parsed. Root cause traced to `CursorStore._mark_interrupted_ops_attempts` (`polylogue/sources/live/cursor.py`): when the daemon stops mid-batch, the dangling `ingest_attempts` row is stamped `interrupted` on the next daemon start, and that is the end of the trail. `convergence_debt` (the retry substrate every other convergence stage uses) receives no entry for interrupted ingest work, so the affected source paths wait for an accidental future touch instead of being requeued by the daemon's own convergence loop. A manual reprocess/rebuild recovers the data, but the interrupted-attempt mechanism keeps re-accumulating it: 24 interrupted attempts spanning 2026-07-18 through 2026-07-31, ongoing at audit time.

Ref polylogue-61jg.

I confirmed this is a **different mechanism** than polylogue-hat0 (PR #3552, merged earlier this session). hat0 fixed a live-append quarantine/deferred-cursor loop (`sources/live/append_ingest.py` + `deferred_cursor.py`) where `byte_offset` never advanced for a specific quarantined append range, causing unbounded re-minting of overlapping raw rows on every watcher tick. This bug is a batch/bulk acquisition path (`ingest_attempts` + `raw_sessions` rows with `parsed_at_ms IS NULL`), touches unrelated files, and involves no append/deferred-cursor code.

## Solution

- `CursorStore._mark_interrupted_ops_attempts` now reads each affected attempt's `source_paths_json` (falling back to the legacy single `source_path` column for older rows) before sweeping status to `interrupted`, and records one `raw_parse_recovery` convergence-debt row per distinct path via the existing `record_convergence_debt`/`convergence_debt_retry` machinery -- the same durable, restart-surviving retry substrate every other convergence stage already uses, following this repo's `false_means_pending` pattern.
- New `daemon/convergence_stages.py` stage, `make_raw_parse_recovery_stage`: `check()` cheaply detects raw rows under a path that are acquired but not yet materialized (no index session by raw_id or native-id alias, `parsed_at_ms IS NULL`, no `parse_error`); `execute()` re-drives `storage.repair.repair_raw_materialization` scoped to exactly that path via a new `source_root` parameter (forwarded through `product/raw_authority.repair_materialization`, the CLI/daemon product boundary this repo's layering convention expects). The stage is registered first in `make_default_convergence_stages` so a recovered session feeds straight into FTS/embed/insights in the same convergence pass. `_drain_convergence_debt_once` already dispatches `convergence_debt` rows to stages by name, so no daemon wiring changes were needed beyond registering the new stage.

## Verification

- New `tests/unit/daemon/test_raw_parse_recovery.py` (5 tests): an interrupted attempt registers debt (both the `source_paths_json` path and the legacy `source_path` fallback); the new stage's `check`/`execute` drains a stuck raw row end to end (also asserts `false_means_pending`); a full simulated daemon-restart flow (`begin_ingest_attempt` -> simulated kill -> reopen `CursorStore` -> debt registered -> `DaemonConverger` drains it -> session materializes). Confirmed each test fails without the fix by stashing the production diff and rerunning (ImportError for the missing stage; debt never registered without the cursor.py change).
- `devtools test tests/unit/daemon/test_raw_parse_recovery.py tests/unit/daemon/test_convergence_stages.py tests/unit/daemon/test_daemon_cli.py tests/unit/daemon/test_ingest_worker_handoff.py tests/unit/storage/test_repair.py` -- 235 passed.
- `devtools verify --quick` -- exit 0 (ruff format/check, mypy --strict, render all --check, layering, degrade-loudly, and the rest of the quick gate all green).
- Not run: `devtools verify --all` (full non-integration suite).

### DIFF
--- a/polylogue/daemon/convergence_stages.py
+++ b/polylogue/daemon/convergence_stages.py
@@ -826,6 +826,119 @@ def make_sinex_publication_stage(
     )
 
 
+_RAW_PARSE_RECOVERY_BATCH_LIMIT = 200
+# Mirrors ``storage.repair.RAW_MATERIALIZATION_EXECUTE_BLOB_LIMIT_BYTES`` as a
+# local literal rather than importing it: new surface code (this stage lives
+# in ``daemon/``) should not import substrate (``storage``) internals
+# directly per this repo's layering ratchet, and ``repair_materialization``'s
+# ``max_payload_bytes`` is a plain bound this stage can restate on its own.
+_RAW_PARSE_RECOVERY_MAX_PAYLOAD_BYTES = 1024 * 1024 * 1024
+
+
+def _raw_parse_recovery_pending_count(db_path: Path, path: Path) -> int:
+    """Count raw rows under ``path`` acquired but never materialized.
+
+    Mirrors the non-terminal branch of ``repair.py``'s candidate query at the
+    cheap read-only level this stage's ``check`` needs: no materialized
+    ``sessions`` row for the raw (by raw_id or native-id alias) and no
+    terminal parse error recorded. It intentionally does not replicate the
+    full authority/quarantine/byte-authority classification -- that
+    refinement happens inside ``repair_raw_materialization`` itself during
+    ``execute``; this is only a cheap "is there plausibly pending work here"
+    probe so ``check`` stays fast and false positives just cost one wasted
+    ``execute`` call rather than silently missing real backlog.
+    """
+    source_db = db_path.parent / "source.db"
+    index_db = db_path.parent / "index.db"
+    if not source_db.exists():
+        return 0
+    normalized_root = str(path).rstrip("/")
+    try:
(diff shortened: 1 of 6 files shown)


Sinity:
ack: CodeRabbit's post-head comment is its own rate-limit notice (no findings to triage), not a review finding. Merging.

PR #3574, opened 2026-08-02
fix(storage): collapse byte-equal duplicates before revision-chain proof

Sinity:
## Summary
Fixes the classifier defect identified in the 2026-08-02 authority-dataflow audit (invariants I4/I5): byte-identical duplicate captures of one source were treated as an unprovable fork, quarantining the entire revision cohort even when only a duplicate observation was involved. This is item 1 of 5 in the raw-authority-redesign Phase 1 plan (polylogue-lb39z).

## Problem
`classify_historical_full_revision_streams` (and its eager sibling `classify_historical_full_revisions`, `polylogue/archive/revision_authority.py`) quarantined an entire full-revision cohort the moment any two members tied on size, without ever comparing their bytes. Two byte-identical captures of the same file — the ordinary result of re-acquiring an unchanged source — tie on size and triggered this path. A live, read-only audit measured 13,671 quarantined `raw_sessions` rows (50.2GB, 79% of quarantined bytes) caused by exactly this, and traced a second-order effect: a stuck baseline from this bug lets the live watcher pump unbounded quarantined rows for a single growing file (one Codex rollout produced 799 quarantined snapshots / 6.28GB of overlapping tails, since the append path can't chain onto a quarantined baseline).

Separately, the classifier's ambiguity verdict was cohort-atomic: one genuinely divergent pair anywhere in a cohort quarantined every member, including a proven prefix chain that led up to the fork.

## Solution
Both classifiers now:
1. **Collapse byte-identical revisions (I4)** onto one representative before any chain-order proof runs. The in-memory classifier groups by payload bytes directly; the streaming classifier hashes each stream during its existing single size-verification read pass (no extra I/O). Every non-representative duplicate mirrors its representative's verdict with a new `relation="duplicate"`.
2. **Localize residual ambiguity to the divergent suffix (I5)** instead of the whole cohort. The classifier now builds the same byte-prefix DAG the eager classifier already built internally, but a node is only `BYTE_PROVEN` if it sits on a single, unbranched path back to the cohort's one true root (no ambiguous parent set, no sibling fork anywhere on the path). A shared ancestor with two mutually-incomparable children now classifies the ancestor `BYTE_PROVEN` and only the fork's children `QUARANTINED`, instead of quarantining all three. Two genuinely disconnected/incomparable roots (no shared ancestor at all) still quarantine everything — there is no anchor to localize from, matching prior behavior for that irreducible case.

`classify_untyped_full_revision_groups` (`storage/sqlite/archive_tiers/revision_governance.py`) is updated to check that **every** decision in a cohort is `BYTE_PROVEN` (not just the first), preserving its existing "whole cohort must be one provable chain" contract now that partial-cohort verdicts are possible.

`classify_raw_revision_cohort`'s baseline/generation-number derivation required no changes: duplicate and ambiguous decisions both leave `predecessor_raw_id=None` (matching the pre-existing convention for ambiguous rows), so the existing `children` dict walk used to number generations never collides.

### Alternatives rejected
Promoting the fork's larger/newer-looking child as the accepted head was considered and rejected: it would make an arbitrary size/sort-order tie-break silently pick a winner among genuinely competing evidence, exactly the kind of silent-wrong-guess this subsystem has zero tolerance for (see the `polylogue-yla8` incident history). Leaving the fork tip quarantined for the (separate, still-unwired) judgment-assertion flow to adjudicate is the safe default and matches the report's own recommendation.

## Verification
- `devtools test tests/unit/storage/test_raw_revision_authority.py` — 24 passed, including new anti-vacuity tests that assert the pre-fix classifier quarantined a byte-equal duplicate pair and a shared-root fork wholesale (both parametrized/dedicated tests reproduce the pre-fix bug, then prove the fix resolves it).
- `devtools test -k "raw_revision or revision_governance or raw_authority"` — 115 passed, 6 failed. All 6 failures reproduce identically with this branch's changes reverted (`git stash`), confirming they are pre-existing and unrelated: 5 are `tests/unit/devtools/test_raw_authority_scale_proof.py` synthetic-corpus repair-pass assertions, 1 is `tests/integration/test_raw_authority_daemon_health_proof.py`, a real-daemon-subprocess probe that timed out under host contention.
- `mypy --strict` on both touched modules plus the test file — no issues.
- `devtools lab policy schema-versioning` — intact (no schema touched by this PR).
- `ruff check` + `ruff format --check` — clean.
- Pre-push `devtools verify --quick` gate — passed (ruff format/check, mypy, render all, topology, layering, closure-matrix, schema roundtrip, manifests, CI workflows, doc-commands, degrade-loudly, hash-boundary-census, classifier-fingerprints, schema-promotion-audit).

Ref polylogue-lb39z

## Summary by CodeRabbit

* **New Features**
  * Identical historical revisions are now recognized as duplicates.
  * Quarantine decisions are limited to divergent or uncertain revision branches.
  * Valid byte-growth revision chains remain preserved, including when later forks occur.

* **Bug Fixes**
  * Prevented partially invalid revision groups from being treated as complete linear histories.
  * Improved handling of streamed revisions and same-size, unrelated captures.

### DIFF
--- a/polylogue/archive/revision_authority.py
+++ b/polylogue/archive/revision_authority.py
@@ -112,7 +112,7 @@ class HistoricalRawRevisionStream:
 class HistoricalRevisionDecision:
     raw_id: str
     authority: RawRevisionAuthority
-    relation: Literal["baseline", "predecessor", "ambiguous"]
+    relation: Literal["baseline", "predecessor", "ambiguous", "duplicate"]
     predecessor_raw_id: str | None = None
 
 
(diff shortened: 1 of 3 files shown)


Sinity:
Applied CodeRabbit's finding 1 (quadratic streamed-blob-read amplification in the `maximal` filter) and finding 3 (vacuous test assertion) directly — both verified and fixed in the latest commit.

Finding 2 (mirror a duplicate's `predecessor_raw_id` onto the duplicate decision) is a real observation but the suggested one-line fix would introduce a worse bug: `revision_governance.py`'s generation walk builds `children = {predecessor_raw_id: raw_id, ...}` keyed by predecessor — giving a duplicate the same `predecessor_raw_id` as its representative makes both compete for the same dict key, risking silently dropping the real downstream chain from generation numbering. This needs a properly tested fix, not a rushed one under merge-train time pressure. Filing a follow-up bead rather than applying it here.

PR #3576, opened 2026-08-02
fix(sources): refuse to session-parse unrecognized SQLite databases

Sinity:
## Summary

Tightens the detection boundary so binary SQLite databases can never reach a session parser opportunistically. Adds a shared, magic-byte-based binary-format detector, fixes one extension-only detection bug, adds an explicit refusal at the single shared raw-bytes detection chokepoint, and ships a read-only sweep plus a separate `--apply`-gated actuator to classify already-ingested miscaptures.

## Problem

The 2026-08-02 authority-dataflow audit (`/realm/data/derived/reports/polylogue-authority-dataflow-2026-08-02.html`, "the residue, dissected") found ~450MB of SQLite databases sitting in `raw_sessions` and misread as "genuinely diverged" session revisions by a byte-diff classifier that was never designed to reason about opaque binary snapshots: two Hermes `state.db` copies (135MB each), one `verification_evidence.db`, and Codex `state_5.sqlite` (40MB). Operator directive: "we should be strict. don't try to opportunistically parse any file as session. ofc these should be handled, but these are not sessions."

Live-archive verification against `/realm/db/polylogue` (read-only) confirmed the exact rows the audit named: all sit at `revision_kind='unknown'` / `revision_authority='quarantined'`, meaning they were written by a path that never entered the governance vocabulary rather than a considered decision.

Two concrete detection gaps caused this:

1. `sources/live/batch_support.py`'s `_parse_payload_as_session_artifact` accepted **any** `.db`/`.sqlite`/`.sqlite3` file under a Hermes-tagged source by bare extension match, never checking the actual bytes or schema — its path-based sibling function already did the correct content-verified check, this one didn't.
2. No shared chokepoint positively refused a recognized-binary payload before a JSON/JSONL decode was even attempted; refusal happened only incidentally, via the `JSONDecodeError` a binary blob happens to raise, with no explicit classification.

## Solution

- **`polylogue/core/binary_signatures.py`** (new): single shared, provider-agnostic magic-byte registry (SQLite/PNG/gzip/JPEG/PDF). ZIP is deliberately excluded — ChatGPT/Claude-ai GDPR export bundles are legitimately zip-shaped sessions — but still detectable via `looks_like_zip_bytes` for reporting.
- **`sources/live/batch_support.py`**: `_parse_payload_as_session_artifact` now requires the same content-verified Hermes `state.db`/`verification_evidence.db` schema check (`looks_like_state_db_path`/`looks_like_verification_evidence_db_path`, a real read-only `sqlite3` connection + required-table check) its path-based sibling already uses, instead of a bare extension match.
- **`sources/dispatch.py`**: `detect_provider_from_raw_bytes_evidence` (the documented single shared raw-bytes detection chokepoint, used by production per-file acquisition and the unclaimed-file sweep) now positively refuses a recognized binary signature with explicit evidence, before any JSON decode attempt.
- **`archive/raw_payload/decode.py`**: `build_raw_payload_envelope` (its own docstring: "ingest_record calls this for every raw, including binary ones") now classifies an unrecognized binary-shaped payload as `ArtifactKind.BINARY_DATABASE`/`BINARY_DOCUMENT` (`parse_as_session=False`, `schema_eligible=False`) with an explicit reason, instead of falling through to an incidental decode failure. This runs after the existing Hermes-marker check, so a genuine, content-verified Hermes state.db still wins.
- **`polylogue/storage/binary_artifact_sweep.py`** + **`devtools workspace binary-artifact-sweep`** (read-only) + **`devtools workspace binary-artifact-reclassify-apply`** (`--apply`-gated): mirrors the `raw-live-source-reconciliation` report/apply split. The read-only report reuses `inspect_raw_artifact` as a pure function (no persistence) to classify every `raw_sessions` row by content; the apply step calls the existing, tested `materialize_artifact_observations` to persist `raw_artifacts` rows. Neither touches `revision_authority` or deletes anything — see Verification below for what the apply step would do if run.

### Explicitly out of scope (flagged for an operator decision)

Hermes `state.db`/`verification_evidence.db` already have a dedicated, content-verified, extensively tested session parser (`sources/parsers/hermes_state.py`/`hermes_verification.py`) whose output backs a real, shipped feature (fs1.14: ATIF/ATOF trajectory + verification-ledger topology projection, `insights/hermes_topology_projection.py`, `insights/hermes_verification_coverage.py`). Confirmed live: the two 135MB `state.db` raw rows and the `verification_evidence.db` row the audit found *do* have `parsed_at_ms` set, meaning this dedicated parser already ran on them and materialized derived "sessions" — this is the deliberate, working feature, not the miscapture.

This PR tightens the *detection boundary* (no dedicated, content-verified parser claims a shape → refuse, classify as a binary artifact) without reversing that existing feature. It does **not** decide whether Hermes SQLite content should stop being represented as sessions at all (a literal reading of I11 in the audit report — "sessions hold conversations" — would require it, but reversing an entire shipped, cross-wired subsystem is a materially larger and riskier change than this bead's detection-strictness scope). That is a separate architectural question for the operator; the sweep below reports these rows informationally (`recognized_session_candidates`) so the decision has the data behind it.

## Verification

- `devtools test tests/unit/core/test_binary_signatures.py tests/unit/core/test_raw_payload_decode.py tests/unit/sources/test_dispatch_evidence.py tests/unit/sources/test_live_batch_support.py` — new regression tests pass (magic-byte refusal for both Hermes-tagged and Codex-tagged unrecognized SQLite payloads, plus a regression guard proving a genuine Hermes state.db still classifies `parse_as_session=True`). 3 pre-existing failures in `test_live_batch_support.py`/`test_dispatch_payloads.py` confirmed unrelated via `git stash` (fail identically on master).
- `devtools verify --quick` — green (ruff format/check, mypy, render all, topology, layering, classifier-fingerprints). One classifier-fingerprint drift was expected and explicitly acknowledged (`devtools lab policy classifier-fingerprints --ack ... --reason ...`): `sqlite_snapshot.looks_like_sqlite_bytes` now delegates to the shared detector, byte-for-byte identical behavior, no semantic reparse.
- Read-only live-archive sweep (`POLYLOGUE_ARCHIVE_ROOT=/realm/db/polylogue devtools workspace binary-artifact-sweep --json`), run against the real archive at `/realm/db/polylogue`, strictly read-only: full-archive run (43,124 rows) is still completing in the background at PR-open time due to per-row SQLite-connect overhead; a manual SQL check against `raw_sessions` confirmed the exact rows the audit named (2x Hermes `state.db` at 135,237,632 bytes each, 1x `verification_evidence.db`, Codex `state_5.sqlite`/`goals_1.sqlite`/`memories_1.sqlite`, all `revision_kind='unknown'`). The full sweep's aggregate counts/samples will be appended to this PR and to bead `polylogue-hbtj2`'s notes once it completes — the operator should use that output to decide whether/when to run `--apply`.

## Follow-ups

- Whether Hermes state.db/verification_evidence.db's session-parser output should be reclassified out of session governance entirely (see "Explicitly out of scope" above) — needs an explicit operator decision given the fs1.14 subsystem's size.
- Running `devtools workspace binary-artifact-reclassify-apply --apply` against the live archive once the read-only sweep's findings are reviewed.

Ref polylogue-hbtj2

### DIFF
--- a/devtools/binary_artifact_reclassify_apply.py
+++ b/devtools/binary_artifact_reclassify_apply.py
@@ -1,0 +1,106 @@
+"""Actuator: persist raw_artifacts classification for binary-shaped raw rows.
+
+polylogue-hbtj2: ``devtools/binary_artifact_sweep_report.py`` only classifies
+and reports. This command actually writes ``raw_artifacts`` rows (via
+``polylogue.storage.artifacts.persistence.materialize_artifact_observations``,
+the same durable-observation machinery the schema-validation lane already
+uses) so the flagged rows carry an explicit, queryable "this is a binary
+database, not a session" classification instead of sitting unclassified.
+
+This is scoped deliberately narrow: it upserts ``raw_artifacts`` rows for
+every raw row (idempotent, safe to re-run) but never touches
+``raw_sessions.revision_authority`` or deletes anything. Whether the
+already-quarantined binary rows this bead found should additionally be
+excised from quarantine's revision-authority state machine is a separate
+decision -- see the PR notes this command's output feeds -- and is not
+performed here.
+
+Default mode is dry-run (report only, zero mutation). Pass ``--apply`` to
+actually write ``raw_artifacts`` rows.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sqlite3
+from pathlib import Path
+from typing import TextIO
+
+from polylogue.paths import archive_root as default_archive_root
+from polylogue.storage.artifacts.persistence import materialize_artifact_observations
+from polylogue.storage.binary_artifact_sweep import scan_binary_artifacts
+
+
+def main(argv: list[str] | None = None, *, stdout: TextIO | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--archive-root",
+        type=Path,
+        default=None,
+        help="Archive root to operate on; defaults to the configured active archive root.",
+    )
+    parser.add_argument(
+        "--limit",
+        type=int,
+        default=None,
+        help="Cap the number of raw_sessions rows classified (unbounded by default).",
+    )
+    parser.add_argument(
+        "--apply",
+        action="store_true",
+        help="Actually write raw_artifacts rows. Without this flag, nothing is mutated.",
+    )
+    parser.add_argument("--json", action="store_true", help="Emit the report as JSON.")
+    args = parser.parse_args(argv)
+
+    root = args.archive_root if args.archive_root is not None else default_archive_root()
+    source_db = root / "source.db"
+    if not source_db.exists():
+        print(f"no source.db at {source_db}", file=stdout)
+        return 1
+
+    # Read-only classification pass first, regardless of --apply, so the
+    # operator always sees exactly what would change before any write.
+    ro_conn = sqlite3.connect(f"file:{source_db}?mode=ro", uri=True)
+    try:
+        plan = scan_binary_artifacts(ro_conn, limit=args.limit)
+    finally:
+        ro_conn.close()
+
+    applied = False
+    observation_count = 0
+    if args.apply:
+        rw_conn = sqlite3.connect(source_db)
+        try:
+            observations = materialize_artifact_observations(rw_conn)
+        finally:
+            rw_conn.close()
+        observation_count = len(observations)
+        applied = True
+
+    if args.json:
+        payload = {
+            "applied": applied,
+            "binary_count": len(plan.binary_candidates),
+            "binary_bytes": plan.binary_bytes,
+            "binary_by_kind": plan.by_kind(),
+            "observations_written": observation_count,
+        }
+        print(json.dumps(payload, indent=2, sort_keys=True), file=stdout)
+        return 0
+
+    mode = "APPLIED" if applied else "dry-run (no mutation performed -- pass --apply to write raw_artifacts rows)"
+    print(f"mode: {mode}", file=stdout)
+    print(
+        f"binary-shaped rows found by this bead's stricter detection: {len(plan.binary_candidates)}"
+        f"  ({plan.binary_bytes / (1024 * 1024):.2f} MB)",
+        file=stdout,
+    )
+    if applied:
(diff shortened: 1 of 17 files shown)


Sinity:
## Live-archive sweep results (full run, read-only)

The full read-only sweep against `/realm/db/polylogue` (43,124 `raw_sessions` rows scanned, `POLYLOGUE_ARCHIVE_ROOT=/realm/db/polylogue devtools workspace binary-artifact-sweep --json`) completed:

- **9 rows / 268.32 MB** now correctly classified as `binary_database` (this bead's new, stricter detection) — all `origin=codex-session`: 6× `state_5.sqlite` (40,116,224 bytes each), 1× `goals_1.sqlite`, 1× `memories_1.sqlite`. These previously sat at `revision_kind='unknown'` / `revision_authority='quarantined'` with no artifact classification at all — exactly the miscapture class the audit named.
- **19 rows / 332.62 MB** are `origin=hermes-session` rows correctly left alone: 2× Hermes `state.db` (135,237,632 bytes each), 3× `verification_evidence.db`, plus ATIF trajectory exports and one ATOF event stream — all classified `session_document`/`session_record_stream` via the existing, dedicated, content-verified Hermes parser (the deliberate fs1.14 feature this PR does not touch).
- **Zero** other binary formats found (PNG/JPEG/gzip/PDF) across the full archive — the broader AC5 sweep confirms SQLite is currently the only miscapture class present; the general-purpose `core.binary_signatures` registry stands ready if that changes.

This confirms the fix works exactly as intended: Codex's own state databases (which have no dedicated session parser — `codex_state.py` deliberately never produces sessions from them) are now positively classified as non-session binary artifacts instead of sitting unclassified in quarantine, while Hermes's dedicated SQLite session parser is untouched.

Follow-up for the operator: run `devtools workspace binary-artifact-reclassify-apply --apply` to persist `raw_artifacts` rows for the 9 flagged Codex rows (dry-run reproduces this exact count before writing). This does not touch `revision_authority`; whether/how to also terminally excise these 9 rows from quarantine's revision-authority state machine is a separate decision (see "Explicitly out of scope" in the PR body for the related, larger question about Hermes's 19 rows).

PR #3596, opened 2026-08-03
fix(storage): stop requiring an impossible index-tier backup attestation

Sinity:
## Summary

`agent_meta_sidecar_purge_apply.py` (polylogue-ioz7) and `attachment_reacquisition.py` both require a verified backup manifest covering `index.db` before mutating it, but the validation call unconditionally requires a cryptographic HMAC attestation for that tier -- an attestation the backup system never produces for derived tiers. No backup manifest, however fresh or complete, could ever satisfy it, so `--apply` was permanently broken for both actuators.

## Problem

Discovered running the operator-authorized polylogue-ioz7 live purge against a same-day, all-5-tiers-present, `verdict=success` backup manifest. `polylogue/maintenance/agent_meta_sidecar_purge_apply.py:201` calls `validate_migration_backup_manifest(backup_manifest, ArchiveTier.INDEX, connection=validate_conn)`, which unconditionally requires exactly one `tier="index"` HMAC attestation entry in the receipt's `attestations` list. But `polylogue/daemon/backup.py`'s `_write_successful_verification_receipt` hardcodes `authority_paths` to only ever include `{"source", "user"}` tiers -- by design, matching the durable-tier architecture where only source.db/user.db are durable and cryptographically attestable; index.db/embeddings.db are derived/rebuildable and were never wired for one. `attachment_reacquisition.py` had the identical bug at two call sites.

## Solution

Split `migration_runner.py`'s validation logic into a shared internal helper plus two entry points:
- `validate_migration_backup_manifest` keeps its existing behavior (attestation required) for durable-tier migrations (`migrate_archive_tier`, unchanged).
- New `validate_backup_manifest_covers_derived_tier` skips the attestation check for non-durable tiers while keeping every other guarantee: manifest/receipt shape, tier inclusion, receipt/manifest byte match, and a byte-exact live fingerprint (sha256/size/user_version) recomputed from the current on-disk file.

Both actuators now call the new function for their `index.db` check. Durable-tier callers (raw_authority.py, raw_append_chain_backfill_apply.py, raw_membership_writeback_apply.py, raw_live_source_reconciliation_apply.py) are untouched -- they correctly pass `ArchiveTier.SOURCE`.

## Anti-vacuity

Every existing test for both actuators monkeypatched the validator entirely, so none of them ever exercised a manifest actually produced by `polylogue ops backup --profile full_evidence --verify` -- exactly the gap that hid this bug. Added `test_apply_accepts_a_real_backup_manifest_from_ops_backup`, which builds the fixture archive, runs the real `backup_archive()` entry point against it, and lets `apply_agent_meta_sidecar_purge` validate that real manifest with no monkeypatching.

## Verification

- `devtools test tests/unit/maintenance/test_agent_meta_sidecar_purge_apply.py tests/unit/storage/test_attachment_reacquisition.py` -- 14 passed.
- `devtools test` on the untouched durable-tier siblings (raw_authority, raw_append_chain_backfill_apply, raw_membership_writeback_apply, raw_live_source_reconciliation_apply, durable_migrations) -- 55 passed, confirming the durable-tier path is unaffected.
- `mypy --strict` on all 5 touched files -- clean.
- `devtools verify --quick` (pre-push gate) -- clean.

Ref polylogue-5kmn7, polylogue-ioz7

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of rebuildable index backups during purge and attachment recovery operations.
  * Added comprehensive checks for backup coverage, inventory, artifacts, receipts, and live fingerprints.
  * Durable backup validation continues to require receipt authentication.

* **Tests**
  * Added regression coverage confirming valid backups allow purging only the intended sidecar sessions and record receipts.
  * Updated recovery and maintenance validation tests.

### DIFF
--- a/polylogue/maintenance/agent_meta_sidecar_purge_apply.py
+++ b/polylogue/maintenance/agent_meta_sidecar_purge_apply.py
@@ -54,7 +54,7 @@ from polylogue.storage.agent_meta_sidecar_sweep import (
 )
 from polylogue.storage.sqlite.archive_tiers.archive import ArchiveStore
 from polylogue.storage.sqlite.archive_tiers.types import ArchiveTier
-from polylogue.storage.sqlite.migration_runner import validate_migration_backup_manifest
+from polylogue.storage.sqlite.migration_runner import validate_backup_manifest_covers_derived_tier
 
 TOOL_VERSION = "agent-meta-sidecar-purge-apply-v1"
 
@@ -198,7 +198,7 @@ def apply_agent_meta_sidecar_purge(
     validate_conn = sqlite3.connect(str(index_db), uri=True)
     try:
         _checkpoint_live_tier(validate_conn)
-        validate_migration_backup_manifest(backup_manifest, ArchiveTier.INDEX, connection=validate_conn)
+        validate_backup_manifest_covers_derived_tier(backup_manifest, ArchiveTier.INDEX, connection=validate_conn)
         plan = scan_agent_meta_sidecar_sessions(validate_conn, source_db, limit=limit)
     finally:
         validate_conn.close()
--- a/polylogue/maintenance/agent_meta_sidecar_purge_apply.py
+++ b/polylogue/maintenance/agent_meta_sidecar_purge_apply.py
@@ -216,6 +216,21 @@ def apply_agent_meta_sidecar_purge(
     if session_ids:
         store = ArchiveStore(archive_root, read_only=False)
         try:
+            # Authoritative re-validation now that the writer lease is held
+            # (ArchiveStore.__init__ acquires it synchronously above) --
+            # matches attachment_reacquisition.py's / migrate_archive_tier's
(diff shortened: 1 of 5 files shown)


Sinity:
Fixed in 0af2eff52: added the authoritative revalidation right after ArchiveStore acquires its exclusive writer lease (ArchiveStore.__init__ acquires it synchronously for read_only=False), before delete_sessions runs -- matching attachment_reacquisition.py's existing precheck+revalidation pattern (which is where I copied it from). Added a concurrency regression test (fake validator raises on its second call) proving the revalidation actually aborts the apply before any mutation.

Left out of scope: your point about reordering BEGIN IMMEDIATE ahead of the validation in attachment_reacquisition.py itself. That's a pre-existing design choice in code this PR doesn't otherwise touch, not something the attestation bug this PR fixes introduced.

PR #1787, opened 2026-06-06
refactor: current split-file archive as the sole storage architecture

Sinity:
## Summary

Makes the split-file archive (`source.db` / `index.db` / `embeddings.db` /
`user.db` / `ops.db`) the only supported Polylogue storage runtime, per #1743.
Removes the old single-file archive paths and compatibility framing, migrates
storage/query/runtime/test/doc surfaces to the current schema and `origin` /
`native_id` / `session_events` / `actions` vocabulary, and lands the full #1743
scope — including the parity/structural tail that was briefly tracked as
separate `post-#1743` issues and has since been folded back into #1743's
Pulled-In Scope section.

Closes #1743.

## Problem

The current-archive runtime had landed at the DDL level, but large parts of the
Python runtime, query layer, and test suite still spoke the pre-split schema:
old physical table/column names (`source_name`, `provider_meta`, `parsed_at`,
`acquired_at`, `source_file_cursor`, `artifact_observations`, old-shape
`pending_blob_refs`), single-DB read/write paths, and dead legacy modules. mypy
`--strict` flagged 112 inconsistencies and the suite had 183 runtime failures —
many from production code reading the wrong tier or stale columns, not just test
drift. Beyond that initial conformance pass, a tail of completion work remained:
two parallel insight-rebuild stacks, native read-surface features the old
runtime had (auto→hybrid elevation, readiness degraded/legacy taxonomy), and a
`provider_meta` escape hatch still on the parser-output models.

## Solution

- **Remove obsolete surfaces:** the dead legacy user-state half of
  `sessions_identity.py` (the `repoint_user_state_by_identity` machinery is
  obsolete now that `session_id = origin:native_id` is deterministic), the
  single-DB `hook_paste_enrichment` fallback, `action_event*`/`provider/events`
  packages, the provider-meta-policy verifier, and the dead `RepositoryWriteMixin`
  user-state stack (the live path routes through `ArchiveStore`).
- **Cross-tier reads:** both the async cached path and the one-shot sync
  `open_connection` now ATTACH the sibling tiers to `index.db` connections, so
  unqualified cross-tier names (`raw_sessions`, `blob_refs`, `pending_blob_refs`)
  resolve (SQLite main-wins keeps index tables unaffected). This was the
  keystone fix for ~50 "no such table: raw_sessions" failures.
- **Column/tier migrations:** raw query layer (`origin` + INTEGER-ms columns,
[...]

PR #3697, opened 2026-08-03
feat(storage): dedupe unindexed byte-identical raw quarantine groups

Sinity:
## Summary
Adds a permanent archive-verification check plus a one-shot devtools actuator for the raw-authority gap measured in polylogue-zm4w8: quarantined `raw_sessions` rows that are pure byte-identical duplicates of another still-quarantined row sharing the same `source_path`, with no indexed twin anywhere -- a population `raw-byte-duplicate-supersession-apply` cannot see because its own classifier requires an already-indexed twin.

## Problem
polylogue-zm4w8 measured (2026-08-03, read-only) 1,777 quarantined `codex-session` raw_sessions rows (22.19 GB) that are redundant re-acquisitions of files already captured by another still-quarantined row -- e.g. one rollout file with nine separate byte-identical `raw_id` rows, all `revision_kind='unknown'`, `revision_authority='quarantined'`. Confirmed live: zero of these duplicate `blob_hash` values have any non-quarantined twin anywhere in `raw_sessions`, so the existing actuator (whose classifier's universe is exactly "quarantined rows with an indexed twin") returns zero candidates for this class by construction. This is polylogue-lkrc's simplified scope in its first concrete execution: reclassify the quarantine backlog, verified by the test suite, no manual sorting.

## Solution
- `polylogue/storage/raw_quarantine_group_dedup.py`: read-only classifier. Groups quarantined `raw_sessions` rows by `(source_path, blob_hash)`, flags groups with >1 member, and excludes any group whose `blob_hash` already has an indexed twin or a non-quarantined member anywhere (that's `raw-byte-duplicate-supersession`'s territory, not this gap's). The lowest `raw_id` in a group is the deterministic representative.
- `polylogue/maintenance/raw_quarantine_group_dedup_apply.py` + `devtools/raw_quarantine_group_dedup_apply.py` (wired as `devtools workspace raw-quarantine-group-dedup-apply`, `--help` verified registered): the actuator. Dry-run by default; `--apply` requires `--backup-manifest`. Unlike every sibling actuator in this family, this one is genuinely two-phase because it must actually materialize content, not just flip a column: phase one promotes exactly one representative raw per group through the real production ingest pipeline (`ParsingService.parse_from_raw` -> `write_parsed_session_to_archive` -> `refresh_session_insights_bulk`), re-verified by reading `index.db` for the resulting `sessions` row rather than trusted from the ingest call's return value; phase two, only once materialization has genuinely landed, marks the rest of the group `revision_authority='byte_proven'` inside a single locked `source.db` transaction with an immutable receipt (`raw_quarantine_group_dedup_receipts`, migration 025) pointing at the representative's `raw_id`/`session_id`. If materialization produces no indexed session (parse error, refused write, non-session content), that whole group is left untouched rather than guessed at. Never deletes blobs or runs GC/VACUUM.
- `polylogue/maintenance/archive_verification.py`: registers `raw-quarantine-group-dedup` in `ARCHIVE_VERIFICATION_CHECKS` (the t0m73 registry pattern), reusing the same classifier module directly so the check and the actuator can never drift apart. This is the permanent, ongoing integrity check the bead calls for -- ERROR when any qualifying group exists, part of the ordinary `ArchiveVerificationReport`.

**Non-obvious decision**: the bead's own design text says "mark the rest `revision_kind='duplicate'`/`revision_authority='superseded'`" -- neither value exists in the schema's closed CHECK vocabularies (`revision_kind` is `full`/`append`/`unknown`; `revision_authority` is `asserted`/`byte_proven`/`quarantined`), and widening either requires a full `raw_sessions` table rebuild (migration 021's precedent). Instead this reuses `revision_authority='byte_proven'` -- the exact mechanism `raw-byte-duplicate-supersession` already established for "quarantined, now proven byte-identical to something real" -- with the actual evidence (which representative raw/session) recorded in a dedicated receipt table, exactly mirroring that actuator's own precedent for the same design tension.

## Root cause (bead item 3)
Investigated read-only against the live archive (`/realm/db/polylogue`, never mutated). Every fully-quarantined byte-identical duplicate group's **latest** member was acquired on 2026-07-19; zero such groups have any member acquired after that date, even though ordinary ingestion has continued and produced other kinds of raw rows as recently as 2026-07-31 (4 days before this session). This is strong evidence the bug is **historical, not live-recurring**: `deterministic_raw_session_id` (`storage/sqlite/archive_tiers/source_write.py`, present since the #1787 split-file rearchitecture) derives `raw_id` from `(origin, source_path, source_index, blob_hash)`, so a byte-identical re-acquisition at a stable `source_index` should already collapse to the same `raw_id` rather than mint a new one -- consistent with new duplicate groups no longer appearing. I could not find the exact commit that closed this gap, but the sharp cutoff at 2026-07-19 with continued ingestion afterward is direct evidence against it still recurring for freshly-acquired codex sessions. If the coordinator wants a more precise root-cause commit identification, that's a narrow follow-up, not a blocker for this bead's three-step scope.

## Verification
- `devtools test tests/unit/storage/test_raw_quarantine_group_dedup.py tests/unit/maintenance/test_raw_quarantine_group_dedup_apply.py tests/unit/maintenance/test_archive_verification.py` -> **79 passed** (includes a full happy-path test that materializes a representative raw through the real production ingest pipeline).
- `devtools test tests/unit/storage/test_durable_migrations.py` -> **43 passed** (schema-version-agnostic assertions absorbed the migration 025 bump cleanly).
- `python3 -m mypy --strict` on every touched module (classifier, actuator, devtools CLI, archive_verification.py, source.py, command_catalog.py, and all three test files) -> clean.
- `devtools verify --quick` -> exit 0 (format + lint + mypy + render-all-check + layering + closure-matrix + schema/lab policy checks), also ran automatically by the pre-push hook.
- `devtools workspace raw-quarantine-group-dedup-apply --help` -> confirms CLI registration.
- Read-only dry-run against the live archive (`POLYLOGUE_ARCHIVE_ROOT=/realm/db/polylogue devtools workspace raw-quarantine-group-dedup-apply --json`, no mutation): `group_count=1822`, `marked_duplicate_count=1837` rows (8.22 GiB reclaimable) -- close to but not identical to the bead's original 2026-08-03 measurement, consistent with concurrent archive activity since then.
- Read-only run of the new check against the live archive confirms it fires `error` with matching evidence, proving the finding before any fix, per the bead's own step 3.
- **`--apply` was intentionally never run against the live archive** -- left for the coordinator to review the dry-run report and execute with a verified backup manifest, per explicit dispatch instruction.

Ref polylogue-zm4w8

## Summary by CodeRabbit

* **New Features**
  * Added a command to identify and deduplicate fully quarantined, byte-identical raw sessions.
  * Supports dry-run reports, JSON output, group limits, and protected apply mode with verified backups.
  * Promotes one representative, marks duplicates as resolved, and records durable operation receipts.
* **Bug Fixes**
  * Archive verification now detects unresolved quarantined duplicate groups.
* **Documentation**
  * Added the command to the developer tools documentation and command catalog.
* **Tests**
  * Added coverage for planning, verification, dry runs, backups, promotion, receipts, and no-op scenarios.

### DIFF
--- a/devtools/command_catalog.py
+++ b/devtools/command_catalog.py
@@ -1091,6 +1091,34 @@ COMMAND_SPECS: tuple[CommandSpec, ...] = (
         ),
     ),
     CommandSpec(
+        "workspace raw-quarantine-group-dedup-apply",
+        "workspace",
+        "Promote one representative raw per fully-quarantined byte-identical (source_path, blob_hash) group.",
+        "devtools.raw_quarantine_group_dedup_apply",
+        use_when=(
+            "polylogue-zm4w8 (measured live 2026-08-03): 1,777 raw_sessions rows (22.2 GiB) among "
+            "the codex-session quarantine backlog are pure redundant duplicates -- same source_path "
+            "AND same blob_hash as another raw_sessions row -- where EVERY member of the group is "
+            "still quarantined (no indexed twin anywhere), invisible to "
+            "raw-byte-duplicate-supersession-apply (which only matches a quarantined raw against an "
+            "already-INDEXED twin). Default is dry-run; --apply requires --backup-manifest pointing "
+            "at a verified source-tier backup (polylogue backup --output-dir <dir> --verify). "
+            "Materializes exactly one representative raw per group through the real ingest pipeline "
+            "(ParsingService.parse_from_raw -> write_parsed_session_to_archive -> "
+            "refresh_session_insights_bulk) so it becomes a genuine indexed session, then marks the "
+            "rest revision_authority='byte_proven' with an immutable per-row receipt "
+            "(raw_quarantine_group_dedup_receipts) pointing at the promoted representative's raw_id "
+            "and new session_id. Never deletes blobs or runs GC/VACUUM -- those are separate, later "
+            "steps."
+        ),
+        examples=(
+            "devtools workspace raw-quarantine-group-dedup-apply",
+            "devtools workspace raw-quarantine-group-dedup-apply --json",
+            "devtools workspace raw-quarantine-group-dedup-apply --apply "
+            "--backup-manifest /realm/staging/polylogue-backup/manifest.json",
+        ),
+    ),
+    CommandSpec(
         "workspace binary-artifact-sweep",
         "workspace",
         "Find raw_sessions rows whose bytes are a non-session binary format (SQLite, etc).",
--- a/devtools/raw_quarantine_group_dedup_apply.py
+++ b/devtools/raw_quarantine_group_dedup_apply.py
@@ -1,0 +1,136 @@
+"""Actuator: promote one representative raw per fully-quarantined byte-identical group.
+
+polylogue-zm4w8: 1,777 raw_sessions rows (22.2 GiB, measured 2026-08-03) among
+the codex-session quarantine backlog are pure redundant duplicates -- same
+``source_path`` AND same ``blob_hash`` as another raw_sessions row, with
+every member of the group still quarantined (no indexed twin anywhere) --
+invisible to ``raw-byte-duplicate-supersession-apply``, which only matches a
+quarantined raw against an already-INDEXED twin.
+
+Default mode is dry-run (report only, zero mutation). Pass ``--apply`` to
+actually promote a representative per group and mark the rest, which
+additionally requires ``--backup-manifest`` pointing at a verified backup
+manifest for the ``source`` tier (see ``polylogue backup --output-dir <dir>
+--verify``). This never runs blob GC or ``VACUUM`` -- that is a separate,
+later, operator-invoked step.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+from pathlib import Path
+from typing import TextIO
+
+from polylogue.maintenance.raw_quarantine_group_dedup_apply import (
+    RawQuarantineGroupDedupApplyError,
+    apply_raw_quarantine_group_dedup,
+)
+from polylogue.paths import archive_root as default_archive_root
+
+
+def main(argv: list[str] | None = None, *, stdout: TextIO | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--archive-root",
+        type=Path,
+        default=None,
+        help="Archive root to operate on; defaults to the configured active archive root.",
+    )
+    parser.add_argument(
+        "--limit",
+        type=int,
+        default=None,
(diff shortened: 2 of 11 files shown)


Sinity:
Addressed all CodeRabbit findings (5 the coordinator listed, plus 2 additional Major findings surfaced by re-fetching the full review thread -- fixing those too since they were real correctness gaps):

1. **Multi-session representative (Major)**: `_materialize_group_representative` now orders by `session_id` and requires exactly one materialized session; zero or more than one both leave the whole group untouched (documented, deliberate invariant -- a multi-session raw doesn't fit the receipt schema's singular `representative_session_id` without inventing ambiguous semantics). Covered by a new test that empirically verifies (via a real ingest run) that a grouped Claude Code JSONL raw genuinely materializes 2 sessions before asserting the group stays untouched.
2. **`limit=0` dishonored (Major)**: fixed the check-after-append off-by-one in `plan_raw_quarantine_group_dedup`; new `test_limit_zero_returns_no_groups` regression test.
3. **Stale `v23` comment (Minor)**: fixed to `v25`.
4. **Migration comment count mismatch (Minor)**: reconciled -- now explicitly labels the bead's original filing-time measurement (1,777 rows/22.2 GiB) vs. this PR's later same-day live re-measurement (1,822 groups/1,837 rows/8.22 GiB).
5. **Untested safety branches (Trivial)**: added tests for both -- representative-produces-no-session, and phase-two rollback on a receipt CHECK-constraint violation.

Also fixed two additional Major findings from the same review thread the initial summary didn't enumerate:
- **WAL checkpoint busy flag ignored**: `_checkpoint_live_tier` only rejected a missing row, not `busy=1` (blocked, WAL not truncated) -- now rejects both, with a test.
- **Phase-one failure left already-materialized groups permanently unreachable**: restructured from "materialize every representative, then batch-mark every group's duplicates" to per-group materialize-then-mark, so an exception on group N no longer discards groups 1..N-1's already-durable work, and groups N+1 onward still get attempted (a per-group exception is now caught, logged, and skipped rather than aborting the whole run).

Commits: 1f441606e (limit=0), 747cb0a81 (WAL busy-flag + per-group materialize/mark + multi-session invariant), d04fafd92 (comment fixes).

Verification: `devtools test tests/unit/storage/test_raw_quarantine_group_dedup.py tests/unit/maintenance/test_raw_quarantine_group_dedup_apply.py tests/unit/maintenance/test_archive_verification.py` -> 84 passed. `mypy --strict` clean on all touched files. `devtools verify --quick` -> exit 0. Merge-gate receipt re-recorded at head `d04fafd92`.

PR #3718, opened 2026-08-04
feat(maintenance): quarantine invalid blob namespace entries

Sinity:
## Summary

Adds `polylogue ops maintenance blob-namespace-quarantine`, a dry-run-first filesystem-only actuator for entries classified outside the canonical blob namespace. Ref polylogue-84ake.

## Problem

Historical SQLite sidecars and abandoned publication files can remain under `blob/`. They must be removed from the live namespace without treating canonical blobs, canonical orphans, GC, or SQLite rows as cleanup targets.

## Solution

The maintenance module fully re-hashes canonical blobs, classifies only invalid namespace entries, and moves the exact rescanned candidate set with same-filesystem `os.replace()` into a sibling quarantine tree. Apply requires the daemon stopped, no writer lease, an exclusive maintenance lease, a clean source WAL checkpoint, an attested source backup that matches live identity, a new explicit receipt directory, path and symlink containment checks, and no pre-existing destination. Immutable before and after receipts include complete inventory digests, source and backup fingerprints, destination tree hashes, and postflight verification. Recovery only classifies committed, rolled-back, or indeterminate state.

## Production preconditions

Run a verified source-tier backup, stop `polylogued`, ensure all archive writers are stopped, choose a new receipt directory outside `blob/`, run the dry run, then pass that exact backup manifest and receipt directory with `--apply`. Run `polylogue ops doctor --blob-integrity --full --format json` after a production pass. The actuator never deletes, invokes GC, changes SQLite rows, moves canonical blobs, follows symlinks, or shells out for filesystem deletion.

## Verification

`direnv exec . devtools test tests/unit/maintenance/test_blob_namespace_quarantine.py` passed: 7 tests cover mixed namespace classification, canonical-hash and backup-identity blockers, daemon/writer/WAL/containment/destination gates, immutable receipts, recovery states, and the real Hermes SQLite snapshot route. `direnv exec . devtools test tests/unit/storage/test_blob_store.py tests/unit/storage/test_blob_integrity.py tests/unit/sources/test_parsers_local_agent.py tests/unit/cli/test_maintenance_registration.py` passed: 103 tests. `direnv exec . devtools render cli-reference --check` reported `sync OK`. `direnv exec . devtools verify --quick` passed format, lint, mypy, generated surfaces, layering, schema, manifests, and CI workflow checks.

### DIFF
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -113,6 +113,50 @@ Heuristics:
 
 ## Subcommands
 
+### `polylogue ops maintenance blob-namespace-quarantine`: offline filesystem quarantine
+
+Read-only by default. This is the actuator for entries that
+`BlobStore.iter_namespace()` classifies outside the canonical
+`blob/<2-hex>/<62-hex>` layout: historical SQLite `-wal`/`-shm` sidecars,
+stranded `.blob.*` files, malformed shards, and non-regular entries. It does
+not classify orphaned canonical blobs and never runs GC.
+
+```bash
+polylogue ops maintenance blob-namespace-quarantine --output-format json
+polylogue ops maintenance blob-namespace-quarantine --apply \
+  --backup-manifest /path/to/verified-source-backup/manifest.json \
+  --receipt-dir /path/to/new/namespace-quarantine-receipt \
+  --output-format json
+polylogue ops maintenance blob-namespace-quarantine --recover \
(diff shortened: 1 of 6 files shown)


Sinity:
Superseded by #3720, which carries the verified blob-namespace quarantine actuator on current master without rewriting this branch.

PR #3736, opened 2026-08-04
fix(storage): expose unpaired action result state

Sinity:
## Summary

Classifies the corrected 49,886 identity-bearing unpaired tool calls and adds a typed action-result state that preserves absent results separately from existing results with an unreported outcome.

## Problem

The actions read model represented both a missing `tool_result` row and a paired row with null outcome signals through the same nullable fields. Readers could not count or reason about unpaired calls without conflating those two evidence states.

## Solution

Adds `ActionResultState` with `no_result`, `outcome_unknown`, and `outcome_reported`; projects it from the canonical actions view and its session-bounded relation; and carries it through `ArchiveActionQueryRow`. Index schema v64 is a `VIEW_ONLY` fast-forward that replaces the derived view and does not reparse raw data or mutate source/blob tiers. The audit records 33,994 Codex, 15,423 ChatGPT, 466 Claude Code, and 3 Claude export identity-bearing no-result calls. No parser pairing defect was established, so the remainder is preserved as named evidence rather than reconstructed heuristically.

## Verification

- `./.venv/bin/python -m devtools test tests/unit/storage/test_archive_tiers_ddl.py::test_actions_view_distinguishes_absent_result_from_unknown_outcome tests/unit/storage/test_archive_tiers_ddl.py::test_actions_view_never_cross_pairs_empty_string_tool_id tests/unit/storage/test_archive_tiers_archive.py::test_archive_facade_exposes_distinct_action_result_absence_states` returned `3 passed`.
- Pre-push `devtools verify --quick` passed: format, lint, mypy, render, layering, schema roundtrip, policy checks, and demo-tour freshness.
- A fresh-worktree `devtools verify --seed-testmon --skip-slow` could not seed because the guard rejects the seed command's own unmarked `.cache/testmon/seed-attempt.json`; no guard was bypassed.

Ref polylogue-cuxz.5.

### DIFF
--- a/devtools/docs_surface.py
+++ b/devtools/docs_surface.py
@@ -432,6 +432,12 @@ DOCS_REFERENCE_ENTRIES: tuple[DocsEntry, ...] = (
     _entry(
         "Race Window Audit", "audits/2026-07-09-race-window-audit.md", "Race-window investigation record.", "archive"
     ),
+    _entry(
+        "Unpaired Tool Results Classification",
+        "audits/2026-08-04-unpaired-tool-results.md",
+        "Classification of unpaired tool-call evidence and action result states.",
+        "archive",
+    ),
     _entry("Audit Record Index", "audits/README.md", "Index of dated investigation records.", "archive"),
     _entry(
         "1498 Cascade Retrospective",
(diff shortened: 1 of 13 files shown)


Sinity:
This PR is superseded by merged PR #3799. Current master already contains the action-result presence state, ranked action relation, session-wide pairing, public result-state model, schema fast-forward, and the associated real-route tests. I am closing this stale draft so it does not remain an ambiguous second implementation.

PR #3738, opened 2026-08-04
fix(maintenance): gate reindex on blob-ref liveness

Sinity:
## Summary

Make source-tier blob-reference liveness a direct rebuild preflight gate, align attachment verification with the canonical parent raw-session referent, and record the live I3 closure boundary.

## Problem

PR #3705 added the guarded reconciliation actuator, but the archive verifier still treated attachment refs as pointing to the obsolete `raw_artifacts` table. The direct reindex route did not run I3 before creating rebuild lifecycle state. A fresh read-only I3 census found `attachment=0` and `raw_payload=73,427`, so an honest blob-pristine or reindex-ready claim is not possible.

## Solution

`archive_verification` now derives every referent mapping from the source-tier canonical liveness tuple. The rebuild entrypoint verifies `blob-refs-liveness` at the archive root before acquiring the rebuild lease or creating a generation. The liveness actuator now refuses any running `polylogued` before validating a backup, creating a receipt, or deleting a `blob_refs` row. The audit documents the result and the required future operator action.

| Requirement | Status |
| --- | --- |
| Parent raw-session attachment referent | Satisfied by the shared canonical mapping and a real source-table regression test. |
| Direct reindex gate | Satisfied before generation lifecycle creation. |
| Offline source-tier apply | Satisfied by an explicit live-daemon refusal test. |
| Live cleanup | Intentionally not performed. The remaining 73,427 raw-payload refs keep the gate red. |

No blob namespace quarantine, generic GC algorithm, raw-authority behavior, source backup, or production archive data changed.

## Verification

- `direnv exec . devtools test tests/unit/storage/test_blob_ref_liveness.py tests/unit/storage/test_blob_gc_ref_type_liveness.py tests/unit/cli/test_blob_ref_liveness_cli.py tests/unit/maintenance/test_archive_verification.py tests/unit/maintenance/test_rebuild_index_ownership.py` produced `97 passed`.
- `direnv exec . mypy` produced `Success: no issues found in 2591 source files`.
- `direnv exec . ruff check ...` produced `All checks passed!`.
- `direnv exec . devtools render all --check` reported every generated surface `sync OK`.
- Read-only live I3 SQL against `/realm/db/polylogue/source.db` returned `attachment|0` and `raw_payload|73427`.

## Adversarial review notes

One completed independent read-only review found that apply tests mocked the offline guard. This PR adds the concrete daemon refusal test. Follow-up source tracing found the stale attachment referent mapping and missing rebuild preflight, both covered by real source-table and public rebuild-entrypoint tests.

Ref polylogue-0v4tn

### DIFF
--- a/devtools/docs_surface.py
+++ b/devtools/docs_surface.py
@@ -432,6 +432,12 @@ DOCS_REFERENCE_ENTRIES: tuple[DocsEntry, ...] = (
     _entry(
         "Race Window Audit", "audits/2026-07-09-race-window-audit.md", "Race-window investigation record.", "archive"
     ),
+    _entry(
+        "Blob-Reference Liveness Closure Audit",
+        "audits/2026-08-04-blob-ref-liveness-closure.md",
+        "I3 live evidence, source-tier reconciliation safeguards, and the direct-reindex gate.",
+        "archive",
+    ),
     _entry("Audit Record Index", "audits/README.md", "Index of dated investigation records.", "archive"),
     _entry(
         "1498 Cascade Retrospective",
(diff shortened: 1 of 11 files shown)


Sinity:
Closing as superseded by merged PR #3742. A fresh-master audit confirmed #3742 carries the same I3 blob-reference liveness gate while preserving newer candidate corpus-fidelity protections.

PR #3755, opened 2026-08-04
fix(devtools): bound Beads state-report tree rendering

Sinity:
## Summary

Bounds the generated Polylogue Beads state-report tree when its report input contains repeated parent-child links or a cycle.

## Problem

The state-report renderer recursively followed its in-memory parent-child map without a path or edge guard. Corrupt or hostile report input could therefore produce excessive HTML or recurse forever.

This is a defensive hardening change for Polylogue's generated HTML report only. It does not repair the `bd list --all` CLI renderer. That production Beads CLI defect is being handled in a separate upstream checkout and PR.

## Solution

`devtools/beads_state_report.py` now suppresses a repeated parent-child edge or a branch that revisits an ancestor, and renders one compact diagnostic in its place. The report keeps the existing output for ordinary acyclic hierarchies.

## Verification

`devtools test tests/unit/devtools/test_beads_state_report.py`

Passed: `15 passed`.

`devtools verify --quick`

Passed: `23 checks passed`.

### DIFF
--- a/devtools/beads_state_report.py
+++ b/devtools/beads_state_report.py
@@ -89,6 +89,7 @@ RECONCILIATION_WINDOW = 1200
 STALE_CLAIM_DAYS = 7  # in_progress untouched this long = zombie claim
 AGED_URGENT_DAYS = 14  # open P0/P1 older than this = urgency not being consumed
 VELOCITY_WINDOW_DAYS = 14  # trailing window vs the window before it
+TREE_MAX_DEPTH = 2  # structure view renders two parent-child levels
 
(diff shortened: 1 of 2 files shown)


Sinity:
Closing this draft because independent review established that it changes Polylogue's report renderer, while bead polylogue-2bc2 requires the actual Beads CLI tree/list route and its large-graph behavior. The reindex canary-evidence gate has taken the active repair slot. The branch is retained for any useful renderer-only work, but this PR does not satisfy the bead.

PR #3784, opened 2026-08-05
fix(sources): classify sidecars before live session admission

Sinity:
## Summary

Keep declared metadata and workflow sidecars as durable raw evidence instead of materializing them as sessions.

## Problem

The validation-off worker shortcut and full/append live ingestion could override path-declared artifact classification, allowing session-shaped workflow records to become conversations.

## Solution

Honor path classification before the fast stream shortcut and route declared non-session payloads through typed raw-artifact admission in full and append live paths.

## Verification

- Focused real-route suite: 101 passed
- Adjacent taxonomy suite: 76 passed
- `devtools verify --quick`: passed
- `git diff --check`: passed

Ref polylogue-b508 and polylogue-ioz7

### DIFF
--- a/polylogue/pipeline/services/ingest_worker.py
+++ b/polylogue/pipeline/services/ingest_worker.py
@@ -19,7 +19,12 @@ from typing import TYPE_CHECKING, Literal, cast
 
 from typing_extensions import TypedDict
 
-from polylogue.archive.artifact_taxonomy import ArtifactClassification, ArtifactKind, classify_artifact
+from polylogue.archive.artifact_taxonomy import (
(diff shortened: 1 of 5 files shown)


Sinity:
Scope disposition: this PR is superseded by the merged sidecar admission work in Ref #3794 and the raw-admission routing in Ref #3772. The newer implementation covers the same full, append, and artifact paths with the shared ZIP admission contract. No additional code from this branch is required.

Sinity:
Closed as superseded by the merged sidecar and raw-admission implementations documented above.

PR #3806, opened 2026-08-05
fix(maintenance): make blob-ref liveness apply safe at scale

Sinity:
## Summary
Ref polylogue-0v4tn. This successor branch makes blob-reference-liveness apply executable at live orphan-ref scale with compact SQLite staging, candidate-relevant locked validation, cached backup attestation under ownership, bounded durable delete batches, and exact prepared-receipt recovery.

## Problem
The prior actuator held BEGIN IMMEDIATE across the full locked classification and delete plan. It also rechecked only the receipt and live fingerprint after ownership, so a backup artifact mutation between validations was not detected. The incident involved 69,340 orphan raw_payload refs, 10.6 GB peak memory, 39.2 GB reads, and rollback after three minutes.

## Solution
The candidate plan is staged in SQLite and retained on the owning connection. Locked hook validation is limited to candidate-relevant source_path/blob_hash pairs; legacy null-hook ambiguity remains fail-closed. The full backup artifact inventory is cached by a cheap stat signature and revalidated under ownership, with a real mutation invalidating the cache and causing a fresh hash validation. Apply deletes in 1,000-row transactions, appends batch progress, and recovers exact rollback, commit, or partial states from the durable prepared receipt without reparsing a candidate population under the write lock.

## Acceptance matrix
| Requirement | Evidence |
| --- | --- |
| Preserve ref-type, rekeyable-hook, and exact candidate identity rules | Existing join checks, exact key receipt digest, and fail-closed legacy null ambiguity remain in the reconciler path. |
| Avoid all-row duplicate Python populations | Candidate rows stay in temp SQLite staging; Python retains only bounded report samples and receipt streaming state. |
| Detect backup tampering after precheck | Real `backup_archive(..., verify=True)` fixture mutates the copied `source.db` between real validators; apply raises before any live delete. |
| Restrict locked hook work and bound lock tenure | Locked hook SQL uses candidate-relevant EXISTS filtering; 10,000 distinct paths execute in 10 batches of `BATCH_SIZE=1000`. |
| Durable prepared evidence and recovery | Real prepared receipt plus injected failure after delete and before commit leaves source rows intact; retry records `recovered_rolled_back` with exact count. Partial presence records `recovered_partial`. |
| Scale regression and no cross-product comparison | 10,000 distinct hook paths exercise the imported locked UDF with bounded calls; 10,000 shared legacy-null rows remain fail-closed with zero UDF calls. |

## Integration
Fresh `origin/master` is `a9f71f303`. The old 31-commit branch was not published as the critical fix. This successor contains only the coherent prerequisite stack and lane changes: `de54b64dc` (SQLite staging and scale path), `c989fab5a` (locked referent revalidation), and `38e1788ff` (backup inventory ownership check, candidate-relevant locking, short-lock batches, and real regressions). It is exactly three commits ahead of fresh master.

## Verification
- `direnv exec . devtools test tests/unit/storage/test_blob_ref_liveness.py tests/unit/storage/test_hook_payload_ref_reconciliation.py tests/unit/storage/test_blob_gc_ref_type_liveness.py` -> `27 passed in 16.10s`, devtools `ok (20.5s)`.
- `direnv exec . devtools verify --quick` -> exit 0; all 24 steps passed.
- The pre-push quick verification at commit `38e1788ff` -> exit 0; all 24 steps passed.
- Commit hooks -> 4 files formatted and linted.

## Production sequence
Run the default dry run first. Create and verify a fresh source backup. Then run `polylogue ops maintenance blob-reference-liveness --apply --backup-manifest <manifest.json> --receipt-file <new-receipt.jsonl>`. The prepared receipt is written before the first delete. Each committed batch appends progress, followed by a committed footer. If the process stops with a prepared receipt, rerun with the same archive and receipt path once to record exact recovery, then choose a fresh receipt path for a new apply. The command remains offline-only and requires the daemon to be stopped.

## Residual uncertainty
No live production archive was mutated. A same-size, same-mtime backup byte replacement remains outside the existing backup attestation threat model; ordinary artifact mutations change the cached stat signature and trigger a full inventory rehash. The full test suite and integration corpus were not run.

### DIFF
--- a/polylogue/cli/commands/maintenance/_blob_integrity.py
+++ b/polylogue/cli/commands/maintenance/_blob_integrity.py
@@ -74,7 +74,7 @@ def _render_blob_reference_liveness_plain(report: BlobRefLivenessReconciliationR
     click.echo(f"Source DB:    {report.source_db}")
     click.echo(f"Mode:         {'apply' if report.applied else 'dry-run'}")
     click.echo(f"Scanned:      {report.classification.scanned_count:,} blob_refs row(s)")
-    click.echo(f"Orphans:      {len(report.classification.candidates):,} row(s)")
+    click.echo(f"Orphans:      {report.classification.orphaned_count:,} row(s)")
     click.echo(f"Deleted:      {report.deleted_count:,} row(s)")
     click.echo(f"Safe to apply: {'yes' if report.classification.safe_to_apply else 'no'}")
     if report.classification.rekeyable_hook_payload_count:
--- a/polylogue/maintenance/blob_ref_liveness_reconciliation.py
+++ b/polylogue/maintenance/blob_ref_liveness_reconciliation.py
@@ -7,6 +7,7 @@ import json
 import os
 import sqlite3
 import time
+from collections.abc import Iterable, Iterator
 from contextlib import suppress
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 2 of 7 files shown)


Sinity:
Successor PR #3811 carries the final-review correction from head a68935f69a716fa64d90d7db8c805bb38feb0664. It adds fail-closed duplicate known-hash hook validation, an inter-batch source-change guard, candidate-bounded hook staging with the source index, and real scale regressions.

Sinity:
This predecessor is superseded by merged PR #3811. PR #3811 carries the final duplicate-identity and snapshot-wide scale correction described in this branch and is already on master. I am closing this stale predecessor rather than rebasing duplicate work.

PR #3803, opened 2026-08-05
fix(rebuild): require provenance receipts

Sinity:
## Summary

Require a fresh schema-inference PASS receipt for every production index rebuild entry point, with receipt evidence bound to the exact archive, durable source identity, source revision, and external corpus.

## Problem

Index rebuilds can mutate lease, generation, candidate, or resume state before proving that schema inference was performed against the same archive and source revision. Aggregate external-ground-truth counts and hashes also cannot prove that each raw row still maps to the same external file, including stale source-path rows.

## Solution

The production rebuild gate now runs before archive resolution, lease acquisition, generation creation, candidate mutation, and replay. The self-contained receipt model retains the typed provenance shape from the rebased upstream work and adds a canonical external inventory plus deterministic `raw_id` to external bundle/file mapping. Each mapping records source path, chosen external relative path, hashes, sizes, and disposition. Aggregate-only receipts, missing rows, stale source paths, changed mappings, changed snapshots, wrong identities, malformed receipts, stale receipts, future receipts, and forged top-level PASS receipts are rejected.

The CLI, HTTP endpoint, and daemon bulk resume path pass a policy-controlled receipt reference. Resume validates the receipt again before replay continues. Result receipts persist consumed evidence. The fixed receipt lifetime is 24 hours with a five minute clock skew allowance. The source revision snapshot hashes immutable raw source fields and excludes replay-generated authority and census fields introduced by the current master fast-forward work, so a valid rebuild does not self-invalidate during replay.

The branch was rebased onto current `origin/master`, including the upstream provenance producer and lifecycle fast-forward changes.

## Verification

- `direnv exec . devtools test $(git diff --name-only origin/master..HEAD | rg '^tests/(unit|storage)/.*\\.py$')` selected 247 tests and passed all 247.
- `direnv exec . devtools test tests/unit/maintenance/test_schema_inference_gate.py tests/unit/maintenance/test_rebuild_index_provenance_gate.py` passed 20 tests.
- `direnv exec . devtools test tests/unit/maintenance/test_rebuild_status.py tests/unit/cli/test_archive_maintenance_cli.py -k 'transaction_cursor_and_no_delta_when_source_unchanged or source_snapshot_drift_marks_candidate_stale_before_promotion'` passed 2 tests with 66 deselected.
- `direnv exec . devtools verify --quick` exited 0. Formatting, lint, strict typing, generated surfaces, layering, schema, policy, and documentation checks passed.
[...]

PR #3808, opened 2026-08-05
fix(rebuild): guard provenance across sharded resumability

Sinity:
## Summary

Add route-level regressions for the two remaining provenance cleanup gaps identified in the final non-delegating review.

## Problem

The full-source test expired the receipt before transaction creation, so it could pass without proving cleanup after both durable records existed. The shard cleanup test called private cleanup helpers directly, so it did not prove the production sharded replay route attempted every sibling after a post-graph provenance failure.

## Solution

The full-source regression calls the synchronous production rebuild route with `raw_ids` absent, lets the real generation store create the inactive candidate and transaction, expires the receipt after both records exist, and verifies both records are removed by setup cleanup. The sharded regression drives the production route through real shard replay, merge, and graph resolution, expires provenance after graph resolution, injects one real inactive-shard discard failure, and verifies sibling cleanup, primary-error preservation, cleanup diagnostics, and terminal-stage exclusion.

## Verification

- `env -u VIRTUAL_ENV direnv exec . devtools test tests/unit/maintenance/test_rebuild_index_provenance_gate.py tests/unit/maintenance/test_sharded_rebuild.py`: 19 passed in 49.79s.
- `env -u VIRTUAL_ENV direnv exec . devtools verify --quick`: passed at head `701c56754e38a78764f5fb5ca33ca5a512ad3b19`, exit 0.
- The pre-push quick verification repeated the same static gate at the pushed head and passed, including format, lint, mypy, generated surfaces, layering, policy labs, and schema promotion audit.

## Anti-vacuity evidence

- Full-source creation captures the real generation id and operation id only after `IndexGenerationStore.create_transaction` returns. It asserts the candidate metadata is inactive and the transaction JSON is readable before expiring the receipt, then asserts no generation directory or transaction JSON survives.
- The sharded case seeds eight temporary source rows and confirms two non-empty production shard buckets. The original shard replay, merge, and graph resolver execute before the test expires the receipt and invokes the real provenance validator. The test asserts every shard discard is attempted exactly once, all successful siblings are removed, only the deliberately failed discard remains, the parent transaction is removed, and `_repopulate_bulk_build_derived_state` plus `repair_session_insights` are never entered.
- The raised provenance error remains the primary exception. Its notes include the failed shard id and the synthetic discard diagnostic.

## Scope

Test coverage only. No production data, Beads state, or delegated agents were used.

Ref #3803

### DIFF
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -492,6 +492,29 @@ operation that already advanced past target *N* will not redo target
 
 ---
 
+### Rebuilding the derived index from source
+
+`polylogue ops maintenance rebuild-index` and the daemon bulk-rebuild route
+require a fresh schema-inference PASS receipt before they acquire the rebuild
+lease or create an inactive candidate. Pass `--schema-inference-receipt PATH`
+or set `POLYLOGUE_SCHEMA_INFERENCE_RECEIPT` in the policy environment. The
+same reference must be supplied when resuming an operation.
+
+Receipts are valid for 24 hours after `generated_at`. The validator allows a
+fixed five-minute clock-skew window for both future timestamps and expiry.
+It binds the receipt to the requested archive root, durable source and user
+tier identities, the current source revision snapshot, and a canonical
+external-ground-truth digest. That digest includes the complete external file
+inventory and a deterministic mapping for every raw row: raw ID, recorded
+`source_path`, selected external relative path, content hash, size, and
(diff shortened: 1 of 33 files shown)


Sinity:
This branch is superseded by #3818, which carries its production-route cleanup regressions, and by the current successor #3825, which contains the final provenance boundary and shard-cleanup implementation. I am closing the stale intermediate branch.

PR #3824, opened 2026-08-05
fix(daemon): gate unsafe cursor authority selection

Sinity:
## Summary

Block convergence and reindex before source selection when cursor authority is unsafe.

## Problem

Readiness detected an ahead cursor and 727 incomparable rows, but raw convergence and reindex did not consume that proof before selecting source rows. Deferred cursors were indistinguishable from other authority gaps.

## Solution

Classify cursor authority gaps with typed states, preserve excluded and semantic terminal exceptions, expose deferred authority as blocking unknown state, and use one shared readiness gate in raw materialization, recovery, and reindex before their first write.

## Verification

The worker completed its quick verification baseline, and focused public-readiness plus route-level mutation regressions cover unsafe blocking and valid exact-frontier progression.

Ref polylogue-xeck9

### DIFF
--- a/docs/evidence/polylogue-xeck9-cursor-authority-census-2026-08-04.md
+++ b/docs/evidence/polylogue-xeck9-cursor-authority-census-2026-08-04.md
@@ -44,7 +44,7 @@ The existing status DTO renders these separately from the proven violation: `cur
 
 ## Root-cause conclusion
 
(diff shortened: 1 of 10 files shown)


Sinity:
Duplicate of #3823 at the identical head commit; retaining #3823 as the review target.

PR #3878, opened 2026-08-07
fix(storage): attest hook reconciliation stages

Sinity:
## Summary

Make hook-match staging transactional, readiness-attested, and fail closed for blob liveness and garbage collection.

## Problem

An interrupted hook-match build could leave temporary relations behind. A later liveness candidate could mistake those partial tables for a complete stage and classify evidence incorrectly.

## Solution

The matcher now builds every relation under a savepoint, records a versioned readiness row only after population and integrity checks succeed, validates row counts and referential consistency on reuse, and clears all stage artifacts on any failure. Blob liveness and GC use the attested stage path and retain candidates when staging is unavailable. Failure injection covers every table-creation checkpoint and population start, followed by a rebuild assertion.

## Verification

- `devtools test tests/unit/storage/test_hook_payload_ref_reconciliation.py tests/unit/storage/test_blob_ref_liveness.py tests/unit/storage/test_blob_gc_ref_type_liveness.py` -> 50 passed
- `devtools verify --quick` -> all 24 steps exit 0
- `git verify-commit HEAD` -> passed

## Follow-ups

No production archive or daemon mutation was performed. Ref polylogue-tiozw.

### DIFF
--- a/polylogue/storage/blob_gc.py
+++ b/polylogue/storage/blob_gc.py
@@ -225,7 +225,21 @@ def _prepare_legacy_hook_liveness(conn: sqlite3.Connection) -> str:
     return "ready"
 
 
-def _blob_refs_still_live(conn: sqlite3.Connection, blob_bytes: bytes) -> bool:
+def _legacy_hook_liveness_status(conn: sqlite3.Connection) -> str:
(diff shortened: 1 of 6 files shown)


Sinity:
@codex review
@coderabbitai review

PR #3904, opened 2026-08-09
fix(daemon): make raw recovery probes fail closed

Sinity:
## Summary

Make the daemon raw-parse recovery probe fail closed, follow the active index, and distinguish a clean empty backlog from probe failure.

## Problem

The recovery probe could inspect the wrong derived index or collapse an exception into an apparently empty recovery backlog. That makes a readiness signal unsafe for reindex planning.

## Solution

- Bind the probe to the active index generation.
- Preserve typed probe failures instead of returning a clean result.
- Add production-route coverage for failure, active-index selection, and the genuine no-backlog path.

## Verification

- `devtools test tests/unit/daemon/test_raw_parse_recovery.py`: 15 passed after exact-head review repairs.
- `devtools verify --quick` at the exact head `4e7e3d3df`: all 24 steps passed.
- Live archive recovery remains separately gated and is not claimed by this PR.


Sinity:
@codex review the exact pushed head f3bb976e618bab6355dbd98b3e86d6a616d6b9db. Review only this head against the structured scope carrier and report actionable correctness, authority, or anti-vacuity gaps.

Sinity:
@codex review the exact pushed head 4e7e3d3df48606316d55a429d1829b3fbf735136 after the P1 repairs. Review only this head against the structured scope carrier.

PR #3913, opened 2026-08-10
test(devtools): guard proofs with production reachability

Sinity:
## Summary

Add a structured production-reachability oracle for proof tests. The oracle verifies that declared production seams call the tested production symbol and that the symbol is reachable from the declared production entrypoint.

## Problem

The previous proof gate could certify orphan helpers, over-approximate routes through nested function bodies or passed callable arguments, and resolve Python imports/caches incorrectly. The hermetic filesystem-boundary half remains a named successor.

## Solution

- Build a Python call graph with correct nested-callable, import, shadowing, package-relative, dotted-import, and class-method semantics.
- Reuse parsed production modules while invalidating the cache when source files change.
- Add mutation-oriented fixtures for the false-positive classes found in exact-head review.
- Keep the existing production-reachability fixture and `polylogue-jdesf` successor for hermetic path boundaries.

## Verification

- `devtools test tests/unit/devtools/test_production_reachability.py` — 7 passed.
- `devtools verify --quick` — all 24 steps exit 0 at `93ab11703c87ac777b9c8513dfa764be5e1a7de1`.

## Scope and residuals

This PR is partial for `polylogue-4v2d3`; it does not claim hermetic test-path enforcement, which remains `polylogue-jdesf`. It does not mutate production or the live archive.

Ref #4v2d3


Sinity:
@circleci run

Sinity:
@coderabbitai review

Sinity:
@circleci run

PR #3918, opened 2026-08-10
fix(daemon): recover typed CAS frontier failures

Sinity:
## Summary

Route typed CAS-frontier failures into stopped-daemon raw recovery only when the existing repair selector authorizes replay. The probe recognizes canonical and historical CAS carriers by exact raw-artifact coordinate, admits previously parsed but unmaterialized raws, and delegates materialization to the existing repair engine.

## Problem

`raw_materialization` already selected `deferred_cas_frontier` and `deferred_codex_cas_frontier` as retry authority. The daemon's path-scoped `raw_parse_recovery` probe initially recognized only untyped error strings. Its first typed-carrier version then diverged from the executor in two ways: validation-failed raws stayed pending forever, while previously parsed but unmaterialized CAS raws were skipped.

## Solution

`polylogue/daemon/convergence_stages.py` imports the established core replay-authority vocabulary and requires the same raw ID, origin, source path, source index, partial-decode support status, failed-validation exclusion, and materialization guard before returning pending work. It keeps legacy raw retries unparsed-only while allowing typed CAS authority for previously parsed raws. `tests/unit/daemon/test_raw_parse_recovery.py` creates real file-backed temporary archive tiers, writes authority and parse transitions through `ArchiveStore`, and exercises `raw_parse_recovery` through the production repair engine until the session materializes.

| Bead | Whole-Bead disposition | Evidence | Remaining scope |
| --- | --- | --- | --- |
| `polylogue-dyica` | Partial | `08c363b0d`; stopped-daemon route regressions; quick gate | Current live census, deployment, backup-gated dry-run/apply, immutable receipt, and reprocessing remain under `polylogue-reindex-source-remediation`. |
| `polylogue-dyica.1` | Satisfied implementation slice | Canonical, historical, validation-failed, and previously-parsed carrier route tests; pre-fix reproductions; quick gate | Its required live successor exists through parent `polylogue-dyica` and remains open as `polylogue-reindex-source-remediation`; no live apply is claimed here. |

## Verification

`devtools test tests/unit/daemon/test_raw_parse_recovery.py -k 'validation_failed_cas_frontier_failure or previously_parsed_cas_frontier_failure'` failed before this correction with two inverse `stage.check(path)` assertions. `devtools test tests/unit/daemon/test_raw_parse_recovery.py` then reported `19 passed`. `devtools verify --quick` completed all 24 checks with exit 0. The suite uses real temporary `source.db` and `index.db` files, not an in-memory substitute. Reverting either eligibility condition restores its corresponding production-route failure, so the regressions constrain the probe rather than a test-local replica.

## Adversarial review notes

The late Codex review found two actionable probe/executor eligibility mismatches. This update adds red file-backed regressions for both: validation-failed rows are not pending, and previously parsed, unmaterialized CAS rows drain through the repair engine.

## What this PR does not do

No production archive, daemon, migration, deployment, reprocess, dry-run apply, or receipt was mutated. The 2026-08-04 preflight observed 112 raw failures, including 111 parse failures and one maintenance failure. This PR does not remeasure that live population. Its current size and any reduction remain unknown until the open successor runs a fresh backup-gated census and reviewed apply.

### DIFF
--- a/polylogue/daemon/convergence_stages.py
+++ b/polylogue/daemon/convergence_stages.py
@@ -22,6 +22,10 @@ from typing import TYPE_CHECKING
 
 from polylogue.config import load_polylogue_config
 from polylogue.core.enums import Provider
+from polylogue.core.raw_failure_evidence import (
+    RAW_FAILURE_DEFERRED_SUPPORT_STATUS,
+    RAW_FAILURE_REPLAY_AUTHORITY_EVIDENCE_KINDS,
+)
 from polylogue.daemon.convergence import ConvergenceStage, StageExecuteReturn, StageExecutionResult
 from polylogue.daemon.convergence_standing_queries import make_standing_query_stage
 from polylogue.logging import get_logger
@@ -884,6 +888,7 @@ def _raw_parse_recovery_pending_count(db_path: Path, path: Path, *, archive_root
         return 0
     index_db = ArchiveLocation.resolve(durable_root).active_index_path
     normalized_root = str(path).rstrip("/")
+    replay_authority_placeholders = ", ".join("?" for _ in RAW_FAILURE_REPLAY_AUTHORITY_EVIDENCE_KINDS)
     try:
         conn = sqlite3.connect(f"file:{source_db}?mode=ro", uri=True, timeout=5.0)
     except sqlite3.Error:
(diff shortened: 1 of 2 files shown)


Sinity:
@codex review

PR #3951, opened 2026-08-11
fix(embeddings): preserve recipe snapshot in status reads (#3951)

Sinity:
## Summary

Archive embedding status now evaluates exact session counts with the same resolved recipe the payload reports. The regression test covers the public payload route across a genuine split archive.

## Problem

`_archive_embedding_status_payload()` resolved the configured model for its payload but delegated exact state counting to a helper that could reload ambient configuration. A model change could therefore let the status payload describe one recipe while classifying completed and terminal rows with another.

## Solution

`polylogue/storage/embeddings/status_payload.py` resolves one `EmbeddingRecipe` through the existing duck-typed configuration seam and passes it to the exact archive-state read. `tests/unit/storage/test_embedding_freshness_invariant.py` materializes an archive under `voyage-4`, resolves status under `voyage-5`, and requires the status payload to report that session pending. The contracts fixture also uses the current `voyage-4-lite` default while explicit recipe-change coverage remains on `voyage-4` and `voyage-5`.

## Verification

- `direnv exec . devtools test tests/unit/storage/test_embedding_contracts.py tests/unit/storage/test_embedding_freshness_invariant.py` -> `64 passed`
- `direnv exec . devtools verify --quick` -> successful, including ruff, mypy, generated renders, layering, and policy checks
- The pre-push hook reran `devtools verify --quick` at `5d58a78b7` successfully before publication.

Anti-vacuity: the regression does not mock `count_archive_embedding_session_state`; it uses `embedding_status_payload` with an actual `index.db` plus sibling `embeddings.db`, so omitting the recipe argument restores ambient `voyage-4` classification and fails the required pending count.

## Bead disposition

Self-contained scope. No Beads were assigned or mutated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Embedding status counts now accurately reflect the configured model and dimensions.
  * Session freshness and pending-embedding results remain correct when embedding configurations change.
  * Updated embedding validation to use the current Voyage-4 Lite model baseline.

### DIFF
--- a/polylogue/storage/embeddings/status_payload.py
+++ b/polylogue/storage/embeddings/status_payload.py
@@ -729,8 +729,8 @@ def _archive_embedding_status_payload(
     if index_db is None:
         return None
     recipe = EmbeddingRecipe.current(
-        model=str(cfg.embedding_model),
-        dimensions=int(cfg.embedding_dimension),
+        model=str(getattr(cfg, "embedding_model", "")),
+        dimensions=_payload_int(getattr(cfg, "embedding_dimension", 0)),
     )
     root = configured_root if configured_root is not None else db_path.parent
     conn = open_readonly_connection(index_db, timeout=STATUS_READ_BUSY_TIMEOUT_MS / 1000.0)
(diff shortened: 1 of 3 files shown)


Sinity:
@codex review

PR #3960, opened 2026-08-12
fix(testmon): run resumable shards with bounded argv

Sinity:
## Summary

Make resumable testmon seed shards executable at the current 4,096-node shard size and retain managed parallel execution.

## Problem

The post-merge seed reached shard 1 and spent 2,700 seconds executing 3,860 nodes before the supervisor timeout. Shard 2 then failed before pytest started with `OSError: [Errno 7] Argument list too long` because every node ID was appended to the `systemd-run` argv. Serial shard execution also turns the approximately 20,000-node seed into hours of fixture work.

## Solution

- write each shard's authoritative node IDs to a run-owned pytest response file;
- remove the collection-only worker setting and use the managed xdist pool (capped at ten workers) for shard execution;
- retain the per-shard checkpoint and outcome ledger so failures remain resumable and do not grant release authority;
- add focused coverage for response-file contents and managed xdist selection.

## Verification

- `devtools test tests/unit/devtools/test_verify.py` — 150 passed;
- `ruff format --check devtools/verify.py tests/unit/devtools/test_verify.py` — clean;
- `ruff check devtools/verify.py tests/unit/devtools/test_verify.py` — clean;
- xdist/testmon probe with four workers over API/property/query surfaces — testmon SQLite database created and writable; 325 passed, 11 pre-existing red nodes in 523 seconds.

The existing red nodes are retained as baseline evidence; this change does not silence or reclassify them.

### DIFF
--- a/devtools/pytest_progress_plugin.py
+++ b/devtools/pytest_progress_plugin.py
@@ -156,10 +156,20 @@ def pytest_collection_modifyitems(session: Any, config: Any, items: list[Any]) -
     _SELECTED_COUNT = len(items)
     limit = _selection_nodeid_limit()
     selected_nodeids = [str(getattr(item, "nodeid", item)) for item in items[:limit]]
+    # Marker metadata is a compact routing index, not a node-id sample. Keep
+    # it complete so seed sharding can isolate load-sensitive/TUI nodes even
+    # when the human-readable node-id sample is capped at 500 entries.
+    selected_node_markers = {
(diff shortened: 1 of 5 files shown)


Sinity:
The initial Circle failure was the missing structured PR-scope carrier, before pytest ran. I have added the v2 carrier to the PR body; the code head remains unchanged at 6042be0d7.

PR #4015, opened 2026-08-19
fix(storage): complete the raw-admission chokepoint with typed grouped and reconstruction arms

Sinity:
## Summary

Completes the raw-admission chokepoint (`polylogue-1fijp`, aggz Invariant 2) by resolving the last three architecturally-distinct creators of `raw_sessions` rows: grouped capture is migrated through the chokepoint under a new typed arm, the copy-forward repair is wrapped as an explicitly-reasoned named exemption, and the live-acquisition paths were re-surveyed and found already routed. Along the way it fixes a real identity bug in the BASELINE arm and closes `polylogue-2tfug`'s replay-ledger gap, both red-first.

## Problem

The bead's own survey named three remaining creators. Re-verifying each against current source changed the picture:

1. **`sources/live/append_ingest.py` / `batch.py`** — the note recorded these as incompatible with the chokepoint. They are not, any more: every call already passes `post_parse=True` and resolves the `POST_PARSE_PENDING` arm. No work needed; the note was stale.
2. **`archive_ingest.py` main path** — already routed through `admit_raw_and_parsed_result`, *except* on its `shared_raw=True` branch, which issued a bare `write_source_raw_session`. That branch is the grouped-JSONL path (a resume-fork chain, a Drive bundle), i.e. the high-traffic case the note was pointing at.
3. **`storage/repair.py` copy-forward** — a hand-assembled `INSERT INTO raw_sessions`. It reads like the last unmigrated call site, and a future sweep would try to "finish" it.

A second census pass over the whole tree, run after the grouped migration landed, found the bead's list was incomplete. Three further classes existed:

4. **`write_raw_blob_and_parsed` / `_result`** (+ two `ArchiveStore` facade methods) — the blob-ref twin of `write_raw_and_parsed_result`, i.e. the same bare `FULL`/`revision=None` limbo write, with **zero callers in `polylogue/` or `tests/`**.
5. **`durable_change_train.py` ×3** — anonymous `INSERT INTO raw_sessions` statements.
6. **`write_raw_payload` / `write_raw_blob_ref`'s `post_parse=False` branch** — a wrapper arm that writes a bare row without consulting the chokepoint.

Two defects surfaced while verifying, both proven red before any fix:

- The `BASELINE` arm is the only arm that never forwarded the caller's explicit `raw_id` to `write_source_raw_session`. A route with its own identity scheme — `hermes_profile_raw_id`, or the grouped deterministic key — addressed one raw row and silently got another. **Correction (cold review):** this PR previously described the harm as memberships recorded against a phantom id. That is wrong for this route — `replace_raw_membership_census` only runs under `shared_key`, which Hermes never had. The real harm is that one Hermes snapshot minted N raw rows for one set of bytes (the duplicate-raw pattern `polylogue-sjf6` removed for Claude Code) while `hermes_profile_raw_id` was computed and then discarded.
- `polylogue-2tfug`: reparsing the raw that is *already* the accepted head moved `sessions.content_hash` and left `raw_revision_heads.accepted_content_hash` behind. `validate_raw_replay_application_receipt` requires the two to agree, so a formal replay audit would report a legitimate parser correction as ledger inconsistency.

## Solution

**Grouped capture → `RawAdmissionArm.SHARED_GROUPED`.** `admit_raw_and_parsed_result`'s `shared_raw` branch now resolves a named arm instead of writing its own row. The persisted row is byte-for-byte what the bypass produced (`native_id` NULL, no revision envelope, source.db's `revision_kind` default) — a grouped payload's identity is its acquisition coordinate, and which sessions it speaks for lives in `raw_session_memberships`, so a revision chain has no single subject to be about. This moves the *decision*, deliberately not the bytes; changing the stored envelope would be a semantic change to source.db and is out of scope for an acquire-side bead.

**Copy-forward → `insert_reconstructed_raw_row()` (`source_write.py`).** This is an architectural exemption, not a deferred migration. `admit_raw_observation` decides what a fresh observation *means* from bytes plus the prior head; a copy-forward has neither — the source is gone and the verdict was already adjudicated by a proven repair plan. Routing it through the chokepoint would resolve `BASELINE`/`asserted` against the absent prior head and **discard the `BYTE_PROVEN` authority the plan established**. `ReconstructedRawRow` makes every envelope column a required field, so the no-nullable-limbo invariant holds on this path by construction; what does not apply is the resolution step. It lives in `source_write.py` rather than beside the chokepoint because `raw_admission.py`'s contract is that it composes writers and issues no SQL of its own — `devtools verify layering` caught the first placement as `writer_module_unmarked_mutation`.

**Dead blob-and-parsed pair → deleted.** `write_raw_blob_and_parsed`/`_result` and their facade methods are removed. Their payload twin is already superseded by `admit_raw_and_parsed_result`, and nothing anywhere calls the blob variants. Keeping a dead *public* entrypoint that bypasses the chokepoint would leave the invariant merely bounded rather than true.

**`durable_change_train` ×3 → `_seed_probe_raw_row()`.** These are **not** the copy-forward class, and the evidence is unambiguous: every caller is a `_probe_*` function writing into a `tempfile.TemporaryDirectory` archive deleted before it returns, and each deliberately seeds an *unusual* row — a `detected_provider` that disagrees with `origin`, a row carrying a `parse_error`, a bare row with no envelope — precisely to prove a migrated reader still handles that shape. Routing them through `admit_raw_observation` or `insert_reconstructed_raw_row` would normalize the row into a well-formed envelope and **destroy the condition under test**. The named helper therefore keeps the shapes and only names the class, so a census can classify it at a glance.

**`post_parse=False` branch → surveyed, documented, not migrated.** No live acquisition route reaches it: every production acquisition caller (`append_ingest.py`, `batch.py`) passes `post_parse=True` and resolves `POST_PARSE_PENDING`. What remains is fixture seeding — `durable_change_train`'s self-probes and `tests/infra` corpus builders, which need row shapes admission would normalize away. The docstring now says so and tells a future reader to treat a new production caller as a bypass, not precedent.

**Resulting census.** Every anonymous `INSERT INTO raw_sessions` in the tree is gone. The only three left are `source_write.py`'s two writer primitives (this tier's declared writer module, which the chokepoint composes) and the one inside the named `_seed_probe_raw_row` helper.

**Multi-session Hermes snapshots → the grouped arm (cold-review BLOCK, fixed).** Honoring the explicit `raw_id` in `BASELINE` exposed a real regression in `parse_sources_archive`: `hermes_profile_raw_id` is keyed per *snapshot* and excludes the session id, so one `state.db` carries N sessions under one acquisition id. `archive_ingest` computed `shared_key` only when `raw_id` was still `None`, so those payloads took the non-grouped path and were admitted per session as `BASELINE`; the second session collided with the first's row (`raw id is already bound to different acquisition evidence`), and `write_pair` catches only `ContentExcisedError`, so the whole batch aborted — 3 parsed sessions became 1 raw row and 0 indexed sessions.

The fix routes them through `SHARED_GROUPED`, which is what a Hermes snapshot actually is: `shared_key` is now computed whenever a blob hash exists, and the generic native_id-free id is derived only when the route did not already supply one, so Hermes keeps its snapshot-scoped identity instead of having it overwritten.

Chosen over the alternative of dropping `hermes_profile_raw_id` from the `BASELINE` arm. That would restore master's behaviour — but master's behaviour *is* the bug in slower motion: the arm ignored the explicit id and derived a per-session one, minting N raw rows for one set of bytes and leaving `hermes_profile_raw_id` dead on this route. Routing the snapshot through the arm matching its real shape fixes the abort and makes that identity mean what its docstring says. Scope is `parse_sources_archive` only (import/reprocess/demo); the live daemon Hermes route admits through `POST_PARSE_PENDING` and never reached this branch.

**`polylogue-2tfug` receipt reissue.** Both ordinary write call sites now detect the narrow shape — same raw, same session, changed hash — and reissue a receipt that moves the head. Detection runs *before* `write_parsed_session_to_archive` so both land in its transaction; placing it after leaves the head update uncommitted and silently lost, which is how the first attempt failed. The decision gets its own `ApplicationDecision.REPARSE_REAFFIRMATION` rather than borrowing `SELECTED_BASELINE`'s "this raw won a contest" meaning — no contest happened. That widens the generated CHECK on `raw_revision_applications.decision`, declared as **index v68 / `CONSTRAINT_ONLY`** (purely permissive; no stored row changes meaning, no parser semantics), so it fast-forwards rather than forcing a rebuild.

## Verification

Every command ran in this lane's own worktree with its own venv (checkout guard confirms `polylogue → /realm/worktrees/1fijp-admission-chokepoint`).

- `devtools verify --quick` — **all ten steps ok**, including `verify layering ... ok` and `lab policy schema-versioning ... ok` (`undeclared index schema deltas found: 0`, accepting the v68 declaration).
- `devtools test tests/unit/pipeline` + all four Hermes suites + raw-admission — **613 passed**.
- `devtools test` over repair, browser-capture-origin-repair, revision-application/replay, accepted-head-reparse and archive-tiers — **216 passed**.
- `devtools test` over the chokepoint, durable-change-train, repair, revision-application/replay, archive-tiers and shared-raw suites, fixed order — **346 passed, 2 failed** (the pre-existing pair below).
- Earlier full-directory runs: `tests/unit/storage` — 3 failed, 2680 passed; `tests/unit/pipeline tests/unit/maintenance` — 3 failed, 1337 passed.

**Every failure is pre-existing, established by controlled baseline rather than assertion.** Reverting `polylogue/` *and* `tests/` to master in this same worktree and rerunning the same files under fixed order (`-p no:randomly`) reproduces each set identically. For the two durable-change-train failures this was re-baselined *after* the probe rewrite specifically to confirm I had not taken ownership of them: master gives `2 failed, 100 passed, 1 xfailed` on that file, and so does this branch, with the same two node ids. The six are `test_durable_change_train.py::test_bootstrap_reconciles_and_persists_interrupted_train_evidence`, `::test_bootstrap_finishes_persisted_applied_train_without_reapplying`, `test_rebuild_paging_content_order.py::test_rebuild_content_order_paging_dedups_first_time_classification_via_content_cache`, `test_message_owner_scope_backfill.py::test_stale_plan_wrong_backup_and_active_daemon_are_fail_closed`, `::test_prepared_recovery_rejects_a_nonidentity_assertion_change_after_commit`, `test_rebuild_parse_apply_split.py::test_rebuild_index_from_source_sync_auto_engages_pipelined_decode`.

**Full `devtools verify` deliberately skipped, under the lane gate policy for bootstrap-predicted runs.** This lane's plain verify reported `selection: bootstrap (absent)` for environment digest `polylogue-e962f60d05f5…`; the attempted run was killed by its own 596s cap partway through a ~2700s bootstrap. That digest now *has* a row in `.cache/testmon/testmondata`, but only the partial one that interrupted run left behind, so a fresh full verify is at best incomplete and at worst a second full bootstrap. Per policy, focused `devtools test` plus `devtools verify --quick` are this PR's gate and the merge receipt is recorded centrally at head rather than by this lane.

Anti-vacuity: the Hermes regression test fails on the pre-fix head with the exact production error and passes after; reverting only the `shared_key` guard reproduces it. The two earlier red commits fail against the pre-fix code with the exact assertions named. The copy-forward path's existing coverage was confirmed by injecting an `AssertionError` into the new constructor — it fires 24 times in `tests/unit/storage/test_browser_capture_origin_repair.py` and zero times in `test_repair.py`, which is why the direct regressions were added. The dead-code deletion is confirmed reachable-by-nobody by grep across `polylogue/` and `tests/`, and `verify layering` independently caught the stale entrypoint manifest.

## Bead disposition matrix

| Bead | Disposition | Evidence | Residual |
| --- | --- | --- | --- |
| `polylogue-1fijp` | **partial** | AC(a) holds for every reachable path after a second whole-tree census: no anonymous `INSERT INTO raw_sessions` remains. The three left are `source_write.py`'s two writer primitives (composed by the chokepoint), and the one inside the named `_seed_probe_raw_row` temp-archive helper; the copy-forward is the named `insert_reconstructed_raw_row` exemption; the dead blob-and-parsed bypass pair is deleted. AC(c)/(d) satisfied structurally by arms 4/5 (unchanged). | Grouped raws still persist a NULL revision envelope (`revision_kind` default `unknown`) — the arm names the decision but does not eliminate the limbo. AC(e) (zero new quarantined rows over 72h of live daemon operation) is unattempted and needs live time, not code. |
| `polylogue-2tfug` | **partial** | AC1/AC2 by `tests/unit/storage/test_accepted_head_reparse_receipt.py::test_reparse_of_accepted_head_keeps_head_and_session_content_hash_in_sync` (red→green, real write path). AC4 by `::test_unchanged_reparse_of_accepted_head_issues_no_new_receipt`. | AC3 not done: `devtools/raw_authority_restart_proof.py` has no coverage for this case. |
| `polylogue-omsw` | **partial** | Arm 4 is wired into all three acquisition routes (`archive_ingest.py`, `append_ingest.py`, `batch.py`). Verified live: `tool-results/toolu_*.json` → `parse_as_session=False, cohort=tool_result_sidecar`; `subagents/workflows/*/journal.jsonl` → `cohort=workflow_journal`. | The codex-database arm does **not** fall out of the resolution arms. `classify_artifact_path` returns `None` for `state_5.sqlite`/`goals_1.sqlite` because that class is decided by *content* (magic bytes), not path — and that refusal already landed under the now-closed `polylogue-hbtj2`, with its own `--apply`-gated reclassify actuator. Reclassifying existing contaminated rows (AC3) remains untouched and is durable-tier work needing operator consent. |

Every residual outcome above carries an open successor with a durable `relates-to` edge: `polylogue-1fijp` → **`polylogue-ogrri`** (grouped-raw NULL revision envelope + AC(e) 72h live observation), `polylogue-2tfug` → **`polylogue-n6rkz`** (`raw_authority_restart_proof.py` coverage), `polylogue-omsw` → **`polylogue-mcnzt`** (contaminated-row reclassification; durable-tier, consent-gated).

Ref polylogue-1fijp, Ref polylogue-2tfug, Ref polylogue-omsw


Sinity:
**Blocked on successor beads before this can go non-draft.**

`pr-scope render` refuses the carrier, and `merge-gate record` refuses the receipt without one:

```
REFUSING to render pr-scope carrier: invalid scope input:
  polylogue-1fijp: partial disposition requires a named successor Bead;
  polylogue-2tfug: partial disposition requires a named successor Bead;
  polylogue-omsw: partial disposition requires a named successor Bead
```

All three dispositions are honestly `partial` (residuals are itemized in the body's matrix), and the lane contract forbids inventing successor IDs or creating beads. Three successors are needed:

1. Grouped-raw revision envelope still NULL after `SHARED_GROUPED`, plus 1fijp AC(e)'s 72h live-daemon observation.
2. `devtools/raw_authority_restart_proof.py` coverage for the accepted-head reparse case (2tfug AC3).
3. omsw's contaminated-row reclassification (durable tier, needs operator consent).

No existing open bead covers any of them — `sp72` is Drive lineage; `hbtj2` and `lzh8` are closed. Once they exist I (or the coordinator) can render the carrier, record the receipt at head, and undraft.

Sinity:
**Update — second census pass + gate substitution.**

Your intel was right that the invariant did not hold, and re-running the census at current HEAD sharpened it. Two corrections to the sites you listed:

- `revision_governance.py:3698` is **gone** — that was the `shared_raw` bare write, migrated to the `SHARED_GROUPED` arm in `82dd39cc4`. Your line numbers predate it.
- `:551` (now `:626`) is real but is **not a live admission path**: every production acquisition caller passes `post_parse=True`. Its `post_parse=False` arm is reached only by `durable_change_train`'s self-probes and `tests/infra` corpus builders. Surveyed and documented rather than migrated.

The census also found two classes neither of us had listed: `write_raw_blob_and_parsed`/`_result` plus two `ArchiveStore` facade methods — the blob twin of the limbo write, **zero callers anywhere** — now deleted; and `:678`, the blob-ref `post_parse=False` arm (tests only).

On `durable_change_train`: the evidence does **not** support the copy-forward class. All three sit in `_probe_*` functions writing to `tempfile.TemporaryDirectory` archives deleted on return, and each deliberately seeds a malformed row (`detected_provider` disagreeing with `origin`; a `parse_error` row; a bare row) to prove a migrated reader still handles it. Wrapping them in `insert_reconstructed_raw_row` would normalize the row and destroy the condition under test. I wrapped them as `_seed_probe_raw_row()` instead — named and census-legible, shapes preserved.

Net: every anonymous `INSERT INTO raw_sessions` is gone. The three remaining are `source_write.py`'s two writer primitives and the one inside the named probe helper.

**Gate:** `devtools why` on this lane's full verify predicts `bootstrap (absent)` (digest `polylogue-e962f60d05f5…`; the attempt was killed by its own 596s cap mid-bootstrap). Per your policy I skipped the full verify — `verify --quick` is green on all ten steps and focused tests are 346 passed / 2 pre-existing failed (re-baselined against master *after* the probe rewrite). Please record the merge receipt centrally at head `1ee2dd35e`.

Still draft, still blocked on the three successor beads from the earlier comment.

PR #4142, opened 2026-08-24
test: bound rebuild memory and repair authority fixtures

Sinity:
## Summary

- partition maintenance rebuild-index tests into four deterministic xdist groups to bound concurrent high-memory rebuilds without reducing coverage
- align archive/pathology fixtures with the canonical Claude authority key
- preserve schema and quarantine repair behavior while making fixture expectations explicit

## Verification

- pre-push quick verification passed at exact head
- 22 rebuild/pathology tests passed
- 5 quarantined accepted-raw repair tests passed
- full 12-worker verification will run through the registered AgentCTL operation before landing

### DIFF
--- a/polylogue/maintenance/pathology_zoo.py
+++ b/polylogue/maintenance/pathology_zoo.py
@@ -82,7 +82,7 @@ _CHATGPT = "chatgpt-export"
 _DESIGN = "claude-design-session"
 _GEMINI = "aistudio-drive"
 _CLAUDE_VINTAGE_NATIVE_ID = "9ed2056f-b415-4f51-b18e-5265f21a67bf"
-_CLAUDE_VINTAGE_LOGICAL_SOURCE_KEY = f"claude-ai:{_CLAUDE_VINTAGE_NATIVE_ID}"
+_CLAUDE_VINTAGE_LOGICAL_SOURCE_KEY = f"{_CLAUDE_AI}:{_CLAUDE_VINTAGE_NATIVE_ID}"
 
 
 PATHOLOGY_ZOO_MANIFEST: tuple[PathologyZooMember, ...] = (
--- a/polylogue/storage/repair.py
+++ b/polylogue/storage/repair.py
@@ -1057,7 +1057,7 @@ def _inspect_quarantined_accepted_raw(
     if (
         origin_from_provider(parsed.source_name) != origin
         or str(make_session_id(parsed.source_name, parsed.provider_session_id)) != str(head["session_id"])
-        or f"{provider.value}:{parsed.provider_session_id}" != logical_source_key
+        or f"{origin_from_provider(parsed.source_name).value}:{parsed.provider_session_id}" != logical_source_key
         or bytes.fromhex(session_content_hash(parsed)) != accepted_hash
     ):
         return _quarantined_raw_item(raw_id, "normalized parser identity or content differs from the accepted session")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,7 @@ import shutil
 import sqlite3
 import sys
 import threading
+import zlib
 from collections.abc import AsyncIterator, Callable, Iterator, Mapping
 from pathlib import Path
 from types import FrameType, ModuleType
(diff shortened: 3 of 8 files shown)


Sinity:
Fixed all findings from the review thread in 554b6bb06. Dispositions below.

**1. `tests/infra/pathology_zoo_fixtures.py:32` — writable clones for copy-based mutation tests (P1).**
Already fixed on this branch by an earlier commit (`2a25b1a9d`, before this comment landed): `pathology_zoo_archive` clones the canonical artifact into a session tmp dir via `clone_pathology_zoo`, and `clone_immutable_tree` explicitly adds `S_IWUSR` to every path in the clone. Verified: `devtools test tests/unit/infra/test_pathology_zoo.py tests/infra/test_timeout_policy.py` (20 passed) and the copytree-then-mutate consumers in `tests/unit/maintenance/test_archive_verification.py` (245 passed alongside `test_reindex_canary.py`).

**2. `tests/infra/pathology_zoo.py:719` — clone pathology zoos from their supplied cache (P2).**
Fixed. `clone_pathology_zoo` no longer rebuilds/looks up an artifact under `default_cache_root()` and validates the caller's root against it; it clones directly from `zoo.archive_root`, so a zoo built with `build_pathology_zoo_ro(cache_root=...)` is respected instead of silently redirected.

**3. `devtools/verify.py:162` — keep testmon enabled while bootstrapping the graph (P1).**
Fixed. `_native_pytest_steps` now runs full-corpus lanes (`testmon_mode == "all"`, covering both bootstrap-via-`--all` and deliberate `--all`) with `--testmon --testmon-env=... --testmon-noselect` instead of omitting `--testmon` entirely, so the corpus run populates `.cache/testmon/testmondata`. Updated `test_full_corpus_does_not_duplicate_testmon_tracing_across_workers` (which had asserted the buggy behavior) to `test_full_corpus_traces_but_never_selects_testmon`, asserting `--testmon` + `--testmon-noselect` present and `--testmon-forceselect` absent.

**4. `tests/infra/pathology_zoo_fixtures.py:32` — invalidate the shared zoo when ingest semantics change (P1).**
Fixed. `_pathology_cache_key()` now folds in the same production fingerprints `tests.infra.workload_artifacts` uses for seeded archives (`_recipe_id`, `_source_semantics_id`, `_archive_schema_id` — parser/materializer/DDL closure), so a parser, ingestion, or archive-schema change invalidates the cached zoo instead of reusing a stale artifact.

**5. `tests/unit/maintenance/test_reindex_campaign.py:124` — pin the shared campaign fixture to one xdist worker (P1).**
Fixed. Added `pytestmark = pytest.mark.xdist_group(name="reindex-campaign")` at module scope so all four tests land on one worker under `--dist=loadgroup`, matching the rebuild-index load-group pattern already in `tests/conftest.py`.

**6. `tests/conftest.py:72` — keep pytest hook parameters required (P1).**
Fixed. `pytest_collection_modifyitems(config, items)` no longer has default values (pluggy omits optional hook params and was calling it with both `None`). Updated the one direct unit-test caller (`tests/infra/test_timeout_policy.py`) to pass a stub `config`.

**7. `tests/infra/pathology_zoo_fixtures.py:32` — preserve the canonical zoo for writable fixture clones (P1).**
Fixed by the same change as #2: since `clone_pathology_zoo` no longer requires its input to equal the canonical cache artifact, cloning from `pathology_zoo_archive` (itself already a clone) in `pathology_zoo_writable` now succeeds. Verified via `tests/unit/maintenance/test_reindex_canary.py` (uses `pathology_zoo_writable`), part of the 245-pass run above.

**8. `polylogue/storage/repair.py:1060` — use origin keys throughout source-v7 census staging (P1).**
Fixed. `_stageable_quarantined_census_cohort` compared `f"{provider.value}:{provider_session_id}"` against the origin-keyed `logical_source_key`; switched to `f"{origin_from_provider(sessions[0].source_name).value}:{...}"`, matching the convention used at the outer call site. Added a red-first regression test, `test_stageable_quarantined_census_cohort_keys_by_origin_not_provider`, that fails with "same-source-path cohort differs from the accepted session" against the pre-fix code (confirmed by reverting just this hunk) and passes with the fix.

**Verification**
- `devtools test tests/unit/storage/test_quarantined_accepted_raw_repair.py -k test_stageable_quarantined_census_cohort_keys_by_origin_not_provider` — red against reverted fix, green restored
- `devtools test tests/unit/storage/test_quarantined_accepted_raw_repair.py tests/unit/storage/test_repair.py` — 116 passed
- `devtools test tests/unit/infra/test_pathology_zoo.py tests/infra/test_timeout_policy.py` — 20 passed
- `devtools test tests/unit/maintenance/test_archive_verification.py tests/unit/maintenance/test_reindex_canary.py` — 245 passed
- `devtools test tests/unit/maintenance/test_reindex_campaign.py` — 4 passed
- `devtools test tests/unit/devtools/` — 850 passed, 2 skipped
- `devtools verify --quick` — all gates ok (mypy, `render all --check` all `sync OK`, layering, oracle-integrity, schema promotion/privacy)

**Residual risk**: the fixed testmon-bootstrap behavior and the pathology-zoo cache-key/clone changes are covered by focused/selected tests above, not a full corpus run — a full `devtools verify --all` was intentionally not run per lane discipline. No findings were skipped.

